PR #33, opened 2016-02-14
linting rules node style

pertrai1:
Have to start to get more strict in our code styles. Will start to implement this into our build before commits.


pertrai1:
New rules and file not passing. Deleting pull request

PR #37, opened 2016-02-16
Overhaul README adding logo

pertrai1:
The README for this repo should only need to
have docs for setting up the web application with
a link to the WeVoteServer at bottom


nf071590:
Why'd you delete this?

pertrai1:
@nf071590 Dale said he did not want to change the README until after version 1 - more than 2 weeks from now - so I deleted so the branch was not dangling out there. I thought it was a good idea but Dale did not think so.

pertrai1:
@nf071590 I restored so you can have your say whether this looks good for a README that splits the repo's into areas of concern. I was thinking of getting more contributors and felt that the README would have confused them since the backend is not needed locally anymore.

PR #56, opened 2016-02-18
save location update

nf071590:
this update allows user to save their location using the /settings/location route of the application, it then redirects them to their ballot page.


nf071590:
making new pull request

PR #63, opened 2016-02-19
Remove links from candidate page, reduce linting errors

lisamcho:
Removed link on candidate's name and organization name on candidate detail page (won't be connected for first release). Removed unused code from render method. Fixed linting errors.

### DIFF
--- a/src/js/actions/PositionActions.js
+++ b/src/js/actions/PositionActions.js
@@ -1,6 +1,6 @@
-'use strict';
-var AppDispatcher = require('../dispatcher/AppDispatcher');
-var PositionConstants = require('../constants/PositionConstants');
+"use strict";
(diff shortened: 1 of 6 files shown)


DaleMcGrew:
Hi @lisamcho This looks good and is ready for merging. :+1: In general I prefer to comment out links instead of deleting them when they are going to be turned back on soon, since there is useful information captured there. For example in components/Ballot/PositionItem.jsx this link isn't trival to recreate, and in the future might want to be commented out instead of deleted: `<Link className="" to="ballot_candidate_one_org_position" params={{id: position.speaker_id, org_id: position.speaker_we_vote_id}}>`

pertrai1:
:+1:

Issue #53, opened 2016-02-18
Address doesn't fit on top


nf071590:
When address is populated in the header region, it goes beneath the icons to the left

nf071590:
I would prefer to put the location/address on the left of all the icons and let the icons float up against the right border and stay flush

pertrai1:
:+1:  On alignment of location

pertrai1:
@DaleMcGrew Do you want me to do anything with the ellipsis on this before end of this milestone?

DaleMcGrew:
Might as well -- we need to solve the problem of long addresses very soon. Thanks.

pertrai1:
Ok, position it in the center as well?

DaleMcGrew:
Either of these (whichever is easier, or in this order of preference if equally easy/difficult):
1) centered between "My Ballot" and the 4 icons
2) ~100px to the right of "My Ballot"

PR #73, opened 2016-02-21
#53 Address in header alignment

pertrai1:
The address fits in the middle of the screen on desktop and moves to the right on mobile devices. If the address is long, it will get cut off to save space


DaleMcGrew:
:+1:

nf071590:
:+1:

Issue #65, opened 2016-02-19
npm test: check more folders


DaleMcGrew:
It seems like `npm test` is only checking certain folders for lint errors. Add all folders with code we want to verify. Does not seem to pick up .jsx files.

Specific folders that don't seem to be included:
/src/js/components
/src/js/routes

We might want to exclude:
/src/js/mock-data

pertrai1:
This has been fixed with #78 pull request

PR #78, opened 2016-02-21
#65 adding jsx to the linting

pertrai1:
NOTE: a lot of errors so you will have to pass `-n` to your commit message to get past the commit hook stopping the commit. There are close to 150 errors now with adding .jsx files. `mock-data` folder being excluded now

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,1 +1,2 @@
 src/js/vendor/*.js
+src/js/mock-data/
--- a/.eslintrc
+++ b/.eslintrc
@@ -101,7 +101,7 @@
     "no-label-var": [2],
     "no-labels": [2],
     "no-lone-blocks": [2],
-    "no-lonely-if": [2],
+    "no-lonely-if": [1],
     "no-loop-func": [1],
     "no-mixed-requires": [0, false],
     "no-mixed-spaces-and-tabs": [2, false],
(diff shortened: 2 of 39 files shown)


DaleMcGrew:
I am handing out jerseys for Little League this morning, but should be able to take on some of the cleanup work late this afternoon Pacific time.

pertrai1:
I am not available until later on this afternoon as well. Will look at the errors and make fixes as well.

DaleMcGrew:
Hi @pertrai1, I notice that in this pull request you remove a number of the comments that start with "VISUAL DESIGN HERE". The link to the mockup is very helpful to me when I'm building out and iterating on a page. What are your thoughts?

pertrai1:
@DaleMcGrew I think that it is important that we have an issue that can always reference a pull request and it is in the issue that there is either a link to the artifacts, or a link to a wiki page that can be created for the artifacts for that particular module. Does that sound like a reasonable plan?

DaleMcGrew:
I'm open to talking about that. I like the improvements to our procedures that we are looking at. Until we have a chance to talk about it and transition, I don't want to lose comments in the code. I will add the "VISUAL DESIGN HERE" links back to the code for now until we get a chance to talk about migrating to a better system for linking to designs.

pertrai1:
@DaleMcGrew Ok, please let others know that is code comments are needed to stay in the code, have some reference so they are not removed, i.e - KEEP, NO REMOVE, etc.

PR #86, opened 2016-02-22
Candidate page updates

lisamcho:
This code 
-Fixes bug with candidate page position_list (where is_oppose and is_support logic was wrong). There is still a bug on the backend API (You can see positions with status "Unknown" where both is_support and is_oppose are true or both are false).
-Fixes bug with position_item statement text (was showing blank statement text when API call returned a non-blank value).
-Formats date on candidate page
-Adds candidate name to each positionItem view.

Note: I got around the issue with the git precommit hooks by running the following:
git commit --no-verify -m

I hope that is okay. The code runs on localhost:3000.

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,6 +14,7 @@
   },
   "dependencies": {
     "keymirror": "^0.1.1",
+    "moment": "^2.11.2",
     "object-assign": "^4.0.1",
     "react": "^0.14.3",
     "react-dom": "^0.14.3",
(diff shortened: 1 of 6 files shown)


lisamcho:
@DaleMcGrew @pertrai1

pertrai1:
@lisamcho @DaleMcGrew Yeah please use the --no-verify until I can get a better handle on the test suite in the next couple of days. :+1:

PR #88, opened 2016-02-23
Staraction hover

nf071590:
StarAction icon now shows cursor:pointer when hovered to indicate that it is clickable to users...


pertrai1:
@nf071590 Is this failing CI because of babel-node and faucet?

pertrai1:
If only because of CI test problem, :+1:

nf071590:
Yeah, I modified build to only lint for now until those dependencies can be tested and resolved on my next request after this one.

PR #89, opened 2016-02-23
Build fix

nf071590:
Fixing build test script and some linting errors. `babel-node` and `faucet` do not exist in *package.json* and this is why our build process keeps failing. Using fallback eslint script until dependencies can be resolved successfully.


pertrai1:
:+1:

PR #99, opened 2016-02-23
Opinions

nf071590:
… indicate that its clickablemaking mouse change to cursor pointer when hovering hamburger icon to indicate that its clickable


nf071590:
changes to file header.jsx and scss additions are only important changes on this commit... others are related to issue that I am fixing and not completed with yet.

PR #121, opened 2016-03-04
Major refactoring

lisamcho:
Hey this will need lots of code review, but I think it's ready to merge into main branch.

The last commit of the flux-prototype branch I think is work-in-progress on the VoterStore / location because you get error (0 , _service.$ajax) is not a function. (right @nf071590 )?

But meanwhile, we can leave the old VoterStore intact but merge in the new AppDispatcher and most of Nick's commits from flux-prototype as well as changes I made to refactor some of the main views and stores based on that. 

Also, there seems to be a lot of files that aren't in use. Some are just renamed with a dot (.BallotStore.js). I put some others that I'd like to delete in the 'deprecated' folder, unless anyone thinks otherwise?

@nf071590 @pertrai1


pertrai1:
@lisamcho This might have been better served in smaller pull requests since it is a major refactor. I am good with this, just took a good amount of time to go through. :+1:

PR #125, opened 2016-03-12
Call change address on keyboard enter as well as submit buttom

lisamcho:
This is a minor change to call saveLocation when user hits enter on the keyboard as well as hitting the Submit button. There are still errors on how the response from the server is handled (response often won't return a valid ballot for some addresses) that this does not attempt to fix.

@DaleMcGrew


pertrai1:
@lisamcho Once the travis build passes, go ahead and merge

PR #126, opened 2016-03-12
Refresh header on address change

lisamcho:
The address in the header now updates when you change the address.
This also removes the logic to check for GOOGLE_CIVIC_API_ERROR on voterAddressSave, since the API is no longer returning this.

@pertrai1


pertrai1:
:white_check_mark:

Issue #123, opened 2016-03-12
Header Top Navigation, Desktop: Lock top bar in place


DaleMcGrew:
Desktop
- [x] Make the top bar (that says “My Voter Guide”) always remain in place, even when you scroll down the page. In mobile, top bar can continue to scroll out of the way as it does currently.
- [ ] Keep the navigation icons in the top right in place when you scroll down the ballot
- [ ] Strangely, when I have google chrome developer tools open, the top bar scrolls up as it does in mobile mode, even when the window is wide enough to display the desktop mode layout.

pertrai1:
Desktop and Mobile is not possible with 2 elements with the current react plugin being used. The Desktop is being pushed now.

pertrai1:
@DaleMcGrew Can we create a backlog milestone and put the second part of this with the submenu :question:

DaleMcGrew:
Done. Second part moved to Issue #132. I also added "Keep the navigation icons in the top right in place when you scroll down the ballot" to the opening description of this Issue.

pertrai1:
@DaleMcGrew So we can go ahead and close this one?

DaleMcGrew:
Hi @pertrai1 there are two refinements I added as checkboxes to the opening description of this Issue.

pertrai1:
@DaleMcGrew I don't think the dev tools is an issue as our users won't be opening dev tools. I have seen this happen on apps that I have worked on in previous projects as the viewport in chrome does not recognize the change when the dev tools is opened below or to the right/left.

In regards to the icons on the top, this is not possible right now. We are literally taking the footer and moving it to the header and that is done with absolute positioning. We can't rely on fixed because it would be against the window and not the container element. When we went to a fixed centered container width, we lost the ability to have the icons fixed. This option will need to become its own story because there will need to have some R&D put towards it.

With that said, I would close this and make sure that the icons have their own story in another milestone. I would estimate at least 1 sprint -2/3 weeks - for this to be researched and developed.

DaleMcGrew:
Thank you Rob for clarifying the challenges to finishing this. I'm bumping this to Version 1.0 (major bug fixes)

pertrai1:
@DaleMcGrew Will we need to focus on the 3rd issue seeing that the user base will not be looking at the app with the dev tools open?

DaleMcGrew:
[...]

PR #128, opened 2016-03-12
#123 - Have the header seen at all times on desktop

pertrai1:
For #123 The Desktop has been fixed but the Desktop and mobile has not been. See comments in the issue for more details.


DaleMcGrew:
:+1:

Issue #124, opened 2016-03-12
Mobile "More" Menu: Make it disappear when you click the hamburger a second time


DaleMcGrew:
1. View 'My Voter Guide' in mobile mode
2. Click the hamburger icon in the upper left of the screen
3. When you click outside of the menu, the menu should go away
- [x] Clicking the hamburger when the menu is open should close the menu
- [x] On an iPhone 6 on Safari, clicking outside the menu did not close it. (Please let me know if you need more details.)

pertrai1:
@DaleMcGrew Take a look at this and test while drinking :coffee:

DaleMcGrew:
I love it! Great work. One minor glitch, the menu doesn't close when you click the hamburger and should.
- [ ] Clicking the hamburger when the menu is open should close the menu

pertrai1:
Ah yes, did not look at that case. Will make sure the icon is not caught behind the body listener

DaleMcGrew:
@pertrai1 I just added one more issue I found on a mobile phone to the opening description.

lisamcho:
I see a lot of issues around the burger menu. I came across this, wondering if it would be easier...
https://github.com/negomi/react-burger-menu

pertrai1:
@lisamcho I love it. You can take this on if you have time or I can. whichever works best for you. Thanks for finding that.

lisamcho:
Today and tomorrow before our v1.0 milestone are kinda busy for me. Punting back to you :)

pertrai1:
Ok the last 3 days have been very busy for me with the first of 3 releases this year for our cloud product and on premise. I will work hard on getting this and others done tonight. @DaleMcGrew @lisamcho

pertrai1:
@DaleMcGrew @lisamcho I am close to having this done but don't think it can be done in the next 8 hours as I am going to get some sleep. This plugin looks good, just have run out of time to make sure it is implemented correctly.

DaleMcGrew:
@pertrai1 waiting on this to see if the design changes being discussed with @ixley will affect how this is implemented.

pertrai1:
@DaleMcGrew please verify

DaleMcGrew:
Hi @pertrai1 this is very nice. Great work! One reproducible issue I am seeing is:
1. Click the hamburger
2. Choose a menu option and it takes you to the page
3. Click the hamburger again -- it doesn't respond. (this is the bug)
4. Click the hamburger again and it works.

PR #129, opened 2016-03-13
#124 - Add ability to hide menu when clicking outside of menu

pertrai1:
When the menu is open in mobile view, if a user clicks anywhere other than the menu, the menu will disappear.


DaleMcGrew:
:+1:

Issue #116, opened 2016-03-01
Voter guide display: Reduce vertical screen real estate


DaleMcGrew:
http://localhost:3000/more/opinions/followed is powered by "routes/More/OpinionsFollowed.jsx"
This loads: /src/js/components/VoterGuide/VoterGuideItem.jsx

- [x] Recapture vertical space, so we can get more Voter Guides displayed above the page fold. Each voter guide should fit into a 100px vertical space. Return to earlier design with 3 "columns". Column 1: image (on the left), Column 2: an area for Name and Twitter, Column 3: Follow and Ignore buttons
The display (by React) of each voter guide item in "GuideList" was refactored from this file, where the design was correct: /src/js/components/VoterGuide/VoterGuideItem.jsx
Wireframe showing vertically-compact display: https://projects.invisionapp.com/share/2R41VR3XW#/screens/89479670

These "More Opinions" pages:
http://localhost:3000/opinions is powered by /src/js/routes/Opinions.jsx
http://localhost:3000/opinions/wv02cand3 is powered by /src/js/routes/Ballot/OpinionsAboutItem.jsx
...load up /src/js/components/VoterGuide/GuideList.jsx
Which then loads /src/js/components/VoterGuide/Organization.jsx

- [x] Refactor these pages to use VoterGuide/VoterGuideItem.jsx (so Organization isn't separated in components/VoterGuide/GuideList.jsx)

pertrai1:
@DaleMcGrew Can you add a wireframe reference to view please?

DaleMcGrew:
@pertrai1 Description text updated with this link: https://projects.invisionapp.com/share/2R41VR3XW#/screens/89479670

pertrai1:
@DaleMcGrew Let me know after reviewing this pull request if I have corrected the appropriate views :ok_hand:

DaleMcGrew:
Hi Rob, thank you for your changes on the "Opinions I'm Following" page:
http://localhost:3000/more/opinions/followed (implemented in "/src/js/components/VoterGuide/VoterGuideItem.jsx") They are the right direction. Here are some comments on your changes:
- [ ] I would like to preserve the icon in mobile mode. It can be smaller in mobile mode, if necessary.
- [ ] To save horizontal space, I think we can take the size of the "Ignore" button down in both Desktop & Mobile. We could also stack Follow and Ignore vertically instead of making them left/right.
- [ ] There is some overlap of the organization name text over the icons.
[...]

PR #130, opened 2016-03-13
#116 - fix DOM to reduce height of each box

pertrai1:
This particular file did not have the correct DOM structure that allowed it to look the way that the other files are structured.


DaleMcGrew:
:+1:

PR #138, opened 2016-03-16
My opinions: stop following and ignore feature + refactoring

lisamcho:
-Made the My Opinions follow/ignore buttons work to remove items per page (consistent with the More Opinions page).
-Refactored GuideStore that the Opinions/MyOpinions page both use to the new style.
-Resolves setState warnings bug on MyOpinions
-Deleted some now deprecated files.
@DaleMcGrew @nf071590


DaleMcGrew:
:+1:  Your changes look good to me, but @nf071590 or @pertrai1 should probably weigh in before it is approved. Great work Lisa!

src/js/actions/GuideActions.js
```diff
@@ -1,24 +1,29 @@
...
-   */
-  follow: function (id) {
-    AppDispatcher.dispatch({
-      actionType: GuideConstants.ORG_FOLLOW, id
-    });
-  }
-};
-
-export default GuideActions;
+var Dispatcher = require("../dispatcher/Dispatcher");
```
pertrai1: preference would be to `import` and `export`. If there is a reason not for doing so, I would say to use `let`

src/js/actions/GuideActions.js
```diff
@@ -1,24 +1,29 @@
...
+
+var GuideActions = (function (_Dispatcher) {
+  function _GuideActions () { }
```
pertrai1: why not use the module factory so you don't have to `new` up?

src/js/actions/GuideActions.js
```diff
@@ -1,24 +1,29 @@
...
+
+  _GuideActions.prototype.ignore = function ignore (we_vote_id) {
```
pertrai1: could you not have the `_GuideActions` object that you can then use `Object.assign` to add to the prototype?

src/js/components/FollowOrIgnore.jsx
```diff
@@ -1,79 +1,26 @@
...
-    else
-      this.props.action.followOrg(this.props.organization_we_vote_id);
-  }
-
-  ignoreOrgLocal (evt) {
-    evt.stopPropagation();
-    this.props.action.ignoreOrg(this.props.organization_we_vote_id);
-  }
+    render () {
+      var ignoreFunc = GuideActions.ignore.bind(this, this.props.organization_we_vote_id);
```
pertrai1: I will only add this one more time so I don't have to do it everywhere ... I think that `let` should be used in place of `var`

src/js/actions/GuideActions.js
lisamcho: Sorry can you clarify what the module factory is? I was copying the format of other actions file that I had seen in the project.  Unless there is a strong reason for having the anonymous function which accepts the dispatcher, we could just instead do the following:
import Dispatcher from "../dispatcher/Dispatcher";

module.exports = {
  ignore: function (we_vote_id) {
    Dispatcher.loadEndpoint("organizationFollowIgnore", { organization_we_vote_id: we_vote_id} );
  },

  follow: function (we_vote_id) {
    Dispatcher.loadEndpoint("organizationFollow", { organization_we_vote_id: we_vote_id} );
  }
....<other functions>
}

pertrai1: One thing is that if this is already being `new`'ed up around the app, I would not expect you to change just for a personal preference that I have. 

The comments I have given are from those personal preferences and commenting on the whole file change example, I do not have a strong reason against the IIFE.

pertrai1:
:white_check_mark:

PR #148, opened 2016-03-19
Link colors inline with the new style guide

pertrai1:
Take note that the hover color does not show much of
a difference from the main color


DaleMcGrew:
:+1:

PR #150, opened 2016-03-20
Fixed browser sync to sync remote devices

pertrai1:
If you are working on the desktop, you can look and see what the external URL is and enter that into the external device and will see that the external device will mimic what you are doing on the desktop.


DaleMcGrew:
:+1:

PR #155, opened 2016-03-24
Cleaned up usability issues. Added/removed links for Version 0.9.

DaleMcGrew:
Made the area a voter can touch much bigger for Links in org Positions on Candidate page, and navigation items. Removed some code we won't use for several versions. Commented out sign in, but left code because several of us may need to use the code soon for testing. Added link to Candidate-specific "More Opinions" page, and built out OpinionsAboutItem.jsx page. (Debugging help needed from Lisa.) Made edits to Location page where we edit address, including adding Focus to the form box on componentDidMount. Removed old "More menu" page. Did some ESLint clean up.


lisamcho:
Looks good to me, except that OpinionsAboutItem doesn't load the items. You initiated an action but didn't configure GuideStore to process the action. Not sure if we should fix it here or merge in then fix.

DaleMcGrew:
I vote for merge and then I'll fix.

lisamcho:
:+1:

PR #157, opened 2016-03-25
Fix fast clicks

lisamcho:
Fixes the bug that if you click really really fast on the support button repeatedly, you can get the count to increase (where it should just toggle up and down 1).


DaleMcGrew:
@lisamcho you will need to approve the updated license (cla) since we updated it with the name "We Vote" since you last approved it. https://www.clahub.com/agreements/wevote/WebApp

DaleMcGrew:
:+1:

PR #158, opened 2016-03-25
Refactor voter store and initial loading of data, redirect user when …

lisamcho:
…ballot is empty, remove cookies, preload opinions

- Refactor addressBox as separate component to avoid duplication between intro and location page.
-Remove setting and reading of cookies for everything except voter_device_id
-Refactor voterStore and initialization of voter/address data to be consistent with other stores
-Preload opinions followed and more opinions along with ballot (in application component so it doesn't matter which page you land on for preloading of data).

### DIFF
--- a/src/js/components/AddressBox.jsx
+++ b/src/js/components/AddressBox.jsx
@@ -1,0 +1,82 @@
+import React, { Component, PropTypes } from "react";
+import { ButtonToolbar } from "react-bootstrap";
+import VoterStore from "../stores/VoterStore";
+import LoadingWheel from "../components/LoadingWheel";
+import VoterActions from "../actions/VoterActions";
+
+export default class AddressBox extends Component {
+  static propTypes = {
+    history: PropTypes.object,
+    saveUrl: PropTypes.string.isRequired
+  };
+
+  constructor (props) {
+      super(props);
+      this.state = { loading: false };
+  }
+
+  componentDidMount () {
+    this.setState({ location: VoterStore.getAddress() });
+    this.listener = VoterStore.addListener(this._onChange.bind(this));
+  }
+
+  componentWillUnmount (){
+    this.listener.remove();
+  }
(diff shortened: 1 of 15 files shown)


DaleMcGrew:
Looks great @lisamcho, I am holding off merging this pull request because I'm thinking that either @pertrai1 or @nf071590 might want to do a quick review first.

src/js/components/AddressBox.jsx
```diff
@@ -0,0 +1,82 @@
...
+      this.setState({ location: VoterStore.getAddress(), loading: false });
+    }
+  }
+
+  _ballotLoaded (){
+    console.log("ballotLoaded");
+    this.props.history.push(this.props.saveUrl);
+  }
+
+  updateLocation (e) {
```
pertrai1: Just from my preference, I like to read code that is clear to know. We all know that e is for event, but I preference is that if you are referring to an event, name it so.

src/js/components/AddressBox.jsx
```diff
@@ -0,0 +1,82 @@
...
+  saveLocation (e) {
+    e.preventDefault();
+    var { location } = this.state;
+    console.log("Saving location", location);
+    VoterActions.saveAddress(location);
+    this.setState({loading: true});
+  }
+
+  render () {
+    var { loading, location } = this.state;
```
pertrai1: Might want to get more into using let so it gets stored yet will not go up the prototype chain as fast as var will. There is more reliable scoping with let and we will be using a rule in linting where no var will not allow for use of var.

pertrai1:
Wow, that was like reading a book that was turned into cliff notes. I can't tell you enough how much joy it is when there is more code removed than code added. Hopefully we can get to a point though where not so much code has to be put into a PR ... for sake of having to put my brain to use.

GREAT work on this refactoring!!!

PR #161, opened 2016-03-26
update changelog with 0.8.0 commits

pertrai1:
This has the changelog addition of the commits made between 0.5 and 0.8 along with the update to documents from `npm run build:docs`


DaleMcGrew:
:+1:

PR #162, opened 2016-03-26
adding a production task, `npm run prod` that avoids browsersync and …

nf071590:
…watching files for changes. also added a timestamp to the browsersync prefix and updated the server task to a module that takes production flag as input.

### DIFF
--- a/server.js
+++ b/server.js
@@ -2,15 +2,15 @@
 const express = require("express");
 const app = express();
 
-const port = 3003;
-const opts = {
-  redirect: true
-};
+module.exports = function (PROD) {
+  const port = PROD ? 3000 : 3003;
+  const opts = { redirect: true };
+  const hostname = PROD ? "wevote.me" : "localhost";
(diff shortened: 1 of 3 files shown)


pertrai1:
@nf071590 Can you sign the new CLA that is on the repo: http://www.clahub.com/agreements/wevote/WebApp. This is part of the build process to check for a person that is doing a PR. Only have to agree to it once.

server.js
```diff
@@ -2,15 +2,15 @@
 const express = require("express");
 const app = express();
 
-const port = 3003;
-const opts = {
-  redirect: true
-};
+module.exports = function (PROD) {
+  const port = PROD ? 3000 : 3003;
```
pertrai1: What is the reason behind the use of `const`?

nf071590: they are static variables that don't change.

pertrai1: cool

PR #164, opened 2016-03-28
Removing deprecated files

lisamcho:
A lot of these files are no longer needed, I left in place the whole folder "Guides" under components which is not being used, but deleted the deprecated files. @DaleMcGrew


DaleMcGrew:
:+1:

PR #173, opened 2016-04-01
Fixing broken image links, bootstrap-map-css warning, address not fou…

lisamcho:
…nd on page reload

-Detects errors on loading opinion images and replaces with placeholder icon if there's a load error
-Gets rid of console warning "Failed to parse SourceMap"
-Fixes issue with reloading an OpinionsAboutCandidate page, where address would not get populated first due to order of civic_id loading.
-Removes use of a deprecated React method (which gets rid of the console warning "Warning: React.findDOMNode is deprecated" in #174

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,7 +26,7 @@
     "keymirror": "^0.1.1",
     "moment": "^2.11.2",
     "object-assign": "^4.0.1",
-    "react": "^0.14.3",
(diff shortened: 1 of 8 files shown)


lisamcho:
Ready for review @DaleMcGrew

DaleMcGrew:
:+1:

PR #175, opened 2016-04-01
Add organization page

lisamcho:
Had to comment out some features that we don't have an API for, but basic page is functioning with org data and positions.


DaleMcGrew:
Awesome!!!!! I'm very close to checking in voterAllPositionsRetrieve.

PR #178, opened 2016-04-01
Added voter guide links. Changed some site text. Fixed README error.

DaleMcGrew:
Added links to among voter guides, candidates and orgs so you can navigate easily. Changed (removed) some site text based on voter testing. Fixed some broken README links.


lisamcho:
:+1:

PR #180, opened 2016-04-01
Fixed Link: limited scope so the "Follow"/"Ignore" buttons aren't affected

DaleMcGrew:
Fixing a bug @lisamcho found in my last Pull Request.


lisamcho:
:+1:

PR #181, opened 2016-04-02
#116 - Reduce vertical display of voter guide boxes

pertrai1:
The mobile view had issues with the content overlapping, so the following was done:

*Make it so that the logo would get smaller when in mobile view
*Redo the columns so they are inline
*Have the buttons stack and show smaller in mobile view


DaleMcGrew:
:+1:

Issue #147, opened 2016-03-19
Install Google Analytics Code in WebApp


DaleMcGrew:
Please note that we will need to add the actual analytics code as an environment variable, as opposed to directly in the code. WeVote.me tracking code: UA-53643204-3
Please let @hdronamraju know when this is done.

This is the analytics code for https://wevote.me (click edit to see)
`<script>
  (function(i,s,o,g,r,a,m){i['GoogleAnalyticsObject']=r;i[r]=i[r]||function(){
  (i[r].q=i[r].q||[]).push(arguments)},i[r].l=1*new Date();a=s.createElement(o),
  m=s.getElementsByTagName(o)[0];a.async=1;a.src=g;m.parentNode.insertBefore(a,m)
  })(window,document,'script','//www.google-analytics.com/analytics.js','ga');

  ga('create', 'UA-53643204-3', 'auto');
  ga('send', 'pageview');

</script>`

This is the code for the c3 site to be created: (click edit to see code)
`<script>
  (function(i,s,o,g,r,a,m){i['GoogleAnalyticsObject']=r;i[r]=i[r]||function(){
  (i[r].q=i[r].q||[]).push(arguments)},i[r].l=1*new Date();a=s.createElement(o),
  m=s.getElementsByTagName(o)[0];a.async=1;a.src=g;m.parentNode.insertBefore(a,m)
  })(window,document,'script','//www.google-analytics.com/analytics.js','ga');

  ga('create', 'UA-75328949-1', 'auto');
  ga('send', 'pageview');

</script>`

pertrai1:
@DaleMcGrew I am not sure what the c3 site is. We can set environment variables but would need to know the process for determining what variable state to use during deployment. Maybe the context of the c3 site will help more.

DaleMcGrew:
Hi @pertrai1 because We Vote will eventually allow political conversation, the network we grow to 10 Million members will need to be running on servers paid for by our 501(c)(4) (due to IRS rules about what a 501(c)(3) can and can't do in order to maintain nonprofit status). This is http://wevote.me.

[...]

PR #182, opened 2016-04-03
#147 Google Analytics for current wevote site

pertrai1:
better idea of what the other site is that will be created.
If this is good enough, close this ticket and open another,
or move this into another milestone when that site will be
created.


DaleMcGrew:
:+1:

PR #184, opened 2016-04-06
Use multiple apis

lisamcho:
This gets rid of calling separate API Endpoints for voter_supports, voter_opposes, support_counts, oppose_counts for each item on the ballot (now uses the multiple api endpoint for each).
Refactors the supportStore to handle the new server responses and make it easy to filter a ballot based on support/oppose status.
Adds a filtered_ballot method to ballot_store (almost ready to link up to the 'Items I support').

Note that starStatuses are still retrieved individually for each item.


DaleMcGrew:
Looks great! I'll merge as soon as the travis check completes. :+1:

PR #185, opened 2016-04-06
Add support ballot

lisamcho:
-Add a filtered_ballot method to BallotStore which calculates a different version of ballot based on SupportStore values.

-upgrade the react router from 1.0 to current version (2.0, which supports query strings), hence we can let the Ballot route take a query search parameter. Note that it's backwards compatible, but we are using some methods that are out of date by 2.0 standards, hence we'll get a couple console warnings until we can update use of deprecated router methods.

-Links up the Ballot component to the new method.


DaleMcGrew:
:+1:

PR #186, opened 2016-04-07
Add Choices remaining ballot

lisamcho:
Add Choices remaining ballot


DaleMcGrew:
Hi @lisamcho it looks like from the code that you are replacing "What I've Starred" with "Choices Remaining". (We will need both.) Thoughts?

DaleMcGrew:
:+1:

PR #190, opened 2016-04-07
Fix bug

lisamcho:
Fixes bug where the opinions for candidate page was overwriting the contents of opinions (general) page. Since opinions are only loaded on either hard page refresh or address change, this was breaking opinions page. 
@DaleMcGrew


DaleMcGrew:
:+1:

PR #191, opened 2016-04-08
Replace multiple API calls with single starAllStatusRetrieve

lisamcho:
Get rid of all the single voterStarStatusRetrieve API calls and replace with a call to the new endpoint that retrieves them all at once.


DaleMcGrew:
:+1:

PR #198, opened 2016-04-10
At this time there will need to be refactoring of

pertrai1:
the ButtonToolbar in order for all buttons to be
consistent accross the app. Until that can happen,
the larger display will continue to have the smaller
buttons on this view.


DaleMcGrew:
👍

Issue #135, opened 2016-03-14
Mobile display too wide, iPhone Safari


DaleMcGrew:
- [x] Many pages are too wide (and offer left/right scrolling) in an iPhone 6 in Safari. (Note: I think our minimum phone width should be iPhone 5) These include Ballot, Candidate, More Opinions
- [x] "1 positive (up arrow)" and "0 negative (down arrow)" should fit on one line (See attached photos, for both ballot page and candidate page)
- [x] We can reduce the candidate photo to ~66% of the current size when in mobile mode.

![mobile-ballot_too_wide](https://cloud.githubusercontent.com/assets/7756031/13753649/05927992-e9d1-11e5-828f-32c9f3c28ead.PNG)
![mobile-candidate_positive_negative](https://cloud.githubusercontent.com/assets/7756031/13753669/1ee7d9b4-e9d1-11e5-9dcb-05f1dc0a8c7e.PNG)

pertrai1:
@DaleMcGrew Are you testing this on the simulator? I have test on iPhone 4 and 5 without any issues except some with candidate photo. Let me know the test procedure you are using for this.

DaleMcGrew:
These screen shots were on an actual iPhone 6. I have noticed a change in the last couple of weeks that solved the problem with the page being too wide on many pages (described after first check box), but occasionally I see a page that is wider than the phone. (I will work on reproducing this.)

The other two check boxes are active issues.

DaleMcGrew:
Hi @pertrai1 I am looking at this now and am still seeing the problems described by all three check boxes. I added the specific pages where I see the width issue. Thank you for taking a look.

pertrai1:
@DaleMcGrew Ok, this should not be an issue to fix with the changes that were now made on #116 that can be used to handle this issue when added. Will work on getting this one done this week, possibly today.

pertrai1:
@DaleMcGrew Please verify fix

DaleMcGrew:
Looks great!

PR #199, opened 2016-04-10
The queries for the change in sizes are added

pertrai1:
so that the app will fit inside of smaller
devices

note that there are some semantic markup fixes
that will be needed once we have a UI that will
take the place of what is currently there. This
could include a major refactor of the React JSX
files.


pertrai1:
This goes with #135

DaleMcGrew:
👍

PR #212, opened 2016-04-15
Fixed glitch with Ballot filtering introduced, relating to missing address

DaleMcGrew:
Fixed glitch with Ballot filtering introduced, relating to missing address. Updated the project About page to give credit based on time volunteered.


pertrai1:
:+1:

PR #214, opened 2016-04-16
Button change from Following to Unfollow and change color

pertrai1:
The button will change to the default red color and the text will change
from Following to Unfollow while the mouse is hovering the button


DaleMcGrew:
👍

PR #215, opened 2016-04-16
Header icons stay in place in desktop mode

pertrai1:
When the viewport is in desktop mode, the icons on the
top will stay in place with the header and subheader. When
the viewport is not desktop, they go away and the icons show
at the bottom footer.

NOTE: Still needs cleanup when UI is put in place


DaleMcGrew:
👍

PR #221, opened 2016-04-20
Add hover to button, bug fixes

lisamcho:
Add a hover state to toggle button.
Rename toggle component FollowToggle, not SupportToggle.
Deprecate a couple more unused files.
Fix bug with opinions on candidate page not being removed immediately.


DaleMcGrew:
👍

PR #222, opened 2016-04-21
Fix candidate reload bug, organize widget files

lisamcho:
This fixes the bug where if you refresh the page on the candidate detail page the support/oppose bar doesn't load.
Also, thought that our widgets were getting unorganized/messy in the root directory and pulled them all into their own folder. 
Pretty sure we're no longer using mock-data anymore either, deprecate!


DaleMcGrew:
👍

PR #223, opened 2016-04-22
Fixed some Javascript errors around undefined variables. Fixed one more "npm test" warning.

DaleMcGrew:
Fixed some Javascript errors around undefined variables. Fixed one more "npm test" warning.

### DIFF
--- a/src/js/routes/Ballot/Ballot.jsx
+++ b/src/js/routes/Ballot/Ballot.jsx
@@ -97,11 +97,11 @@ export default class Ballot extends Component {
     let ballot = this.state.ballot;
     let ballot_props = BallotStore.ballot_properties;
     let ballot_length = 0;
-    if (ballot !== null && ballot.length !== null) {
+    if (typeof ballot !== "undefined" && ballot !== null && ballot.length !== null) {
       ballot_length = ballot.length;
     }
     let ballot_caveat = "";
-    if (ballot_props !== null) {
+    if (typeof ballot_props !== "undefined" && ballot_props !== null) {
       ballot_caveat = ballot_props.ballot_caveat;
     }
 
@@ -129,7 +129,7 @@ export default class Ballot extends Component {
       </div> :
       <div></div>;
 
-    const ballotItems = ballot !== null ?
+    const ballotItems = typeof ballot !== "undefined" && ballot !== null ?
       ballot.map( (item) =>
           <BallotItem key={item.we_vote_id} {...item} />
         ) :
--- a/src/js/routes/Ballot/Ballot.jsx
+++ b/src/js/routes/Ballot/Ballot.jsx
@@ -97,11 +97,11 @@ export default class Ballot extends Component {
     let ballot = this.state.ballot;
     let ballot_props = BallotStore.ballot_properties;
     let ballot_length = 0;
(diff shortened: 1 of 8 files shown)


src/js/routes/Ballot/Ballot.jsx
```diff
@@ -97,11 +97,11 @@ export default class Ballot extends Component {
     let ballot = this.state.ballot;
     let ballot_props = BallotStore.ballot_properties;
     let ballot_length = 0;
-    if (ballot !== null && ballot.length !== null) {
+    if (typeof ballot !== "undefined" && ballot !== null && ballot.length !== null) {
```
lisamcho: This seems unnecessarily verbose. What is the purpose of checking for undefined or null separately? Why not just check for if (ballot && ballot.length)

lisamcho: Or better yet, return the loading wheel if you don't have a ballot. Then you don't have to check any more.

src/js/routes/Ballot/Ballot.jsx
```diff
@@ -129,7 +129,7 @@ export default class Ballot extends Component {
       </div> :
       <div></div>;
 
-    const ballotItems = ballot !== null ?
+    const ballotItems = typeof ballot !== "undefined" && ballot !== null ?
```
lisamcho: ditto

src/js/stores/BallotStore.js
```diff
@@ -77,7 +77,8 @@ class BallotStore extends FluxMapStore {
 
   reduce (state, action) {
 
-    if (action.res.success === false)
+    if (typeof action.res === "undefined" ||
```
lisamcho: I would suggest just getting rid of this line. I can't think of any time when the action.res.success is false and we need to shortcut. The returning of state is covered by the default case.

src/js/stores/SupportStore.js
```diff
@@ -47,8 +47,12 @@ class SupportStore extends FluxMapStore {
   }
 
   reduce (state, action) {
-
-    let we_vote_id = action.res.ballot_item_we_vote_id;
+    let we_vote_id = "";
+    if (typeof action.res !== "undefined" &&
```
lisamcho: ditto

src/js/routes/Ballot/Ballot.jsx
pertrai1: I agree about checkout `(ballot && ballot.length)` and also have to wonder why not just `(ballot.length)` because if that is true, then we know there is a `(ballot)`

DaleMcGrew: I have updated this code per your comments. Just having "ballot.length" threw an error when "ballot" was undefined.

src/js/routes/Ballot/Ballot.jsx
DaleMcGrew: Checking for just "ballot" seems to handle both "ballot !== null" and "ballot !== undefined" situations.

lisamcho:
Looks great! 👍

src/js/routes/Ballot/Ballot.jsx
pertrai1: Ah forgot to switch t hat around in my thinking process during that note

PR #232, opened 2016-04-25
Removed components/Guides files because we don't need them yet

DaleMcGrew:
Removed components/Guides files because we don't need them yet. Trying to get bundle size down for faster load times.


nf071590:
👍

PR #234, opened 2016-04-26
Turn Facebook login back on, remove unnecessary ballot caveat box

lisamcho:
This doesn't remove the FB.login() error (which was a bug already present) but other than that it makes Facebook login working (as before), and adds an app signout function which deletes the cookies and calls the new voterSignOut endpoint.
@DaleMcGrew


DaleMcGrew:
👍  This looks great @lisamcho.

Issue #233, opened 2016-04-25
Two voter_device_id cookies created


DaleMcGrew:
1. Visit ballot. Look in your cookies to make sure you have a voter_device_id
2. Navigate to the Save Address page.
3. Refresh the page, and it creates a second voter_device_id for the path "/settings"

App should only have one voter_device_id

DaleMcGrew:
Note that if verifying this fix on localhost in chrome, it will appear to still be broken:
http://stackoverflow.com/questions/335244/why-does-chrome-ignore-local-jquery-cookies/347997#347997
I verified that this works if we don't use "localhost", here for example: http://127.0.0.1:3000/more/opinions/followed

pertrai1:
@DaleMcGrew your change does the trick for this. closing based on #236

PR #236, opened 2016-04-26
Fix for Issue #233 "Two voter_device_id cookies created"

DaleMcGrew:
Simple change to fix the issue where multiple voter_device_id's get created.


pertrai1:
👍

PR #241, opened 2016-04-27
remove duplicated jquery loading

lisamcho:
Here's another thing that will reduce loading time.
jQuery was being loaded in the vendor files (which I think will get compiled to bundle.js), AND loaded via CDN. Removing it from vendor files greatly reduces size of bundle.js


DaleMcGrew:
Great find @lisamcho! @nf071590 or @pertrai1 do you see any issues with removing jQuery from the bundle?

nf071590:
I also noticed this. We need to rereference dependencies in the app after its removed from the bundle.

DaleMcGrew:
@nf071590 Should we re-reference dependencies in the app prior to approving this pull request?

nf071590:
It looks like @lisamcho did this.

DaleMcGrew:
Cool, thank you @nf071590 👍

nf071590:
Test to make sure all the Ajax requests still work and it can be merged.

lisamcho:
OK, I see a thumbsup so I'm merging.

pertrai1:
@lisamcho great catch. @DaleMcGrew it is a good time now to get our static files being requested from our S3 bucket so we can test the round trip requests. We would also need to implement the CI process when files our put into the S3 bucket. We will probably be using the eastern region of Amazon for an S3 bucket.

@nf071590 @josephevans - I have a preference of hosting our own vendor files instead of CDN versions. Agree or disagree? We can move this to an issue if need be.

Thank you for finding @lisamcho

PR #247, opened 2016-05-02
FB likes

lisamcho:
Add FB share button.
Move FB initialization code to application rather than signin page (since needed for all pages now).
Added basic opengraph (OG) markup to html (needs more to customize what is shown as FB share preview).


DaleMcGrew:
👍  Cool feature! Lisa, your fb_share branch wouldn't let me comment inline, so I'll add my main comment here. We need to hosting WebApp on two sites, so everywhere the domain name "wevote.me" is shown, needs to come from the config.js file. I'll confirm the merge for now so we can fix that in another pull request.

PR #248, opened 2016-05-03
Add loading screen

lisamcho:
Add a basic loading page that shows up before the javascript is loaded. 
Just put a blue page with "WeVote" on it for now. This could definitely use a designer's touch. @ixley

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -15,7 +15,13 @@
     	<meta property="og:image"         content="http://www.wevote.me/wevotelogo.png" />
   </head>
   <body>
-      <div id="app"></div>
+      <div id="app">
+        <div id="loading">
+          <div class="">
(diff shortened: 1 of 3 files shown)


DaleMcGrew:
👍

pertrai1:
@DaleMcGrew @lisamcho I don't understand this splash screen. It loads and then is removed yet I still have the spinner waiting for the candidates to load. Is that the expected behavior?

lisamcho:
Wasn't sure, we could change this to wait for the candidates loaded to
elminate the spinner entirely.

pertrai1:
I do agree but I am just trying to wrap my head around looking at what looks like some what of a "coming soon" splash page, and if it takes more than 5 seconds for something to happen, I might leave. Could you add something like "Loading app...." ??

PR #258, opened 2016-05-09
Remove autofocus on searchbox and prevent zoom

lisamcho:
Prevent zoom, remove autofocus on searchbox (at least for now) to prevent keypad from covering up page


DaleMcGrew:
👍

PR #265, opened 2016-05-16
Position item - (Incomplete)

ixley:
These commits address some minor styling updates and then basic styling for the Position Item component.

** I didn't figure out how to load the icon images, and I wasn't clear whether the functionality exists yet to display org statements (of support or opposition) and for users to like or comment on them. I didn't add any of that markup/styling yet, and it would probably be simplest to stylize only once the functionality is in place.


DaleMcGrew:
Hello @pertrai1 and @lisamcho do you have any bandwidth to do a code review on Jeff's changes? (This week is very heavy with preparation for We Vote presentations for me.)

pertrai1:
@DaleMcGrew yes I can review the code

src/js/components/Ballot/PositionItem.jsx
```diff
@@ -20,19 +20,34 @@ export default class PositionItem extends Component {
 
     let image_placeholder = "";
     if (this.props.speaker_type === "O") {
-        image_placeholder = <i className="icon-org-lg icon-icon-org-placeholder-6-2 icon-org-resting-color"></i>;
+        image_placeholder = <i className="icon-org-lg icon-icon-org-placeholder-6-2 icon-org-resting-color position-item__avatar"></i>;
```
pertrai1: Is all CSS going to be BEM?

src/sass/components/_position.scss
```diff
@@ -1,7 +1,59 @@
...
+
+@include media-object('.position-item', '.position-item__avatar', '.position-item__content');
+
 .position-item {
-  .position-source {
-    color: rgba(171,171,171, 1);
-    font-size: 0.9em;
-    font-weight: 200;
-  }
+  position: relative;
```
pertrai1: Should one class have this many responsibilities?

src/js/components/Ballot/BallotItem.jsx
```diff
@@ -21,9 +21,9 @@ export default class BallotItem extends Component {
   }
 
   render () {
-    return <div className="ballot-item well well-skinny gutter-top--small">
+    return <div className="ballot-item gutter-top--small">
```
pertrai1: Why would the bootstrap well not be used

src/js/components/Ballot/BallotItem.jsx
```diff
@@ -21,9 +21,9 @@ export default class BallotItem extends Component {
...
 
-        <span className="display-name">
+        <span className="display-name h2">
```
pertrai1: I don't follow the class name h2

src/js/components/Ballot/BallotItem.jsx
ixley: I am (somewhat aggressively) trying to remove as much 'visual nesting' of elements as possible. `.well` didn't seem like it was necessary, but I'm open to any push back.

src/js/components/Ballot/BallotItem.jsx
ixley: I was using the Bootstrap class to add some quick styling. Ultimately, I will want to pull heading styles in differently, but it was a quick and easy way to accomplish this styling for the time being.

src/js/components/Ballot/PositionItem.jsx
ixley: I think BEM naming is a helpful way to organize styles – at least for right now. It forces us to focus on the semantics of what components do in a way that just using utility classes doesn't. Ideally, In the near term, I hope that it will help bring more clarity to and unify similar components. In the longer term, I hope to transition to a more React-specific hybridized approach with encapsulated styling, but using BEM naming seems like a good interim approach. Again, I'm still open to alternative suggestions.

src/sass/components/_position.scss
ixley: @pertrai1 Can you be more specific? I'm trying to link names to the components they style. I know there is still some refactoring to be done, but I'm not clear what 'this many responsibilities' refers to.

pertrai1: Sure. This class has the responsibility right now of handling the box model, position, and how it looks. Would it not be better to have those broken apart and possibly change the class name for something more generic that could be reusable?

ixley: I'm okay with breaking up those 'responsibilities' further as it makes sense, but I'd prefer to not do it with utility classes. I would like to eventually get to a more generic pattern, but am okay with having some redundancy of styling between different components in the mean time. This is partly because I'm still wrapping my head around the way things are broken up, assuming there is some degree of refactoring of components that will need to happen anyway (based on brief talks with @DaleMcGrew), and that it still seems like an (iterative) step forward.

@pertrai1 I do still want to move towards more abstraction of patterns and styling, but find this a helpful step. If you think it would be better to adopt more abstraction from the jump, let me know.

PR #267, opened 2016-05-19
Fix to cookie expiration problem. Updated "About" page with volunteer credits.

DaleMcGrew:
Fix to cookie expiration problem. (Setting the cookie expiration year to 9999 seemed to cause problems. I changed this to the year 2200.) Updated "About" page with volunteer credits.


pertrai1:
👍

PR #268, opened 2016-05-20
Converted links to candidates and organizations to use the TwitterHandle if available.

DaleMcGrew:
Converted links to candidates and organizations to use the TwitterHandle if available. (Commented out -- there is a routing issue I will submit an Issue for.) Removed broken candidate-specific opinions link.

### DIFF
--- a/src/js/components/Ballot/CandidateItem.jsx
+++ b/src/js/components/Ballot/CandidateItem.jsx
@@ -14,6 +14,7 @@ export default class Candidate extends Component {
     we_vote_id: PropTypes.string.isRequired,
     twitter_description: PropTypes.string,
     twitter_followers_count: PropTypes.number,
+    twitter_handle: PropTypes.string
   };
 
   render () {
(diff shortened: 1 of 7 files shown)


lisamcho:
@DaleMcGrew What is the benefit of swapping out the links? Having a routing for individual pages loading is good for SEO. But swapping out the links for everything internally will create additional API calls everytime you switch pages within the app and slow it down.

DaleMcGrew:
Hi @lisamcho, just like organizations take "possession" of a URL like "https://twitter.com/defenders" we want them to take possession of "https://wevote.me/defenders". This gives organizations and candidates a quick way to jump their constituents straight to their We Vote page, and will help with adoption.

lisamcho:
@dale, right. We want to route wevote.me/defenders to the Defenders organization page. I absolutely agree. I disagree that we should change all internal links within the app to use /defenders rather than /wevoteid. If you note on the NotFound page, it has a loading screen with a spinner while it is fetching the we_vote_id. 

If we change all the internal links, we will create this additional delay anytime you're switching pages.

DaleMcGrew:
Hi @lisamcho I hear you about introducing an API call and delay. The reason to change all of the internal links as https://wevote.me/TWITTER_HANDLE is to a) educate voters/organizations that they can find anyone by their twitter handle, and b) so that when the url is copied and emailed around the internet, it is human readable and understandable. I think the delay is worth it in this case.

lisamcho:
@DaleMcGrew I am also wondering if it would be better to do this on the backend. Twitter handles are unique. We don't technically need we_vote_ids for anything other than as a unique identifier right? Why not replace all we_vote_ids with twitter handles when an organization or candidate has one on the backend and otherwise keep it as a wv12345 number? That would avoid the performance issue and create friendly urls without the need for a notfound page.

DaleMcGrew:
I like the idea of retrieving candidates directly with ballotItemRetrieve by we_vote_id OR twitter_handle (or Vote Smart Id, or maplight_id, etc.), or organizations by twitter_handle with organizationRetrieve.

One way to avoid calling TwitterIdentityRetrieve, would be to create a TwitterHandleStore that aggregated the links between TwitterHandles and we_vote_ids (from ballotItemOptionsRetrieve, or other lists of candidates or orgs), so as the voter moves around the app we don't have to hit the server for each TwitterHandle.

The way the data comes in, we need to maintain a candidate entry (which changes from election to election) as well a politician entry (which follows the person for life). So we have 2+ entries for each candidate/politician. Each candidate / politician has at least 8 identifiers used across a variety of systems, so our strategy has been to be able to search against any of them, including Twitter handle.

lisamcho:
Hmmm, I guess a lot of ways to do this. Better to discuss on Saturday?

DaleMcGrew:
That sounds good.

DaleMcGrew:
I will resubmit in a different form later.

PR #273, opened 2016-05-20
Bookmarks

lisamcho:
Add bookmarks page that links to candidate page for candidates or else the ballot with an anchor tag for measures and offices.
Fix positioning of star for ballot items (was hidden behind stuff).


DaleMcGrew:
👍

Issue #272, opened 2016-05-20
Abbreviate Twitter follower counts


ixley:
Let's abbreviate Twitter follower counts like Twitter does. Basically, aim for the count to not go beyond 4 characters (excluding commas/decimals). The exact count can be kept – perhaps for 'title' to show on hover if it seems useful.

< 10,000 - add commas (5,035)
10,000 – 99,999 - round to one decimal place (45.8K)
100,000 – 999,999 - round to nearest thousand (847K)
1,000,000 – ?? - round to ten-thousand (1.45M)

lisamcho:
Hmmm, anyone know any good libraries that will take care of this for us? @DaleMcGrew @pertrai1 @nf071590

pertrai1:
@lisamcho @ixley @DaleMcGrew 
```
function nFormatter(num) {
     if (num >= 1000000000) {
        return (num / 1000000000).toFixed(1).replace(/\.0$/, '') + 'G';
     }
     if (num >= 1000000) {
        return (num / 1000000).toFixed(1).replace(/\.0$/, '') + 'M';
     }
     if (num >= 1000) {
        return (num / 1000).toFixed(1).replace(/\.0$/, '') + 'K';
     }
     return num;
}
```

`nFormatter(2500)` => 2.5k

ixley:
Awesome - thanks @pertrai1. I'll make some adjustments and attempt to add this.

PR #276, opened 2016-05-21
Change jsx to appropriate structure on candidate card including itema…

lisamcho:
…ctionbar and itemsupportopposecounts

**This is WIP, not ready for merging with master.**
I have changed the jsx on the candidate card, itemActionBar and itemSupportOpposeCounts so that @ixley can more easily style them.

Note: I'm missing one item on the candidate card marked as TODO because it will require some refactoring, but this should get you started at least. 

I also commented out a lot of inline styling. Please delete if you're not going to use it.


ixley:
@lisamcho this looks great! Per Rob's suggestion, I'm going to try working from your branch to style this before we merge back to the main repo.

lisamcho:
@ixley Merged in your commits and added some more of my own on top of it.

lisamcho:
@ixley I'm copying over your comments from the PR on the repo on my github account since that pull request is now closed...
need some additional clarification....

Return a variable that states whether the majority supports or opposes a candidate (in ItemSupportOpposeCounts). If it’s 50/50, show as support. (This is to show the color icon for the majority position.) I thought this would be simple enough that I could do it, but I wasn’t able to get it working
**_So there's a function percentageMajority on this page, that you can just check if it's >= 50. Do you want to assign a separate boolean variable like isMajoritySupport? That would just be:
const isMajoritySupport = this.percentageMajority() >=50;_** 

Pass back value for whether user supports or opposes (to show thumb icon at top). I know this is in the ItemActionbar, so we just need it in CandidateItem too.
**Done! **

Show the bar container (‘well’) with zero counts even if there are no positions taken in your network. Counts for support and oppose will both be ‘0’ and there won’t be a filled in bar - just the gray background
**Did you talk with @DaleMcGrew ? He implemented the code that purposely hides the counts if there's no positions. I said I didn't like the UX of having it appear/disappear when you click support/oppose, he said he wanted to remove unnecessary clutter.**

If possible, set up the same markup/functionality on the Candidate page/route as well. It will look very similar to the Candidate Card.
**For purposes of modularity, just reused the CandidateItem component on the Candidate route. The only difference I saw was the (read more) link. If there ends up being lots more differences let me know and we can split them out again.

wevote#272 is also relevant if you want to grab it.

I also wasn't sure what the most appropriate way to add basic dropdown button functionality was. I can use Bootstrap's built-in function, but if there's a cleaner, more appropriate way you know of, maybe you could add something for the selected position button state?
**I don't know a cleaner way, bootstrap sounds good to me.**

ixley:
@lisamcho Great! Responses below...

I thought it would be cleaner to have a boolean check, so I'll try your suggestion of `const isMajoritySupport = this.percentageMajority() >=50;`. (Some of this stuff I have an idea what the logic should be, just not how to code it, so thanks!)

I think with this design, it does make sense to show the empty bar well – at least for now. It makes the transition more fluid when the user selects a position, and if we want to look at designing a different empty state we can address it later. Sound okay, @DaleMcGrew?

re https://github.com/wevote/WebApp/issues/272 - I can take a stab at the number formatting function that Rob posted, but you can also add this if you want since just knowing where to place things will be more obvious to you. Don't use the conditions from the snippet as is, though – reference the logic I listed, as it varies a bit. There will also be no need for numbers in the billions.

I tried using Bootstrap for the dropdown functionality, but don't think it's going to work as is with React. There is a [React-Bootstrap](https://react-bootstrap.github.io/) project (not yet at v1) but it seems excessive just for this functionality, and probably unnecessary otherwise. For now, maybe the simplest thing is to just create a state that toggles whether the drop-down menu is open or closed (could be an inline style to display or hide), but again, I'm not sure what the best approach is. Thoughts?

DaleMcGrew:
@ixley & @lisamcho I am ok with showing the support/oppose bar for now when there are no support or oppose counts. We can also play with the support/oppose bar being visible on the Candidate page, but not visible on the ballot listing (where vertical space is even more at a premium).

lisamcho:
replacing with a new pull request

PR #286, opened 2016-05-27
UX Language updates based on user testing.

DaleMcGrew:
UX Language updates based on user testing. The only file requiring code review is src/sass/layout/_header.scss for @ixley


DaleMcGrew:
I am approving so we can do further user testing today. 👍

PR #287, opened 2016-05-27
Turning on internal twitter handles for candidates.

DaleMcGrew:
Turning on internal twitter handles for candidates. We aren't using organizational twitter handles for internal links yet due to bug with switching between two twitter handles on NotFound.jsx page.


DaleMcGrew:
I need to do some work on a couple of these pages to get over the hurdles Jeff is facing, so I'm going to approve this pull request now. 👍

Issue #278, opened 2016-05-23
Show position in OrganizationPositionItem just like PositionItem


ixley:
Add position info to OrganizationPositionItem. Currently, it shows a rating without a timestamped year, but it doesn't show explicit positions. Show full position info with the same logic as in PositionItem.jsx. 

Bonus points if you can put the code for .explicit-position in a component (like PositionRatingSnippet) and modularize the whole thing.

DaleMcGrew:
@ixley please verify fix once pull request #288 is approved.

ixley:
I haven't been able to check this because nothing is showing up with explicit positions (only ratings). Were you seeing thumbed positions when you tested this?

PR #288, opened 2016-05-28
I got Support/Oppose moved into PositionSupportOpposeSnippet Widget.

DaleMcGrew:
Hello @ixley this is a solution to issue #278. Some of the variables can be renamed and some style work is needed, but it is in good shape.

### DIFF
--- a/src/js/components/VoterGuide/OrganizationPositionItem.jsx
+++ b/src/js/components/VoterGuide/OrganizationPositionItem.jsx
@@ -2,6 +2,7 @@ import StarAction from "../../components/Widgets/StarAction";
 import React, { Component, PropTypes } from "react";
 import { Link } from "react-router";
 import PositionRatingSnippet from "../../components/Widgets/PositionRatingSnippet";
+import PositionSupportOpposeSnippet from "../../components/Widgets/PositionSupportOpposeSnippet";
 
 export default class OrganizationPositionItem extends Component {
   static propTypes = {
@@ -9,6 +10,7 @@ export default class OrganizationPositionItem extends Component {
   };
 
   render (){
+    var position = this.props.position;
     let { ballot_item_display_name,
       kind_of_ballot_item,
       vote_smart_rating,
@@ -40,23 +42,10 @@ export default class OrganizationPositionItem extends Component {
             <span className="position-rating__candidate-name">{ballot_item_display_name}</span>
           </Link>
           
-          {/* wasn't sure how to get this to work - Jeff */}
           {/* show explicit position, if available, otherwise show rating */}
-          
-          {/* make into component */}
-          <div>See issue #278</div>
-          {/* if (position.is_support || position.is_oppose) {
-            org_position = <div className="explicit-position">
-              { position.is_support ? <img src="/img/global/icons/thumbs-up-color-icon.svg" width="20" height="20" className="explicit-position__icon" alt="Supports" /> : "[oppose icon]" }
-              <div className="explicit-position__text">
-                <span className="explicit-position__position-label">
-                  { position.is_support ? "Supports" : "Opposes" }
-                </span>
-                <span> {this.props.candidate_display_name}</span>
-              </div>
-            </div>);
-          } else if (vote_smart_rating) */}
-          <PositionRatingSnippet rating = {vote_smart_rating} rating_time_span = {vote_smart_time_span} />
+          { vote_smart_rating ?
+          <PositionRatingSnippet {...position} /> :
+              <PositionSupportOpposeSnippet {...position} /> }
 
         </div>
         {/*Running for {office_display_name}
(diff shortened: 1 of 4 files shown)


src/js/components/VoterGuide/OrganizationPositionItem.jsx
```diff
@@ -40,23 +42,10 @@ export default class OrganizationPositionItem extends Component {
...
-              <div className="explicit-position__text">
-                <span className="explicit-position__position-label">
-                  { position.is_support ? "Supports" : "Opposes" }
-                </span>
-                <span> {this.props.candidate_display_name}</span>
-              </div>
-            </div>);
-          } else if (vote_smart_rating) */}
-          <PositionRatingSnippet rating = {vote_smart_rating} rating_time_span = {vote_smart_time_span} />
+          { vote_smart_rating ?
```
ixley: @DaleMcGrew I thought this was supposed to be the other way around... If there is an explicit position, show that, otherwise show the Vote Smart Rating. Won't orgs with explicit positions also commonly have Vote Smart ratings?

src/js/components/Widgets/PositionSupportOpposeSnippet.jsx
```diff
@@ -0,0 +1,61 @@
...
+      className = "explicit-position__icon";
+      alt = "Opposes";
+      positionLabel = "Opposes";
+      isSupportedBy = "is Opposed by"
+    } else {
+      src = "/img/global/icons/mixed-rating-icon.svg";
+      className = "position-rating__icon position-rating__icon--mixed";
+      alt = "Neutral Rating";
+      positionLabel = "Neutral Rating of";
+      isSupportedBy = "is viewed Neutrally by"
```
ixley: In what context would this get used? I don't remember discussing neutral positions among what I've been referring to as explicit positions (support vs oppose).

ixley:
I'm not really qualified to catch any code issues that may be present, but it looks good to me. Added a couple questions inline.

pertrai1:
👍  This can be merged based on the code working. Any changes to the semantics of naming conventions can be made by @ixley  in another PR.

ixley:
thanks @pertrai1 What's the best way to pull down someone else's fork/branch to review before a PR gets merged? I was able to clone Lisa's forked repo previously when we were working together but got permission errors when trying to clone Dale's branch to look at this. I'm assuming there's a more straightforward way?

pertrai1:
I will add steps here today when I get back on laptop so you can pull from other repo branches. 
Rob 


Sent via my Samsung Galaxy Note® Edge

-------- Original message --------

ixley:
@pertrai1 FYI - I was able to pull from Dale's branch after all (my SSH connection was just blocked initially). Would probably still be good to post instructions about pulling from other people's repos, but if that's the recommended practice, I can do that – just wanted to check.

pertrai1:
@ixley - I have added documentation for doing `fetch` and `pull` from a remote user branch that is based on a PR. https://github.com/wevote/WebApp/commit/6c993a87d40f6ae1973bebbe958e0f5cf16ed2c9

src/js/components/VoterGuide/OrganizationPositionItem.jsx
DaleMcGrew: Hi Jeff, if an organization has a Vote Smart Rating AND an endorsement, they are stored as separate position entries. The API server will serve only the support/oppose position entry if there is a support/oppose position entry AND rating for a candidate. So this logic on this line will never have to "decide" to show (or prioritize) a rating vs. an explicit endorsement -- the switch only reacts the kind of position it is.

src/js/components/Widgets/PositionSupportOpposeSnippet.jsx
DaleMcGrew: Hi @ixley, an example of a neutral "information only" rating would be KQED (or other nonpartisan organization) providing analysis, but no endorsement or opposition.

ixley: Hmm... Okay. Semantically, it still feels different than the (explicit)
position ratings. I understand that this might be a simpler way to execute,
but I find the grouping to be a bit confusing (FWIW).

src/js/components/VoterGuide/OrganizationPositionItem.jsx
ixley: Okay, that makes sense. But does that also conversely mean that `vote_smart_rating` will only be true if there is no support/oppose position? If _not_, we would still want to check for that first, right? I will have to look at your sample org later but I was not seeing any explicit positions (support/oppose) show up.

DaleMcGrew: There should not be a vote_smart_rating when there is a support/oppose position. In general, I prefer to set up logic to look for cases that need specific data *first*, which is why I look for vote_smart_rating, and if not found, then look for the support/oppose/neutral position.

ixley: Okay - helpful to know. When you get a chance, let me know if things are displaying as intended on your end and I'll take another look tonight.

PR #289, opened 2016-05-28
Prefix Bootstrap classes

### LINKED ISSUE #259: Prefix [Bootstrap] classes
Since we are currently using Bootstrap but will also be using custom utility classes (along with everything else), let's prefix all Bootstrap classes for clarity, and namespace other classes, as appropriate. (This may just be a stopgap if we eventually move to more encapsulated styles, but it still seems helpful.)

**Does anyone know of a good package or tool to prefix classes** by path? We only need something to rename premade classes like with Bootstrap.

Examples:
`tb-*` for all Bootstrap classes
`u-*` for custom utility classes
`track-*`, or `analytics-*` if class names get used as hooks for event tracking
`[component-name]` for all components or reusable elements (BEM naming) - not a prefix, but consistent naming convention

ixley:
This pull request:
1. Adds the prefix `bs-` for all Bootstrap classes. I spot checked to a reasonable extent but there may be places I missed. Prefixing these classes now makes it immediately obvious what classes are Bootstrap utility classes for better maintainability.
2. Switches to a We Vote hosted version of Bootstrap instead of the CDN that was previously pointed to.
3. Updates the Prop `bsClass` for all React-Bootstrap components. This will now need to be defined for any React-Bootstrap component to add the correct prefixed classes. [Ref](https://react-bootstrap.github.io/components.html#buttons-props)
4. Adds [Autoprefixer](https://github.com/postcss/autoprefixer/blob/master/README.md) (to prefix add/remove relevant (css selector) vendor prefixes.
5. Adds [Source Maps](https://www.npmjs.com/package/gulp-sourcemaps) for sass files. This can be used for js too if anyone wants.

Bootstrap Social classes are *not* included, so this update may upset some styling for anything related to Bootstrap Social. It seemed like it is only minimally used, but if it's important, we can update it as well.

*closes #259


pertrai1:
👍

Issue #279, opened 2016-05-23
Add org position to unfollowed orgs on Candidate page


ixley:
Currently, unfollowed orgs with positions on a candidate will show up below followed orgs on the candidate page, but without their positions. Show positions for unfollowed orgs too, beneath their twitter bio. (organization.jsx)

@DaleMcGrew can you confirm this is what you want?

DaleMcGrew:
That is correct. It is helpful to see an organization's opinions about a
candidate before you follow it. The variables with the is_support /
is_positive _rating etc are already in the Organization Jsx card /
component, ready to style.

-Dale

On Monday, May 23, 2016, Jeff French <notifications@github.com> wrote:

> Currently, unfollowed orgs with positions on a candidate will show up
> below followed orgs on the candidate page, but without their positions.
> Show positions for unfollowed orgs too, beneath their twitter bio.
>
> @DaleMcGrew <https://github.com/DaleMcGrew> can you confirm this is what
> you want?
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly or view it on GitHub
> <https://github.com/wevote/WebApp/issues/279>
>


-- 

Dale John McGrew
Co-founder, We Vote USA
dale.mcgrew@wevoteusa.org
www.WeVoteUSA.org <http://wevoteusa.org/>
415-516-8777
Twitter: @WeVoteUSA <https://twitter.com/WeVoteUSA> & @DaleMcGrew
<https://twitter.com/dalemcgrew>
Protecting the interests of voters. Fueled by passion.

DaleMcGrew:
@ixley you should be able to drop in this code now:
          { vote_smart_rating ?
          <PositionRatingSnippet {...position} /> :
              <PositionSupportOpposeSnippet {...position} /> }
I am about to lose internet connection or else I would do it!

ixley:
I reordered the condition to be
```
{ position.is_support || position.is_oppose ?
    <PositionSupportOpposeSnippet {...position} /> :
    <PositionRatingSnippet {...position} />
}
```
but it doesn't seem like it's quite working because everything is currently showing up as ratings (I can't find any explicit positions). It's hard for me to check because I wasn't seeing many explicit positions before, but the ones I checked are all ratings now.

DaleMcGrew:
[...]

PR #293, opened 2016-05-31
Added a position display snippet for Information Only. Added position…

DaleMcGrew:
… display to voterGuidesToFollowDisplay related to Issue #279 and Issue #278. Is a good place for @ixley to verify styles. I recommend checking out http://localhost:3000/progressivekick and then navigating to Kamala Harris page to see the Progressive Kick position from that page.


DaleMcGrew:
👍

PR #295, opened 2016-06-01
updated media query mixin + small style adjustment

ixley:
Added a more robust mixin for handling media queries and adjusted ballot item section headings for small screens


DaleMcGrew:
👍

Issue #224, opened 2016-04-22
Ballot not loaded unless you click "Save Address"


DaleMcGrew:
1. Make sure you are connected to API server (as opposed to local WeVoteServer)
2. Clear "voter_device_id" from cookies
3. Visit WebApp, and you are redirected to intro page: http://localhost:3000/intro
4. You should see an address in the top navigation bar. Do NOT click "Save Address"
5. Click on the "Ballot" icon, and you are taken to an empty ballot page.

Proposed solution: If we get to the ballot page and the ballot is empty, reach out to API server. (voterBallotItemsRetrieve? voterAddressSave?)

lisamcho:
Can't reproduce. I get a blank address in top navigation bar in step #4.

DaleMcGrew:
@lisamcho Are you on Wifi connected to a building, or on a mobile device via cell phone service?

lisamcho:
Wifi on a building.

DaleMcGrew:
Do you get a geo-located address when you use https://wevote.me as opposed to your localhost? We might also need to update our GeoIP address database -- it sounds like your wifi device might have been assigned an IP address we don't recognize.

lisamcho:
I don't get a geo-located address for either wevote.me or localhost. I used to always get "none, none, none". Now I always get blank.

lisamcho:
I think that this bug is caused by the voterAddressRetrieve endpoint returning a bogus google_civic_id....see response below.
{address_found: true, normalized_state: "", normalized_line1: "", guess_if_no_address_saved: true, text_for_map_search: "San Francisco, CA 94110"…}
address_found
:
true
address_type
:
"B"
google_civic_election_id
:
"San Francisco, CA 94110"
guess_if_no_address_saved
:
true
latitude
:
""
longitude
:
""
normalized_city
:
""
normalized_line1
:
""
normalized_line2
:
""
normalized_state
:
""
normalized_zip
:
""
status
:
"VOTER_ADDRESS_NOT_FOUND, GUESS_IF_NO_ADDRESS_SAVED, VOTER_ADDRESS_RETRIEVE-VOTER_LOCATION_FOUND_FROM_IP *** San Francisco, CA 94110 ***, , GOOGLE_CIVIC_API_ERROR: Failed to parse address, VOTER_ADDRESS_RETRIEVE-ADDRESS_FOUND"
success
:
true
text_for_map_search
:
"San Francisco, CA 94110"
voter_device_id
:
[...]

PR #296, opened 2016-06-02
A solution to issue #224, 'Ballot not loaded unless you click "Save A…

DaleMcGrew:
…ddress"'. This code needs to be refactored in order to avoid multiple calls to voterBallotItemsRetrieve on some pages. I have left console.log code in place so we can pick up on the refactoring/optimization work when the delivery schedule permits.

### DIFF
--- a/src/js/stores/BallotStore.js
+++ b/src/js/stores/BallotStore.js
@@ -110,14 +110,20 @@ class BallotStore extends FluxMapStore {
     let newBallot = {};
 
     switch (action.type) {
+      // Dale 2016-06-02 On some pages we are calling BallotActions.retrieve() as many as three times in order to
+      // avoid problems like Issue #224. This should be optimized.
 
       case "voterRetrieve":
+        // console.log("BallotStore, voterRetrieve");
+        // We need to refresh the ballot here (in addition to voterAddressRetrieve) because there are routes that you
+        // can go to directly (e.g. wevote.me/ballot) that don't reach out and retrieve the ballot items any other way.
         BallotActions.retrieve();
         return state;
 
       case "voterBallotItemsRetrieve":
         key = action.res.google_civic_election_id;
         newBallot[key] = action.res;
+        // console.log("BallotStore, voterBallotItemsRetrieve, key: " + key);
 
         return {
           ...state,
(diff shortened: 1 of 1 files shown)


src/js/stores/BallotStore.js
```diff
@@ -110,14 +110,20 @@ class BallotStore extends FluxMapStore {
...
 
     switch (action.type) {
+      // Dale 2016-06-02 On some pages we are calling BallotActions.retrieve() as many as three times in order to
+      // avoid problems like Issue #224. This should be optimized.
 
       case "voterRetrieve":
+        // console.log("BallotStore, voterRetrieve");
+        // We need to refresh the ballot here (in addition to voterAddressRetrieve) because there are routes that you
+        // can go to directly (e.g. wevote.me/ballot) that don't reach out and retrieve the ballot items any other way.
         BallotActions.retrieve();
```
lisamcho: Hmm, did you try the code without this? I don't think it should matter which route you're on. voterRetrieve is called by the application, then voterAddressRetrieve is called, and ballotItems shouldn't be retrieved twice.

DaleMcGrew:
I did test it without the BallotActions.retrieve() under "voterRetrieve", and I found a couple of cases where no ballot was returned. I am sure that after this milestone we can review the retrieve dependencies and optimize the code to only make the voterBallotItemsRetrieve call once. I am glad to have a solution that solves the problem from the perspective of the voter, even though it requires extra calls to the same API endpoint in certain circumstances.

src/js/stores/BallotStore.js
```diff
@@ -127,13 +133,26 @@ class BallotStore extends FluxMapStore {
       case "voterAddressSave":
         key = action.res.google_civic_election_id;
         newBallot[key] = action.res;
+        // console.log("BallotStore, voterAddressSave, key: " + key);
 
+        // We need to refresh the ballot here after the address is changed. If we don't, it is possible to change
+        // the address and be left with the ballot for the prior address.
+        BallotActions.retrieve();
```
lisamcho: I don't think that you should change this either. The voterAddressSave endpoint returns the new ballot from the new address, so the currently the BallotStore is saving the new ballot information. There's no need to retrieve the ballot again.

src/js/stores/BallotStore.js
DaleMcGrew: I did test it without the BallotActions.retrieve() under "voterRetrieve", and I found a couple of cases where no ballot was returned. I am sure that after this milestone we can review the retrieve dependencies and optimize the code to only make the voterBallotItemsRetrieve call once. I am glad to have a solution that solves the problem from the perspective of the voter, even though it requires extra calls to the same API endpoint in certain circumstances.

src/js/stores/BallotStore.js
DaleMcGrew: I think we need to talk through the dependencies in BallotStore to reach the real solution to this issue, and that requires that we have a conversation. Unfortunately given the deadline and all of our schedules, at the earliest we will be able to have that conversation Saturday morning. I think it is very risky to leave this highly visible bug in place until Saturday morning given we will be showing the app to a number of new people early Saturday morning at NDoCH. I think we should work with the solution in this pull request, and then treat the next step as optimization, refactoring so that we don't need to make a second or third call in certain circumstances. I am not a fan of the extra API calls included in this pull request, but what I have checked in works in all of the situations I tested. Removing any of these extra "BallotActions.retrieve();" lines causes the issue to be apparent to a voter in at least one of the cases I tested.

PR #298, opened 2016-06-03
Refactor ballot page logic

lisamcho:
@DaleMcGrew I looked into the ballot bug some more and found some weird logic on the ballot component that may have been contributing to the weird behavior. Seems to work on my machine. See if it works on yours and hopefully we can avoid the messy 3x api calls?


DaleMcGrew:
👍  That passed all the tests I could think of on my local machine. Great work @lisamcho!

PR #302, opened 2016-06-03
Add dropdown component, make positions stay on page unless ignored.

lisamcho:
-Add dropdown
@ixley I think that this is basic functionality for a dropdown. Should I leave the styling in your expert hands from here?

-On both OpinionsToFollow page and Candidate pages, make each organization position toggle and stay on the page (not disappear) when you click Follow, but disappear when you click "Ignore"


DaleMcGrew:
:+1

PR #305, opened 2016-06-09
Fix double-clicking

lisamcho:
When quickly double-clicking on support/oppose, make sure you only send one request.


DaleMcGrew:
👍

PR #312, opened 2016-06-18
fix React Burger Menu functionality, requires some hacks, resolves #309

### LINKED ISSUE #309: Header link areas not working, top ~18px
In desktop mode, try to click any icon or link in the top 18-20px of the browser display area. There seems to be something blocking all links / link areas from working.

ixley:
The styles and some of the functionality for the 'burger menu' were a bit wonky and required some cleanup. The header was essentially getting covered by unconstrained elements from the burger menu that were bleeding out and blocking nav items. The fix still relies on some hacks because the markup is not structured quite as intended for this plugin, but it will hold for now. We can better address it when we more fully redesign the nav, header, etc.


DaleMcGrew:
👍

PR #313, opened 2016-06-19
fix headroom styling

ixley:
just a few tweaks to fix headroom styles along with html and body height


DaleMcGrew:
👍 Thanks for taking care of this Jeff!

PR #321, opened 2016-07-01
Sign in with Twitter, verify that you can manage an organization, candidate or new Twitter entry

DaleMcGrew:
Sign in with Twitter, verify that you can manage an organization, candidate or new Twitter entry. Fixed several eslint warnings. Started adding code to allow organizations to edit their existing positions once they have authenticated with Twitter.


DaleMcGrew:
👍 I am approving this myself so I can do Twitter and Facebook testing on the demo server https://wevote.me

PR #326, opened 2016-07-05
First and very rough draft of instant search

DaleMcGrew:
First and very rough draft of instant search. Works with the new API endpoint searchAll. /src/js/components/SearchAllBox.js needs styling work, and changes for many search situations, but it is a solid start.


pertrai1:
👍

DaleMcGrew:
Thank you @pertrai1!

PR #329, opened 2016-07-09
Removing src/js/config.js from github tracking

DaleMcGrew:
Removing src/js/config.js from github tracking since it should be under the control of each developer. New developers can copy src/js/config-template.js. Updated installation documentation too.


DaleMcGrew:
👍

PR #335, opened 2016-07-11
Added starting files for Measure and Office. Added organizational positions for current election AND other elections.

DaleMcGrew:
Added starting files for Measure and Office, so clicking through some search results won't throw error. Added split between organizational positions for the voter's current election vs. organizational positions for all other elections.


DaleMcGrew:
I am approving my own pull request because we need to test it on the live server. 👍

Issue #332, opened 2016-07-09
Instant search: When you click outside of the search drop-down, close the drop down. 


DaleMcGrew:
Instant search box code is here: /src/js/components/SearchAllBox.js
I'm hoping we can mimic the behavior of the hamburger, slide out navigation, where the left navigation closes when you click outside of the navigation menu.

DaleMcGrew:
@pertrai1, any attempt I have made to hide the "search-container" drop-down with the "onBlur" action in SearchAllBox.js has led to the Link not taking the view to the candidate or measure page (that is, when you search for a candidate or organization, and click on a candidate's or organization's name). We need to figure out another way to hide the "search-container" drop-down when you click outside of that drop-down area. Please see my most recent fix in pull request #338.

pertrai1:
@DaleMcGrew Ok, I will change the event to the search container or body and not on the search input field. It really does not belong on the input anyways. I also have some changes that I have seen where the input is right in Firefox but not in Chrome. I am off until Thursday so I will have some time to try to get the search issues wrapped up.

DaleMcGrew:
Thank you @pertrai1!

DaleMcGrew:
Nailed it! Thank you @pertrai1.

PR #336, opened 2016-07-11
332 Removing search results on blur

pertrai1:
When there is a user click event when results are showing from the
search results anywhere on the page, the search results will be removed
from being shown and the value is removed from the search input box.


pertrai1:
The PR is for #332

DaleMcGrew:
👍

PR #338, opened 2016-07-11
The onSearchBlur code was breaking the ability to click on a selection in the Instant search drop-down.

DaleMcGrew:
The onSearchBlur code was breaking the ability to click on a selection in the Instant search drop-down. Moved the code that shows and hides Instant search results into utils/search-functions, and we are now calling the "exitSearch" code once we arrive at Candidate or PositionList routes.

@pertrai1, any attempt I have made to hide the "search-container" drop-down with the "onBlur" action in SearchAllBox.js has led to the Link not taking the view to the candidate or measure page. We need to figure out another way to hide the "search-container" drop-down when you click outside of that drop-down area.

### DIFF
--- a/src/js/components/SearchAllBox.js
+++ b/src/js/components/SearchAllBox.js
@@ -2,6 +2,7 @@ import React, {Component, PropTypes} from "react";
 import { Link } from "react-router";
 import SearchAllActions from "../actions/SearchAllActions";
 import SearchAllStore from "../stores/SearchAllStore";
+import { enterSearch } from "../utils/search-functions";
 
 export default class SearchAllBox extends Component {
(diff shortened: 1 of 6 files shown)


pertrai1:
Good catch. Did not check that use case when adding this. Good job of refactoring. I will look and see if we can prevent the capture phase and bubbling so that the search container links are not effected by this. I will merge and pull latest for testing.

pertrai1:
@DaleMcGrew I don't have write access for the repo any more to merge code :+1:

DaleMcGrew:
@pertrai1 You should have access now

pertrai1:
:+1:

PR #339, opened 2016-07-11
Now supporting two candidate searches in a row, and two organization searches in a row.

DaleMcGrew:
Now supporting two candidate searches in a row, and two organization searches in a row. Needed for Instant search to work in more cases.


DaleMcGrew:
I am approving this so we can test on mobile devices. 👍

PR #340, opened 2016-07-14
Instant search results close on body click

pertrai1:
This gets us going in the right direction. Still needs to have major refactoring but it gets you going in the right direction for the presentation this weekend. I can do more before the presentation if you need more.

references: #332


DaleMcGrew:
👍

Issue #341, opened 2016-07-14
Instant Search: Box expand/contract behavior


DaleMcGrew:
- [x] When browser is wide enough to show top navigation icons, the "My Voter Guide demo" text should not be hidden when you click in the search box.

DaleMcGrew:
👍  Nailed it, thank you!

PR #343, opened 2016-07-15
Clean up of specific instant search issues

pertrai1:
Please check to see if the issues are resolved. Because of the need to refactor this, there might be some edge cases there were not taken care of. Use the happy path of showing in Google Chrome if can be.

Issues: #341 and #343 

@DaleMcGrew


DaleMcGrew:
👍

Issue #342, opened 2016-07-14
Instant Search Box: Positioning


DaleMcGrew:
Issues Thursday night:
- [x] When 693px and narrower, enter a search term so you see the drop-down. When the drop-down is open, the left nav hamburger doesn't react to clicking except for a very small click area.
- [x] When 768px and wider, the search box jumps up vertically.
- [x] When 768px and wider, the width of the search box jumps as you expand the browser. The way it is expanding is just fine leading up to 768px wide, so no special rules are needed for 768px and wider.
- [x] When 767px and narrower, the header (with the Instant search box) is about 20px taller than it should be
- [x] STILL AN ISSUE: Put the chrome window into developer mode. When the browser window is 329px (and narrower), remove the "My Voter Guide demo" text and make sure the search box stays on one line. Currently it drops down to a second line. I think this is important to fix because I suspect some older phones force the browser window to smaller than 329px wide.
- [x] TO BE TESTED: On an Android phone (I don't know the exact model or browser), the search box was on a second line, and it shouldn't be

The instant search box has some positioning issues.
- [x] The instant search box is now lower than it should be in all Chrome (and Safari) browser widths EXCEPT between 770px and 805px, when is is positioned correctly vertically.
- [x] Now that the instant search box is in place, when you click on the hamburger, so that the left navigation menu slides out, the search box remains visible *over* the menu that slid out. The Instant search box should not be visible.

DaleMcGrew:
@pertrai1 I added one more bullet item in the description of this issue. Still working on another comment too... (coming soon)

DaleMcGrew:
All comments are above now.

pertrai1:
@DaleMcGrew - the 768px issue is something to do with how Bootstrap handles media queries at that size and I am still trying to find out what properties it is putting in the DOM that makes that issue arise.

pertrai1:
@DaleMcGrew I tested the changes on my Samsung Galaxy Note Edge and the instant search is not dropping to a new line.

DaleMcGrew:
Thank you Rob! It looks really good, and is ready for our presentation tomorrow. Thank you! One item we can deal with later:
- [ ] The "add friends" and "Who I can follow" text is showing up within the Instant search drop down and should not be visible.
<img width="590" alt="screen shot 2016-07-15 at 6 03 56 pm" src="https://cloud.githubusercontent.com/assets/7756031/16891898/e3d5a954-4ab6-11e6-920f-5267a9e80106.png">

[...]

PR #344, opened 2016-07-16
#342: Instant Search Position Fixes

pertrai1:
Handle the items that were listed as issues that centered around the instant search box. These can be found in the issue #342  

Signed-off-by: Rob Simpson <rsimpson2@me.com>


DaleMcGrew:
👍

PR #345, opened 2016-07-16
Fix for #342 z-index overlay

pertrai1:
The Add Friends and Who I Can Follow were stacked on top of the search container without having the correct z-index for the elements. There also was an issue where the search box would be stacked on top of the hamburger menu when it was open. This has been resolved.


DaleMcGrew:
👍

PR #346, opened 2016-07-16
Removing code coverage not used on README

pertrai1:
A couple of the code coverage tools we have not been using so the buttons are removed.


pertrai1:
retest this please

PR #350, opened 2016-07-30
Migrated twitterIdentityRetrieve calls to own Store and Actions files…

DaleMcGrew:
…. Added the creation of an Organization record when a Twitter account sign in happens. Added call to organizationSave API endpoint. Removed console.log printouts from many files. Changed our abbreviateNumber rounding function to match Twitter's single digit "69.4M" display.


DaleMcGrew:
👍

PR #353, opened 2016-08-04
Fix modal styling

### LINKED ISSUE #352: Modal not showing over page as popup
I am trying to get a Modal display working. It is currently showing up at the bottom of the page though. @ixley thinks it might be because of the prefix we are adding to all bootstrap classes.

You can see this here:
http://localhost:3000/defenders
...and click on the % rating. It will try to pop up the Modal.

I am including the "EditPositionAboutCandidateModal" component in this component:
src/js/components/VoterGuide/OrganizationPositionItem.jsx
src/js/components/VoterGuide/EditPositionAboutCandidateModal.jsx

ixley:
I found a way to use this component without rolling back the prefixing. If it creates unresolvable issues in the future, we can still roll back the prefixing, but I would obviously prefer not to, if possible.

@DaleMcGrew I couldn't tell how the modal trigger was being bound, but the trigger content/text should be given a pointer cursor.

resolves #352

### DIFF
--- a/package.json
+++ b/package.json
@@ -76,7 +76,7 @@
     "pre-commit": "^1.1.2",
     "react-addons-linked-state-mixin": "^0.14.3",
     "react-addons-test-utils": "~0.14.0",
-    "react-bootstrap": "^0.28.1",
+    "react-bootstrap": "^0.28.5",
     "react-headroom": "^1.7.3",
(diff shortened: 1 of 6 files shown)


DaleMcGrew:
Thank you @ixley! I will add the css to cause the cursor to turn into a pointer on hover. 👍

PR #354, opened 2016-08-06
First decent EditPositionAboutCandidateModel and refactoring

DaleMcGrew:
Converted PositionItem to use position object instead of scattered variables. Updated FollowToggle to hide the Follow button if you are looking at your voter guide. Solved some react warnings re: missing key's on list items. Implemented start of data-driven display of EditPositionAboutCandidateModel. Collapsed closing tags in several places.


DaleMcGrew:
👍

PR #360, opened 2016-08-10
renamed variable link_to_ballot_item_page...

zachmonteith:
also added a conditional based on link_to_ballot_item_page to determine whether candidate's image is a link


DaleMcGrew:
👍

PR #363, opened 2016-08-12
Implemented new navigation scheme & added "My Profile" page

DaleMcGrew:
Implemented new navigation scheme that moves ballot filtering into a drop-down at the top of the ballot. Added "My Profile" page that goes to your Twitter page.


DaleMcGrew:
👍

PR #365, opened 2016-08-13
Updated site language to "Your" instead of "My". Made the position entry autosize.

DaleMcGrew:
Updated site language to "Your" instead of "My". Made the position entry field an autosize Textarea Increased the text size of major interface elements.


DaleMcGrew:
👍

PR #366, opened 2016-08-13
Switch easily between "Who You Can Follow" and "Who You're Following".

DaleMcGrew:
Added "FollowingDropdown" drop down navigation so we can switch easily between "Who You Can Follow" and "Who You're Following".


DaleMcGrew:
👍

PR #368, opened 2016-08-13
custom compile of bootstrap

ixley:
This PR introduces no style changes, but sets up Bootstrap to compile from Sass files instead of the flat CSS file we had before. This will allow us to pass our own variables and be selective about what things we want to import, for better customization and integration.


DaleMcGrew:
@josephevans will any of these changes affect your production release scripts?

ixley:
This still creates the same output – a single bootstrap.css file. I would be interested in knowing what the production release process is or where to review the scripts though, so I can be mindful of it in the future.

DaleMcGrew:
Hi @ixley I don't know that this will affect our build process--I am just checking with @josephevans  to be sure. I'm approving this now. 👍

ixley:
@DaleMcGrew - of course, that's completely reasonable. I had no reason to assume it would disrupt anything, but that's partly because I am not familiar with the production release process. Having at least a basic familiarity or understanding would at least allow me to bring things like this up if there's the possibility it would have an effect.

PR #370, opened 2016-08-16
fixed measure naming collision, created and commented out MeasureCons…

zachmonteith:
…tants.  Created the constants file, because it was called for and didn't exist, but since neither Dale nor I could quickly determine what these constants were for, I simply commented all references to them out.  The naming collision is the main fix of this pull request (components/ballot/measure is now components/ballot/measureitem to disambiguate from routes, tracked down and replaced all mentions of former name)


DaleMcGrew:
👍

PR #371, opened 2016-08-17
fix glyphicons directory path

ixley:
The default Glyphicons path got broken when creating custom compile of Bootstrap – this fixes it.
This commit removes a bunch of redundant lines that were left as reference, but the key edit is for $icon-font-path.


zachmonteith:
👍

PR #373, opened 2016-08-18
Changes to Navigator Icons

zachmonteith:
Added text to header icons, resized search bar, changed color/highlighting of navicons, renamed and relocated navigator files.

TODO: standardize size of navicons, get active state to work properly for header as it currently works for footer.


DaleMcGrew:
👍

PR #375, opened 2016-08-18
added state to header navicons, experimented with VoteSmart Popovers

zachmonteith:
two files in this one need work: PositionRatingSnippet and the new file, RatingPopover, both contained in src/js/Widgets


DaleMcGrew:
👍

PR #376, opened 2016-08-19
Modified toggle between "public" and "friends only" states.

DaleMcGrew:
Iteration on trying to get React Popover working.


DaleMcGrew:
👍

PR #381, opened 2016-08-20
Deleting src/js/components/Measure file that was renamed to src/js/components/MeasureItem.

DaleMcGrew:
Deleting src/js/components/Measure file that was renamed to src/js/components/MeasureItem.


DaleMcGrew:
👍

PR #382, opened 2016-08-21
Fixed issue where support/oppose bar was not showing up for ballot_it…

DaleMcGrew:
…ems that aren't on your primary ballot. Added positionCountForOneBallotItem API endpoint. Sped up initial load of ballot. Fixed bug with ballot not loading if you start on candidate or measure page.


DaleMcGrew:
👍

Issue #324, opened 2016-07-01
Display explanation of what a Vote Smart rating means


DaleMcGrew:
Throughout We Vote we have ratings that have come from Vote Smart. When someone clicks on a rating, pop up a div that explains how the rating was created. (Use one of the bootstrap popups.)

<img width="381" alt="screen shot 2016-07-01 at 10 40 46 am" src="https://cloud.githubusercontent.com/assets/7756031/16529528/4e2d920c-3f78-11e6-9149-ac027e73fc57.png">

ixley:
@dalemcgrew Can you provide a sample of the text you'd like here?

DaleMcGrew:
Thank you @ixley, lets start with this:
Source: VoteSmart.org
Ratings are given by the organization, and collected by the nonprofit Vote Smart. 
(down arrow icon) 0% is a low score, and (up arrow icon) 100% is a high score. 
Ratings can be invaluable in showing where an incumbent has stood on a series of votes in the past one or two years, especially when ratings by groups on all sides of an issue are compared. Some groups select votes that tend to favor members of one political party over another, rather than choosing votes based solely on issues concerns. Please visit the group's website or call 1-888-VOTESMART for more specific information.

ixley:
@DaleMcGrew I'm not quite ready to take this - was just trying to help add some definition so it's ready to go.
This is also probably more text than is appropriate for a popover, so it's worth considering whether this would work in a modal or whether we can trim down the text a bit.

DaleMcGrew:
Fair enough @ixley! Ideally the popover could show some of the text, but then have a way to reveal the rest. Perhaps a way to scroll down?

ixley:
Yeah, that's reasonable to try. A technical blocker is that React-Bootstrap doesn't currently support custom prefixing for all classes, which makes it largely incompatible with the namespaced classes I prefixed. I wasn't clear how important it would be to leverage third-party Bootstrap 'plug-ins', but it's seeming like it might make sense to roll back the Bootstrap class namespacing for the time being so it's not an impediment.

pertrai1:
Possibly a good scenario would be to have the side bar open up with the menu hidden and have the information shown about the source that would allow for the complete content to be revealed.

ixley:
I think a popover (large ‘tooltip’) could be reasonable but the amount of
text makes it feel a bit heavy. Additionally, in the current context (each
item citing vote smart as the source), the content become redundant
[...]

PR #385, opened 2016-08-22
added text for VoteSmart popover, changed color of position-rating__s…

zachmonteith:
…ource text

this was a simple change, as Dale/Jeff already got the popover working, but with sample text.  changed text to approved VoteSmart Rating explanation, modified css color so users will more easily recognize they can click to read popover (light grayish blue, so not typical hyperlink color but similar enough to encourage a click)

This is intended to resolve issue #324

### DIFF
--- a/src/js/components/Widgets/PositionRatingSnippet.jsx
+++ b/src/js/components/Widgets/PositionRatingSnippet.jsx
@@ -14,7 +14,7 @@ export default class PositionRatingSnippet extends Component {
     var src;
     var className;
     var alt;
-    //var Popover = <RatingPopover id="testID" placement="top" />;
+
     if (rating >= 65){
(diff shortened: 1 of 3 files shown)


zachmonteith:
one quick note here: changed title text to "VoteSmart Ratings" because user can see this text, but perhaps this is not what we want the title to be, as these ratings are actually from the organizations themselves and merely aggregated by VoteSmart.  This title could instead be "Organizational Ratings", "Candidate Ratings" (with a conditional statement for Measures), or just "Ratings".  Since the purpose of the text is to explain the VoteSmart sourcing, I left the title as VoteSmart Rating for now pending further discussion.

Also, I could be wrong, but isn't the name of the nonprofit actually Project Vote Smart?  I'll gladly make that change if necessary.

DaleMcGrew:
👍

Issue #325, opened 2016-07-01
Clicking on Rating bar should take you through to Candidate page


DaleMcGrew:
When you are on the ballot page, a click anywhere on the positions bar should take you to the candidate's page.
<img width="546" alt="screen shot 2016-07-01 at 10 47 03 am" src="https://cloud.githubusercontent.com/assets/7756031/16529691/323c3f84-3f79-11e6-9aac-9e9afd6621f5.png">

pertrai1:
@DaleMcGrew Is this not needed for the shootout? I do see Next but wanted to make sure that it is not needed for the shootout and is there just because there was no resources for this to make it into the shootout.

DaleMcGrew:
@pertrai1 It would be ideal to have for Saturday, but is not critical.

DaleMcGrew:
Thank you @zachmonteith, this is working very well.

PR #386, opened 2016-08-22
made all ballot item card elements redirect on click, with pointer cu…

zachmonteith:
…rsor

this resolves issue #325 and then some (not only ratings bar, but all elements on a ballot item card now redirect user to the appropriate candidate or measure page upon click, and users will see pointer hand cursor if they hover over any clickable items)


DaleMcGrew:
👍

Issue #377, opened 2016-08-19
Position toggle switch between "Public" and "Friends Only"


DaleMcGrew:
I have put in place a temporary toggle button that shows whether a position is visible to "Friends Only" vs. "Public":
![screen shot 2016-08-18 at 8 17 32 pm](https://cloud.githubusercontent.com/assets/7756031/17797827/dab7408c-6580-11e6-8024-b19457b6b97e.png)

My ideal case would be that it would look like the toggle switches seen here, and be right-aligned above the "Post" button:
<img width="78" alt="screen shot 2016-08-18 at 8 21 10 pm" src="https://cloud.githubusercontent.com/assets/7756031/17797882/568b6e5e-6581-11e6-91b6-dac8fcdd33b9.png">
http://www.bootstraptoggle.com/
I tried installing this, but couldn't get it to work. I suspect the "bs-" class prefixing preventing this from working.

If that toggle switch isn't possible, these would be ok, but second-best:
http://www.bootstrap-switch.org/examples.html

ixley:
@DaleMcGrew Do you have a branch I could take a look at? 

From just scanning the demo/doc pages for both of those, it appears that yes, Bootstrap Toggle is dependent on external Bootstrap styles, which could cause it to break (visually) if those external classes are not as it expects them. It's possible that you could customize around this, but it didn't jump out at me. 

Bootstrap Switch appears that while it is based on Bootstrap's vanilla styles and patterns, it has its styles set up independently, which should make it easier to customize (ultimately what we want).

I am happy to expand upon my position more, but in short, plugins should be more or less standalone, and ideally extensible (i.e. modular). It's reasonable for them to have their own dependencies, but requiring an entire external **style** library feels inappropriate (as well as somewhat antithetical to React). I realize I am taking an opinionated stance on this, so I would be open to hearing other opinions on the matter as well. I know this requires a bit more thoughtfulness and selectivity when looking for plugins, but it seems more practical to me to keep our dependencies as modular and customizable as possible.

DaleMcGrew:
The very latest code we worked with is in the develop branch.

[...]

PR #388, opened 2016-08-23
added toggle switch for positions public vs friends

zachmonteith:
intended to resolve issue #377 

TODO: hide toggle switch if screen is too small to accomodate


DaleMcGrew:
👍

PR #389, opened 2016-08-24
Updated variable name & added retrievePositionsCountForOneBallotItem to organizational voter guide page

DaleMcGrew:
Updated candidate_display_name to general variable name "ballot_item_display_name". Added retrievePositionsCountForOneBallotItem to organizational voter guide page, to retrieve counts that aren't on your current ballot.


DaleMcGrew:
👍

PR #393, opened 2016-08-25
Fixed disappearing organizations. Updating position counts as orgs are followed.

DaleMcGrew:
Fixed issue with organizations disappearing from the "More opinions" section of a candidate's page when followed. Now updating position counts as organizations are followed. Improved variable name guideList associated with GuideList component.


DaleMcGrew:
👍

PR #395, opened 2016-08-26
Reworked the way that we select orgs from voterGuidesToFollow and mov…

DaleMcGrew:
Reworked the way that we select orgs from voterGuidesToFollow and move them to position_list. Changed "this.listener" to more specific variable name.


DaleMcGrew:
👍

Issue #403, opened 2016-08-26
Position Toggle Switch should appear on Your Page for each ballot item with a position


zachmonteith:
currently the position toggle switch (to change a post from one only your "Friends" can see to one available to the "Public" and vice-versa) appears on the Ballot page and on the individual ballot item's page, but not on the list of ballot items a user has taken positions on (available by clicking "Your Page" in the hamburger menu/left column menu, or navigating to /{the user's twitter handle}).

First step will be to extract the toggle from its current home in ItemActionBar into a standalone file, in order to increase its flexibility and compartmentalization.  

Second step, integrate into PositionItem and OrganizationPositionItem, to appear if signed_in_with_this_twitter_account === true.

DaleMcGrew:
I am working on changing the value of signed_in_with_this_twitter_account based on live data.

DaleMcGrew:
Fixed.

DaleMcGrew:
Hi @zachmonteith I think this was working, but isn't working any longer. When I go to my page now (signed into my Twitter account http://localhost:3000/WeVote ), I don't see the "Friends/Public" toggles any more. Can you please take a look?

DaleMcGrew:
@zachmonteith can you please verify this? I believe I fixed it.

zachmonteith:
Yes, it's working both on ballot page (where it is pulled in by ItemActionBar) and on "Your Page" (where it is pulled in by OrganizationPositionItem).  I tried it locally and in the version running on wevote.me and toggled a few positions, everything seems great.

zachmonteith:
👍

PR #406, opened 2016-08-26
extracted PositionToggleSwitch and included in OrganizationPositionItem

zachmonteith:
intended to resolve issue #403 

TODO: currently, signed_in_with_this_twitter_account is always set to true (line 61 of js/components/VoterGuide/OrganizationPositionItem), so these toggles will appear for anyone looking at anyone else's positions, instead of only the authorized user seeing the toggle.


zachmonteith:
oops, name of the new file is not PositionToggleSwitch, but PositionPublicToggle

zachmonteith:
as commit name suggests, this pull request now also resolves issue #407

DaleMcGrew:
👍

Issue #402, opened 2016-08-26
Broken image showing in some places. Should show placeholder.


DaleMcGrew:
The bulk of this issue is resolved (thanks @zachmonteith!) but:
- [ ] Spacing still doesn't match the correct images:
![screen shot 2016-08-27 at 7 58 44 am](https://cloud.githubusercontent.com/assets/7756031/18028164/1cdb44e8-6c2c-11e6-9133-86ca94ff362a.png)

DaleMcGrew:
Hi @zachmonteith, I notice that since this change, if a candidate doesn't have an image link in the system, the placeholder isn't shown. 
To reproduce:
1) Sign in with Twitter.
2) Go to "Charlotte, NC" election.
3) Express an opinion for a down-ballot candidate that doesn't have an opinion.
4) Go to your page and you will see an empty space where the placeholder should be.
![screen shot 2016-08-31 at 7 06 46 am](https://cloud.githubusercontent.com/assets/7756031/18132025/daff334c-6f49-11e6-8bff-9ecc3522ee0b.png)

DaleMcGrew:
Verified, thank you @zachmonteith!

PR #409, opened 2016-08-27
renamed and retooled ImageHandler, resolves issue #402

zachmonteith:
All avatar images now run through the same handler.  When images return an error they are now replaced with either an icon that looks like a chess rook (for orgs and twitter handles) or a circle human head silhouette for candidates, hexcolor #A9A9A9 (a light gray) for all.  

TODO:  Left the blue underline that appears when cursor hovers over icons, not sure whether that is desired behavior.


zachmonteith:
going to bed now, but currently working on a fix for spacing/dynamic sizing issues for replacement icons, which are sometimes appropriately sized (such as for PositionItems and OrganizationPositionItems but not candidate cards or a few other places.  size should be passed in as a prop, or at least specified with a css selector included in the classname passed through.

src/js/components/Ballot/PositionItem.jsx
```diff
@@ -86,9 +87,8 @@ export default class PositionItem extends Component {
       {/* One Position on this Candidate */}
         <Link to={speakerLink}>
           { position.speaker_image_url_https ?
-            <img className="img-square position-item__avatar"
-                  src={position.speaker_image_url_https}
-                  width="50px"
+            <ImageHandler className="img-square position-item__avatar"
+                  imageUrl={position.speaker_image_url_https}
```
DaleMcGrew: Hi @zachmonteith, it seems like it would be fine to force an image width with a prop passed into ImageHandler (like the original img tag had).

DaleMcGrew:
👍  Great work @zachmonteith. I am approving this and would suggest that you add the ability to force the image size with a prop value passed into ImageHandler in a subsequent check in.

PR #413, opened 2016-08-27
Fixes magnifying glass search button styling 

kyleludlow:
Fixes magnifying glass search button styling by moving class up one nest level in header.scss


DaleMcGrew:
👍

PR #416, opened 2016-08-28
UI fixes seen in production

pertrai1:
These might not be intended to be seen this way according to the changes
that I am proposing. Seen to be a bit off when I look at the app.

### DIFF
--- a/src/sass/components/_itemActionBar.scss
+++ b/src/sass/components/_itemActionBar.scss
@@ -11,7 +11,7 @@
 .item-actionbar {
   display: flex;
   border-top: 1px solid #DDD;
-  margin-top: 1em;
+  margin-top: 1em 0;
   padding-top: 10px;
 
   &__btn {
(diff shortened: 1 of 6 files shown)


DaleMcGrew:
This looks great @pertrai1. I will be able to test it out on my local machine in a bit, and then merge it in.

DaleMcGrew:
👍

DaleMcGrew:
@pertrai1 I think this pull request removed the icons from the top navigation -- they are no longer appearing.

pertrai1:
@DaleMcGrew Is this on wevote.me right now? It seems that the icons are there and are not absolute to the right of the window. If this is not on wevote, this commit can be reverted. Most of what I see on the UI seems to be font sizes that are extremely large.

DaleMcGrew:
Hi @pertrai1, I didn't push the changes live to WeVote.me -- I will revert the styles necessary to get the top nav items back. Thank you!

pertrai1:
👍

src/sass/components/_itemActionBar.scss
```diff
@@ -11,7 +11,7 @@
 .item-actionbar {
   display: flex;
   border-top: 1px solid #DDD;
-  margin-top: 1em;
+  margin-top: 1em 0;
```
ixley: @pertrai1 did you mean to change `margin-top` to `margin`?

pertrai1: @ixley great catch. Did not even see that.

PR #417, opened 2016-08-28
Adding meaning pull request and issue request messages

pertrai1:
These templates can make for meaning messages that are found in pull
requests and in issues. I have an issue I want to add and having a good
template to go by can help those who look into the issue have the
information that is needed.


DaleMcGrew:
👍

Issue #411, opened 2016-08-27
Show the office the position is for, when looking at a voter guide


DaleMcGrew:
How it currently shows:
![screen shot 2016-08-27 at 8 50 03 am](https://cloud.githubusercontent.com/assets/7756031/18028422/c7647b30-6c33-11e6-8ad7-82be5ba39375.png)

The "office" line we want to copy from the CandidateCard:
![screen shot 2016-08-27 at 8 54 30 am](https://cloud.githubusercontent.com/assets/7756031/18028428/e72e78b2-6c33-11e6-8c04-3db25df0ee34.png)

DaleMcGrew:
Hi @zachmonteith, I worked it out and it works well now!

PR #430, opened 2016-09-01
OfficeNameText now functional #411, OfficeStore only used in Office routes

zachmonteith:
changes in this pull request:
1) renamed a few files and variables to avoid name duplication/collisions (most obvious of these is GuidePositionList) and be more descriptive
2) removed deprecated classnames from ImageHandler 
3) created OfficeNameText component to display party name and office name for candidates #411 
4) removed OfficeStore from places where it was only being called to populate office_name, which **TODO: will be passed in with "position" and "candidate" objects**
5) preliminary changes to OfficeStore in preparation for Office routes fix


DaleMcGrew:
👍

PR #431, opened 2016-09-01
ImageHandler fix: pass in kind_of_ballot_item, remove conditional sta…

zachmonteith:
…tments #402

This is a fix for the bug Dale encountered where some candidates were not getting image placeholders if they didn't have an image to begin with.  

Also eliminated redundant code, especially conditional statements depending on whether ballot items had images: all will now be handled by ImageHandler, which will return a placeholder icon if the ballot item doesn't have an image url OR if it is a broken/bad image url


DaleMcGrew:
👍

PR #432, opened 2016-09-02
Adjusted variable names passed into OfficeNameText to match new fields from API server

DaleMcGrew:
Adjusted variable names passed into OfficeNameText to match new fields from API server. Tweaks to github issue and pull request starter text templates. Changed "connect" path so we don't prevent @connect Twitter handle from using site. Added stub files for taking in we_vote_id to show a voter's voter guide when not using Twitter.


DaleMcGrew:
👍

PR #448, opened 2016-09-07
Transitioning to compressed office and measure display on the main ballot page.

DaleMcGrew:
Transitioning to compressed contests on the main ballot page.

### DIFF
--- a/src/js/components/Ballot/OfficeItemCompressed.jsx
+++ b/src/js/components/Ballot/OfficeItemCompressed.jsx
@@ -1,0 +1,81 @@
+import React, { Component, PropTypes } from "react";
+import { Link, browserHistory } from "react-router";
+import StarAction from "../../components/Widgets/StarAction";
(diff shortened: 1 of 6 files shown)


src/js/components/Ballot/OfficeItemCompressed.jsx
```diff
@@ -0,0 +1,81 @@
...
+    ballot_item_display_name = capitalizeString(ballot_item_display_name);
+
+    return <div className="office-card__container">
+      <div className="office-card">
+        <div className="office-card__media-object">
+          <div className="office-card__media-object-anchor">
+
+          </div>
+          <div className="office-card__media-object-content">
+            {
```
zachmonteith: We shouldn't need information about support/oppose in network for Offices; while users can comment on an office using the ItemPositionStatementActionBar, we aren't allowing them to support or oppose the office itself

zachmonteith:
👍

Issue #444, opened 2016-09-07
Remove 2nd Header Bar


DaleMcGrew:
In mobile mode, the 2nd header bar should take less vertical space, and the text should not wrap.
![screen shot 2016-09-07 at 7 51 43 am](https://cloud.githubusercontent.com/assets/7756031/18316636/46cd72c4-74d0-11e6-93cf-9dfa221aa2e7.png)

Dale would like to be trained in the layout styles for the header so he can make tweaks.

ixley:
Can we just remove this? We've discussed before how these nav items are relevant in certain contexts but probably don't warrant taking up that much prime space all the time.

DaleMcGrew:
I'd like to leave this bar in place for user testing. I know that now it doesn't feel right the way it is displayed currently. This 2nd header bar will feel less "heavy" when we make it *much* shorter. I think we will also want to change the buttons we put there after some experimentation. Facebook has a 2nd header bar like this, and I suspect it is based on what they've learned through user A/B testing.

ixley:
@DaleMcGrew I can appreciate getting feedback from users on this, but I still think there are better ways to surface these options _in relevant contexts_ without having them take up prime space all the time.

zachmonteith:
I just made the header bar shorter itself without changing the icon/text font size.  could change icon size too if you want it shorter still, just let me know.

ixley:
I'm removing this with Dale's acquiescence.

PR #449, opened 2016-09-07
fix for issue #444

zachmonteith:
created a quick and dirty new css classname to make second header bar take up less vertical space

### DIFF
--- a/src/sass/layout/_boxmodel.scss
+++ b/src/sass/layout/_boxmodel.scss
@@ -15,6 +15,10 @@
     padding: .75em;
 }
 
+.fluff-none {
+  padding: 0;
+}
+
 .fluff-loose--top {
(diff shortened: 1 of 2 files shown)


src/sass/layout/_boxmodel.scss
```diff
@@ -15,6 +15,10 @@
     padding: .75em;
 }
 
+.fluff-none {
+  padding: 0;
```
ixley: @zachmonteith don't add a single zeroed value for a utility class. Just removing `fluff-tight--full` would accomplish the same thing. In the end, we are removing this entire bar for now, so I'm closing this PR – just wanted to give you some feedback on your change.

ixley:
The subheader is being removed in #452 so this PR isn't necessary.

PR #457, opened 2016-09-11
Change ballot dropdown to filter nav

ixley:
Changes the Ballot filter dropdown menu to a button-group nav.

This PR can be merged but would benefit from a bit of cleanup.
- The conditional for adding `active` class to selected tab should be cleaned up (I wasn't sure how to)
- The conditional for the date tooltip on Election Name heading could be cleaner. Currently it still triggers an empty span if there is no election date. Ideally, the overlay trigger for the tooltip would not even get used, but I wasn't sure how to target the wrapping container to be conditional.
- Because the Bookmarks page has a different layout, the transition is jarring. Consider putting ballot-specific bookmark content in the same layout. An easier patch would just be to add a similar heading to the Bookmarks page so the positioning of the filter nav is at least consistent/similar.


src/js/components/Navigation/BallotFilter.jsx
```diff
@@ -0,0 +1,32 @@
...
+    params: PropTypes.object,
+    ballot_type: PropTypes.string
+  };
+
+
+  render () {
+    const {ballot_type} = this.props;
+
+    return <div className="btn-group">
+      <Link to="/ballot" className={ ballot_type === "ALL_BALLOT_ITEMS" ? "active btn btn-default" : "btn btn-default"}>
```
ixley: Find a better way to add `active` as a conditional class for selected nav items.

src/js/routes/Ballot/Ballot.jsx
```diff
@@ -162,20 +160,23 @@ export default class Ballot extends Component {
         <h3 className="text-center">{this.emptyMsg()}</h3>
         {emptyBallotButton}
       </div> :
-      <div></div>;
+      null;
+
+    const electionTooltip =  election_date ? <Tooltip id="tooltip">Ballot for {election_date}</Tooltip> : <span></span>;
```
ixley: Find a better way to create a conditional so that the tooltip doesn't even get triggered if there is no content.

DaleMcGrew:
👍

ixley:
@DaleMcGrew thoughts on those things I flagged for cleanup? I know they're probably pretty simple – I'm just not familiar with the appropriate way to set it up. Seemed too small to add as issues, but I can do that too. @zachmonteith - maybe you're familiar and could show me how?

DaleMcGrew:
Good suggestions @ixley. One of us can tackle it soon. I created Issue #459 for this.

PR #458, opened 2016-09-11
use icons with tooltips for visibility toggle

ixley:
Change visibility toggle to use icons with tooltip for clarification.

*I'm still interested in exploring a dropdown button instead of a toggle, or other adjustments to make sure it is noticeable but not overly prominent.


DaleMcGrew:
👍

PR #462, opened 2016-09-14
.card layout (css) refactoring

### LINKED ISSUE #456: Cards (Measures/Offices) w/o images need adjustment for position icon
Cards with positions that don't have avatars/images, such as Measures or Offices, need the position of the support/oppose icon adjusted.

image

image

ixley:
This PR is primarily centered around stylesheet refactoring for the .card pattern/layout and all it's related elements across multiple components. Most of the cleanup was behind the scenes, but the card layouts should be a bit more consistent and less fragile too.

I anticipated the refactoring effort would be significant, but it also highlighted the complexity of the styles being used. I'm not sure what a more appropriate approach would be, but I'm open to feedback or discussion along these lines. It might help to document the structure of these card layouts as well—perhaps as a simple tree map—but again, it feels a little awkward to need to do something like that. **I've also temporarily shelved setting up the pattern library as I've found the default templating languages to be frustrating to use. I'll keep it on my list, though.**

*I checked most routes/components, but couldn't recall how to preview the position info within the modal. Please double check that and/or let me know how to preview it.

fixes #456


DaleMcGrew:
👍 Very cool @ixley! I did a code review on all files and it looks great.

PR #463, opened 2016-09-14
Show your network's opinions if there are any for a Measure on the ballot list.

DaleMcGrew:
Show your network's opinions if there are any for a Measure on the ballot list.


DaleMcGrew:
👍

PR #467, opened 2016-09-15
Fixed bug with Share on Facebook button. 

aayalah:
The Share button was not appearing on Firefox or Internet Explorer, so I fixed the bug that was preventing it from working.


DaleMcGrew:
Thank you @aayalah! 👍

Issue #465, opened 2016-09-15
Position statement text too long on Measure & Candidate pages


DaleMcGrew:
### Please describe the issue (What happens? What do you expect?)
1. Go to either a Candidate or Measure page that has a position with a very long statement, like this one: https://wevote.me/measure/wv02meas115
2. Some of the statements take up way too much vertical screen space.
3. We should limit the amount of text that displays to ~4 lines of text, and offer a "More" link that "unfurls" the text so you can read all of it. I don't think we need to offer a "shrink" link once the text is unfurled.

<img width="434" alt="screen shot 2016-09-14 at 5 38 46 pm" src="https://cloud.githubusercontent.com/assets/7756031/18534698/d80f1756-7aa2-11e6-9a25-4786d16cc4a5.png">

DaleMcGrew:
Verified!

PR #472, opened 2016-09-19
implemented ReadMore text truncator for position statements #465

zachmonteith:
Created ReadMore.jsx in components/Widgets.  uses react-text-truncate to truncate text, could easily be imported wherever needed.  

Pass in as props: 
text_to_display (required), 
link_text (the text for the link to expand the truncated text, defaults to "More"),
collapse_text (text for link to collapse back to truncated text, defaults to  "...Less"), and 
num_of_lines (the number of vertical lines of text which will appear in the truncated version, defaults to 3)

included in both Position Statements (Oppose/Support and Information Only)


DaleMcGrew:
👍

Issue #435, opened 2016-09-03
Create "copy link" Modal for ShareButtonDropdown


DaleMcGrew:
In Widgets/ShareButtonDropdown we have a "copy link" option. Turn this "copy link" on, and when the option is selected, open a modal with the link in a select box that allows the person to copy the link to their clipboard.

To get started, I suggest searching for "EditPositionAboutCandidateModal" for example code you can start with.

Next step after this is to show a preview of the link in the modal.

abbynormative:
I'm interested in working on this. Can you post a link to a page in the WeVote app where the copy link option should appear?

DaleMcGrew:
Hello, you can use a dummy link  for now in the modal.  The copy link option is currently commented out in the ShareButtonDropdown component.

abbynormative:
Ok. Can you post a link to one of the pages that has this button? I just want to clarify what exactly I'm working on, and to have a page to view when testing my code.

DaleMcGrew:
Any page with the support/oppose/share buttons. https://wevote.me/measure/wv02meas100

DaleMcGrew:
Verified! Great work @zachmonteith.

PR #475, opened 2016-09-20
created CopyLinkModal for ShareButtonDropdown #435

zachmonteith:
Tested on desktop and mobile browser: selecting "Copy Link" from share dropdown now activates a modal with an input field containing the ballot item url and a button to copy that url directly to the user's clipboard using package react-copy-to-clipboard.  
On click, red text appears telling user they have successfully copied the url, and suggesting what they might do next.


DaleMcGrew:
👍

Issue #397, opened 2016-08-26
Make position source link active: open new tab in desktop mode, open modal with content in mobile mode


DaleMcGrew:
<img width="490" alt="screen shot 2016-08-26 at 6 15 37 am" src="https://cloud.githubusercontent.com/assets/7756031/18006525/c0b22d30-6b54-11e6-9d4a-96cfe19de6b8.png">

When a person clicks on the source link for a position:
- [x] In desktop mode, open the link in a new tab
- [ ] In mobile mode, open the link in a modal that can be closed

zachmonteith:
Unable to get modal to load and display full external webpage, so commented it out.  mobile users will not see a "view source" link, desktop users will open link in new tab as desired.

DaleMcGrew:
Agreed. We won't offer a "view source" link in mobile mode for the foreseeable future. Verified.

Issue #451, opened 2016-09-08
'.edit-position-action' renders even when I cannot edit a position


ixley:
`.edit-position-action` renders (around `.explicit-position`) for all items (even when I am not able to edit it. The functionality to edit is still (correctly) unavailable to me, but the element should not render, especially since it changes the cursor.

DaleMcGrew:
Verified!

PR #476, opened 2016-09-21
view source desktop, mobile in progress #397

zachmonteith:
Included in this PR:
new classnames for elements meant only to appear in mobile browser or desktop browser respectively: .interface-element--desktop and .interface-element--mobile.  elements with these classnames will only appear in the appropriate browser, using same breakpoints/min width of header vs footer navicons.

view source link (for position statements): on desktop, opens source webpage in new tab.
on mobile, work in progress.  goal was to have source webpage display in a bootstrap modal, but we were unable to get that working properly, so modal was commented out: currently, mobile users will not see a link to view the source of a position.

### DIFF
--- a/src/js/components/Ballot/PositionItem.jsx
+++ b/src/js/components/Ballot/PositionItem.jsx
@@ -1,7 +1,7 @@
 import React, { Component, PropTypes } from "react";
 import { Link } from "react-router";
 import ImageHandler from "../../components/ImageHandler";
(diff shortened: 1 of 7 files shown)


zachmonteith:
latest commit is a fix for #451.  "Edit this position" title, and the span containing it, was a holdover from the EditPositionAboutCandidateModal that has been commented out.  Commented out all appearances of the span with this title, and all functions designed to handle state of the modal we are not using: once we reintegrate and fix the edit position modal, these can easily be reintroduced.

DaleMcGrew:
👍

Issue #471, opened 2016-09-16
Support/Oppose buttons stack vertically on small phones


DaleMcGrew:
### Please describe the issue (What happens? What do you expect?)
On all phones, the "Support" and "Oppose" buttons stack one on top of the other. I think we can remove the text ("Support" and "Oppose") for those two buttons when the view is too small. The Public/private toggle also overlaps the "Share" button on some phones.
### Steps to reproduce the problem (1, 2, 3...), including links
1. Go to local ballot "Oakland, CA 94602"
2. Look at any proposition where you haven't taken a position yet, and notice the stacked  "Support" and "Oppose" buttons

![screen shot 2016-09-16 at 7 36 30 am](https://cloud.githubusercontent.com/assets/7756031/18589615/5cf0b2e8-7be0-11e6-80db-fbae9ce46000.png)

Example of public/private toggle covering "Share" button:
![img_2421](https://cloud.githubusercontent.com/assets/7756031/18590543/208ae6bc-7be4-11e6-8a4c-8c811d697a6c.PNG)

DaleMcGrew:
Verified!

ixley:
I appreciate saving space by removing the text labels on small screens, but I wonder if it makes more sense to hide the 'Share' label (instead of 'Support/Oppose') because 1) sharing is more a secondary action, and 2) the thumbs up/down icons look very similar without labels and/or coloring. Thoughts @DaleMcGrew @zachmonteith ?

![image](https://cloud.githubusercontent.com/assets/160662/18771631/50712f92-80f5-11e6-890e-7bad90336e47.png)

DaleMcGrew:
Hi @ixley I would be worried that removing just the word "Share" won't reclaim enough horizontal width on the smaller phones. I also think the "thumbs up" and "thumbs down" icons are easier to figure out than the share icon (without the word).

zachmonteith:
I thought about removing the word Share but I agree with @DaleMcGrew, the symbol is slightly more ambiguous and also we want to be encouraging users to share in order to scale up quickly.

I like the idea of color coding the icons though!  Same color scheme we are currently using for the positions you have already taken would look good and maintain consistency.  I also don't think this would confuse users as the other option disappears once you pick one, so the ability to make the choice/knowledge that you have made your choice is still clear.

ixley:
Though I agree color can help with distinction, I don't think adding color in this context is appropriate. If you guys both prefer Share have the emphasis here, that's fine.

PR #477, opened 2016-09-21
3 bug fixes, added ReadMore to ItemPositionStatementActionBar

zachmonteith:
BUG 1: The share link works for candidate links, but not for measure links.
FIX: "/candidate/" was hardcoded into the urlBeingShared, added conditional based on "type" of ballot item

BUG 2: Vote Smart popover not working in some places.
FIX: prop "popover_off" was being passed through, a holdover from EditPositionAboutCandidateModal. stopped passing it, but left the prop in case we ever want to turn off the popovers.

BUG 3: if the “view source” link is “www.cnn.com” it tacks it onto the end of the local url, so it goes here: http://localhost:3000/measure/www.cnn.com
FIX: added js if/else to see if url includes "http://", if not, put it on the front of the url.

Also, per Dale's request, user's own position statement, when viewed from the ItemPositionStatementActionBar, no longer allows user to edit by clicking anywhere, or shows entire statement.  Statement is now truncated by ReadMore if longer than 3 vertical lines, and user must click "Edit" if they want to edit.

### DIFF
--- a/src/js/components/Widgets/ItemActionBar.jsx
+++ b/src/js/components/Widgets/ItemActionBar.jsx
@@ -85,13 +85,13 @@ export default class ItemActionBar extends Component {
               <span className="btn__icon">
                 <Icon name="thumbs-up-icon" width={icon_size} height={icon_size} color={icon_color} />
               </span>
-              <span>Support</span>
+              <span className="interface-element--desktop">Support</span>
             </button>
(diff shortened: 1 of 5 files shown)


zachmonteith:
this PR now also resolves #471, using the new classnames added in my last PR.  Also improved fix for bug 3, adding toLowerCase and only checking for "http" in case the url begins https

DaleMcGrew:
👍

Issue #428, opened 2016-08-31
Position comments are not showing line breaks that have been entered


ColettePhair:
Line breaks aren't appearing in position explanations - so for content longer than 1 paragraph it's a big glob of text like in the "Money Out Voters In" position here - https://wevote.me/measure/wv02meas112

DaleMcGrew:
Verified.

PR #478, opened 2016-09-22
added expanded_text_to_display to ReadMore to preserve line breaks #428

zachmonteith:
react-text-truncate was not preserving line breaks in position statements, so used map function to split text on line breaks and put each section of text between line breaks into its own <span>.  As this would break truncate, stored expanded text in its own variable.  Truncated version will not preserve line breaks, but we wouldn't want it to as this would quickly overwhelm the limited number of lines visible (default: 3). Instead, if this.state.readMore is true, we display expanded text, otherwise we display truncated text.


DaleMcGrew:
👍 Great work @zachmonteith and @nickbarry! I loved your collaboration.

Issue #415, opened 2016-08-28
Search Box not hidden when clicking on body


pertrai1:
Is it a new experience that when you click away from the search container that it should still be there? Before if you clicked on the body, the search container would be hidden.

DaleMcGrew:
Hi @pertrai1, you are totally right--we seem to have broken that. Thank you for creating an issue for it!

ixley:
I noticed this behavior on the dropdown buttons in the ItemActionbar as well. All 'overlay' UI elements (modals, dropdowns, etc.) should have default behavior of closing/exiting when clicking/tapping outside the element (or hitting `esc`). 'Locked' modals windows are the only exception I can readily think of.

luojxxx:
I was talking with Zach and he mentioned that the problem goes away after upgrading from iOS9 to iOS10.

Potentially this is related to Safari 10 fully supporting the ES6 standard: https://developer.apple.com/library/content/releasenotes/General/WhatsNewInSafari/Articles/Safari_10_0.html
iOS9 Safari previously had relatively low support for the standard (it's all the way on the right):
https://kangax.github.io/compat-table/es6/
And given that babel doesn't fully support all ES6 features, it may not transpile all our ES6 code back down to supported ES5 (for instance Proxies are not supported by babel):
https://babeljs.io/docs/learn-es2015/

If you really want to track down this bug, you might be able to cross reference the unsupported ES6 features from the compatibility table for mobile and what isn't supported by babel. Although I'm relatively new to react and I'm new to the project, so I might be wrong about things here. But I wanted to share the information here in case it's helpful =)

DaleMcGrew:
I believe this has been resolved.

PR #484, opened 2016-09-22
added onBlur to make search results and share dropdown disappear #415

zachmonteith:
added existing closing functions to onBlur listener, now if users click somewhere else on the screen, search results and/or share dropdown will close.


DaleMcGrew:
👍

PR #485, opened 2016-09-22
Fixed timing issue with onBlur and onClick. Replaced onClick (which fires AFTER onBlur) with onMouseDown (which fires BEFORE onBlur).

DaleMcGrew:


DaleMcGrew:
👍

Issue #474, opened 2016-09-20
Name & position statements show over photo on mobile browser


zachmonteith:
Since implementation of read more link truncation, alignment of candidate name, position, and position statement text is off, partially obscuring the candidate image. NOTE: We believe this issue is caused by transitioning a "p" (paragraph tag) to a "div" for the class "explicit-position__text" in:
/src/js/components/Widgets/PositionSupportOpposeSnippet.jsx and
/src/js/components/Widgets/PositionInformationOnlySnippet.jsx

![img_4323](https://cloud.githubusercontent.com/assets/17035647/18682946/12cff340-7f23-11e6-9759-d126e6069945.PNG)
![img_4324](https://cloud.githubusercontent.com/assets/17035647/18682947/12f434b2-7f23-11e6-8db6-c3f62b787e69.PNG)

ixley:
@zachmonteith is this live or only on the develop branch? I'm not able to reproduce this on the live site.

zachmonteith:
this is live on wevote.me, but only experienced the behavior on a mobile browser.

1) navigate to wevote.me in mobile browser (I used chrome on iPhone 5s)
2) search for "league of pissed off voters" (or really, "piss" will do the trick)
3) scroll down to candidates in their voter guide

ixley:
I still can't reproduce in mobile (iOS) Chrome or Safari. Maybe clear your
cache?

zachmonteith:
hmm, that didn't seem to fix it.  Specifically I am at url https://wevote.me/theleaguesf, my current address is set to San Francisco 94116 and I am looking at their Positions for Other Elections.  bug beings with Preston Picus (he is the fourth candidate I can see), also happens with Kamala Harris, Jackie Speier, Bernie Sanders, Jane Kim, and Victor Hwang.

ixley:
@zachmonteith I'll leave this to you to troubleshoot unless you can provide more info with what's happening to the layout (try emulating mobile in Chrome desktop and using the inspector). I did notice that there's an anchor tag wrapping `card-child__avatar` which should probably just be wrapping the image instead (`card-child__avatar` and `card-card__content` should be siblings), but I can't comment on whether that's causing the problem since I can't recreate it.

![img_8612](https://cloud.githubusercontent.com/assets/160662/18686596/4d6df5ac-7f31-11e6-8490-02e317bd6d18.PNG)

ixley:
[...]

PR #494, opened 2016-09-24
Misc visual cleanup

ixley:
- cleans up leftover layout updates for .card-child media-object layouts (may address #474)

- fixes some visual bugs in ItemActionBar
- small adjustments for links following position statements (view source, show more/less, edit)
- removes `interface-element--desktop` and `interface-element--mobile`
    - added a more appropriate breakpoint for the Support/Oppose button labels
    - made the `view source` link visible for all screen sizes along with an icon to indicate it's an external link. Once there is a more mobile-friendly option, fine to show in a different context.
- renames `position-statement__edit-position-link-long` to `position-statement__edit-position-link` and renames previous `position-statement__edit-position-link` to `position-statement__edit-position-link--area` which better describes that it stretches the link over the whole area. That said, it's not clear that it's still needed, but I left the css in case.

### DIFF
--- a/src/js/components/Widgets/ReadMore.jsx
+++ b/src/js/components/Widgets/ReadMore.jsx
@@ -34,7 +34,7 @@ export default class ReadMore extends Component {
           link_text = "More";
         }
         if (collapse_text === undefined) {
-          collapse_text = " ...Less";
+          collapse_text = " ...Show Less  ";
         }
         let expanded_text_array = text_to_display.replace(/(?:\r\n|\r|\n){2,}/g, "\r\n\r\n").split(/(?:\r\n|\r|\n)/g);
 
@@ -60,7 +60,7 @@ export default class ReadMore extends Component {
               />
           </span>;
         } else {
-          return <span>{expanded_text_to_display}<a href="#" onClick={this.toggleLines}><br/>{collapse_text}</a></span>;
+          return <span>{expanded_text_to_display}<a href="#" onClick={this.toggleLines}>{collapse_text}</a></span>;
         }
     }
 }
--- a/src/sass/components/_itemActionBar.scss
+++ b/src/sass/components/_itemActionBar.scss
@@ -14,12 +14,20 @@
   margin: 1em 0;
   padding-top: 10px;
 
+  &__btn-set {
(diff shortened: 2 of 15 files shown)


src/js/components/Widgets/ReadMore.jsx
```diff
@@ -34,7 +34,7 @@ export default class ReadMore extends Component {
           link_text = "More";
         }
         if (collapse_text === undefined) {
-          collapse_text = " ...Less";
+          collapse_text = " ...Show Less  ";
```
ixley: Do we even need this 'Show less' toggle? I like displaying a truncated version when appropriate, but once someone clicks to show more, is it likely to add that much more clutter (length) to the page that folks will want to collapse it again? If it is a more common case that these statements will be so long people will want to toggle them 'open and closed' maybe we should consider alternative patterns. ?

DaleMcGrew: We added the "...Show Less" so we could play with it and see how it feels, and I have found that it really does help, especially in mobile view.

DaleMcGrew:
👍

src/sass/components/_itemActionBar.scss
```diff
@@ -14,15 +14,23 @@
...
 
+  &__btn {
+    // currently uses Bootstrap .btn styling
     &--position-selected {
       color: $selected-color;
     }
   }
 
+  &__position-btn-label {
+    @include breakpoints (max mid-small) {
```
ixley: This is a more appropriate breakpoint (and approach IMO) to hide position button (support/oppose) labels. @zachmonteith Utility classes are fine too (but I want to create and use them more intentionally - easily identifiable, reusable, consistent, and organized). We can chat more about this if you'd like.

src/js/components/Widgets/ReadMore.jsx
ixley: Okay - I exposed the (view source) link on mobile as part of this but made sure that the Show Less link stays inline with the text while the view source link wraps underneath. I'm fine with the Show Less link if it's useful but I want to be mindful of having lots of tiny little links near each other)

Issue #491, opened 2016-09-23
Add Truncate to Positions You Can Follow


DaleMcGrew:
When looking at a candidate or measure, the organizations shown under "Positions You Can Follow" include an organizational description. In mobile, I would like to limit the text to only two lines, with a "More..." link. (e.g. "Official Twitter page for the...")
![img_2447](https://cloud.githubusercontent.com/assets/7756031/18793745/111eb02c-8171-11e6-825a-3a7a9b5330dd.PNG)

DaleMcGrew:
This fix rocks! I love how it keeps the vertical space under control. 👍

PR #498, opened 2016-09-27
Improvements to ReadMore

zachmonteith:
Improvements include:
Organizational Descriptions, displayed in list view, now are truncated by ReadMore, with two lines visible (#491).

added regex to remove ascii control text that was interfering with handling of line breaks

ReadMore will now not truncate text if it has fewer than the maximum number of lines AND each line is shorter than 38 characters

removed parens around View Source, gave it a className so it won't display in mobile and will have mid-gray color

also changed margins slightly for twitter follow buttons/twitter followers badge so that organization cards take up less vertical real estate


DaleMcGrew:
👍

Issue #487, opened 2016-09-22
Copy link problems


DaleMcGrew:
### Please describe the issue (What happens? What do you expect?)
On iPhone 5, when I use the ShareDropdown Copy Link option, it offers a second modal, and doesn't bring the link into the clipboard.
- [x] The copy or Select All button is 1px too low on iPhone and Desktop Firefox
- [x] The Select All link doesn't work on Mac Firefox.

zachmonteith:
current behavior: on iOS, clicking "Select All" button selects all text in the input field and brings up options to copy.  clicking input field itself does NOT select all.

Android: select all selects the text, but then user has to click and hold to bring up clipboard options.

desktop: clicking input field or Select All button (if window is narrow enough to show the button) selects all text in input field, clicking "copy" button (when visible) copies to clipboard and displays copied text in red.

DaleMcGrew:
@zachmonteith I found some more issues with this while testing.

DaleMcGrew:
@zachmonteith implemented a good enough solution for this upcoming release.

PR #500, opened 2016-09-28
fix for Copy Link problem on mobile #487

zachmonteith:
new iOS update includes support for javascript copying text to clipboard, which would resolve issue #487... if we could assume that all users would be using the latest version, and that android has a similar fix.

So instead, I have created a variable, browser_supports_CopyToClipboard.  Currently set to false, with future plan to include browser/device detection and assign to true if the browser and device support the behavior.  

When browser_supports_CopyToClipboard === false, in mobile view, the copy button will not appear.  Instead a "Select All" button will appear, with the idea that the user will have the option to copy the text once it is selected using typical iOS/android cut/copy functionality.  I have also added a select all onFocus function to the input field containing the URL, so the user can click the input field OR the button to highlight the URL.

Desktop view behavior remains the same in all cases, as does the behavior if browser_support is truthy.

### DIFF
--- a/src/js/components/Widgets/CopyLinkModal.jsx
+++ b/src/js/components/Widgets/CopyLinkModal.jsx
@@ -29,7 +29,7 @@ export default class CopyLinkModal extends Component {
 
 render () {
   let urlBeingShared = this.props.urlBeingShared;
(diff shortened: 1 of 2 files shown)


DaleMcGrew:
👍

PR #501, opened 2016-09-28
document.getElementById replaces this.refs for select all #487

zachmonteith:
replaced this.refs.input.select() with document.getElementById("url-to-copy").select() in the select-all button (and onFocus of the input field) for CopyLinkModal, as the former only worked in desktop browser, not in mobile.


DaleMcGrew:
👍

PR #502, opened 2016-09-28
replaced select() with setSelectionRange(0, 999) #487

zachmonteith:
One more test: rather than invoking select, i am invoking setSelectionRange, to force the mobile browser to select the first 1000 characters in the input field, then offer typical cut/copy options.


DaleMcGrew:
👍

PR #503, opened 2016-09-28
added sizeClassName to ImageHandler, removed underline for placeholde…

zachmonteith:
…r icons, renamed vars in CopyLinkModal

This PR tackles 4 things. 
1) placeholder icons weren't always appearing at the correct size (especially on Office page when filling in for a candidate) so added prop sizeClassName, a string to be passed in, added to the class name to customize the size of the icon.

(NOTE: we were already passing in the desired className as a prop, so I could have just included size in that existing className, but figured this way could be easier to troubleshoot/customize.  happy to consolidate these two className props if people would prefer it, or that seems cleaner/DRYer/etc)

2) placeholder icons were showing blue underline on hover, so added classname "no-underline" to all anchors surrounding icons to prevent this behavior.

3) renamed fn "copied" in CopyLinkModal to "updateWasCopied", and state.copied to state.was_copied to make names more declarative/easy to understand

4) removed some deprecated class names from ImageHandler

### DIFF
--- a/src/js/components/Ballot/CandidateItem.jsx
+++ b/src/js/components/Ballot/CandidateItem.jsx
@@ -67,18 +67,19 @@ export default class CandidateItem extends Component {
     let candidate_photo_url_html;
     if (candidate_photo_url) {
       candidate_photo_url_html = <ImageHandler className="card-main__avatar"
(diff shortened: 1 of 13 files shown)


DaleMcGrew:
👍

Issue #509, opened 2016-09-29
Firefox: Weird icons on voter guide page


DaleMcGrew:
### Please describe the issue (What happens? What do you expect?)
1. On Mac Firefox, go to: https://wevote.us/theleaguesf
2. You will see many icons displayed left-to-right, when it should only show one

<img width="639" alt="screen shot 2016-09-28 at 7 55 17 pm" src="https://cloud.githubusercontent.com/assets/7756031/18939509/ab9f5154-85b5-11e6-9500-198f435c8d33.png">

ixley:
Fun! So it appears that it's something to do with the image handling. What's rendering is the image alt text 'candidate-photo' using the sizing and icon font intended for the placeholder icons. So while Webkit seems to trigger the image handler to swap out the image tag for the placeholder icons (`<i class="icon-icon-person-placeholder-6-1... >`, Firefox, in these cases is leaving the `image` tag that then renders the alt text with the icon font, (though it doesn't seem to be failing all the time since the placeholder icons do show up in some cases).

Since I'm less familiar with how the `<ImageHandler>` component works and how it does or doesn't identify the error state, I'll assume someone else can field this better than I.

DaleMcGrew:
Verified! Thank you @zachmonteith.

PR #510, opened 2016-09-29
workaround for firefox ImageHandler problem #509

zachmonteith:
it seems that firefox does not recognize a missing image source as an error, and instead replaces the image with a string (i believe the last five characters there are "photo", though I'd need to spend more time to decipher the whole string).  Rather than digging TOO deeply on how to get firefox to see this as an error, I simply added an extra statement to the conditional in ImageHandler's assignment of the constant "image", so that if there is no source for the image, we use a replacement icon.

also removed commented out, deprecated prop default assignment from ReadMore.


DaleMcGrew:
👍

Issue #506, opened 2016-09-29
Improve appearance of Search suggestion cards


nickbarry:
When the user types in the Search box, search suggestions appear below the search box. We'd like to improve the appearance of those suggestions. Each suggestion "card" should:
- [x] Be proper case ("Nancy Pelosi" not "NANCY PELOSI")
- [x] Display a thumbnail of the suggestion
- [ ] Perhaps other improvements; see how Facebook styles theirs for inspiration

zachmonteith:
Hey Nico, hope I'm not stepping on your toes on this one; as a preliminary step toward improving search results, I included the textFormat util we are using every other place that candidate or measure names are displayed to ensure they are in the proper case.  Didn't touch thumbnail images or any other improvements, in case you had already started on that, but the case issue was easy enough to solve quickly!

zachmonteith:
we will need to update the search result API to include image urls if we want thumbnails in search results.  See screenshot below (specifically, the field result_image which is coming back "" for candidates that have images)
![screen shot 2016-09-30 at 3 07 43 pm](https://cloud.githubusercontent.com/assets/17035647/19008377/c855810e-871f-11e6-81d9-e96f0ac26f73.png)

DaleMcGrew:
Hi Zach, the result_image (on the live server) now returns a value for CANDIDATE and ORGANIZATION results. For now please use a spacer image for OFFICE and MEASURE results so the search results text all lines up horizontally.

DaleMcGrew:
Looks great!

PR #511, opened 2016-09-29
added capitalizeString to SearchAllBox #506

zachmonteith:
We were using the fn capitalizeString() in all places that ballot item display names appear, other than in search results.  now, also in search results!


DaleMcGrew:
👍

PR #513, opened 2016-09-30
User position summary

ixley:
*_Sorry - this PR mixes new design with some unrelated cleanup._

- Adds position 'summary' above (user's) position statement area
  image
  - this also removes the thumbs-up/down icon from next to the card (candidate/measure) display name since it felt redundant. I left the code commented in case I overlooked cases where the two are not displayed together. We can discuss if you anticipate an issue.
  image -> image

- Introduces a 'no position' icon to distinguish between 'informational' statements or those with not explicit position and the 'mixed-rating' icon that was being used
  image

- Includes a variety of misc cleanup:
  - adds light formatting to About page
  - adds external link icons to links that open with `target="_blank"`
  - adds some new utility classes - don't use yet - I will introduce them once I settle on a convention - happy to discuss if you'd like
  - resets global/base/html font-size. (still more cleanup to do here.)


DaleMcGrew:
👍  Thank you @ixley!

Issue #396, opened 2016-08-26
Implement "Who You're Ignoring" page


DaleMcGrew:
<img width="441" alt="screen shot 2016-08-26 at 6 12 42 am" src="https://cloud.githubusercontent.com/assets/7756031/18006394/253183e2-6b54-11e6-853f-a3c1635f1829.png">

Use this page as example code:
https://wevote.us/opinions_followed

Show all of the organizations that voter is ignoring. Offer an "Edit" interface where voter can "Stop Ignoring". This will require both React code and WeVoteServer code.

DaleMcGrew:
Great work @sarah-clements! 👍

PR #517, opened 2016-10-01
improved format and structure of OpinionsFollowed

zachmonteith:
In this PR:
1) changed view source links to reflect an earlier PR: mid-gray color, no parens.

2) moved route OpinionsFollowed to the root level

3) switched OpinionsFollowed from using deprecated VoterGuideItem to have structure more similar to Opinions: similar to GuideList, created component OpinionsFollowedList.

4) added "editMode" state to OpinionsFollowed, so users can toggle whether or not they are following particular orgs from this list.

5) made CopyLinkModal input field read-only to prevent a console warning

NEXT STEPS
- [ ] add a route (and a link to it on OpinionsFollowed) for orgs/users you are ignoring #396 
- [ ] more elegant solution to the array vs object prop type issue (see comment on line 50 of OpinionsFollowedList)
- [ ] related to last item, formatting of OpinionsFollowedList: button children spacing still appears even when edit mode is true and buttons are turned off (see screen shots below)
screen shot 2016-10-01 at 12 51 19 pm
screen shot 2016-10-01 at 12 51 29 pm

### DIFF
--- a/src/js/Root.jsx
+++ b/src/js/Root.jsx
@@ -26,7 +26,7 @@ import Measure from "./routes/Ballot/Measure";
 import NotFound from "./routes/NotFound";
 import Office from "./routes/Ballot/Office";
 import Opinions from "./routes/Opinions";
-import OpinionsFollowed from "./routes/More/OpinionsFollowed";
+import OpinionsFollowed from "./routes/OpinionsFollowed";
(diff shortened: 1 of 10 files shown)


DaleMcGrew:
👍

Issue #526, opened 2016-10-04
Create dynamic, SEO optimized <title> and <description> tags for all relevant routes/pages


zachmonteith:
Currently, We Vote web app only has one title and one description for the entire app.  The title, regardless of what url users access or what is rendered in DOM, appears as simply "We Vote", and the description text is "Putting you in the driver's seat of American democracy."

Ideally, we would have different titles for each section of the site (perhaps they could be not dissimilar from headings on pages; candidate page could have the display name as title, "Your Ballot" for ballot page, office name for offices, etc).  would still be good to have "We Vote" appear in title, could be something like "Your Ballot - We Vote".  Might also be fun to include a favicon? Though less necessary for driving traffic in from google searches than unique titles.

Having the same description for all pages seems like less of an issue (the description text is what appears as a sort of subtitle in google results) as long as it is catchy, but could be even better to make it unique to each page (something like "Positions on {candidate_name}, candidate for {office_name}" or something like that.

zachmonteith:
SO question posted by our own @pertrai1 back in January on the topic of dynamic titles in react:
http://stackoverflow.com/questions/35106647/dynamic-header-titles-with-react

zachmonteith:
also useful http://stackoverflow.com/questions/26266675/how-do-i-keep-document-title-updated-in-react-app

ixley:
I imagine the technical implementation for this will be relatively simple compared to the work of thoughtfully crafting contextual descriptions for different route contexts. 

Creating a Google Doc where we can outline: https://docs.google.com/document/d/1S8Q9ZbBHh2VwYTHh2Kc60PLkEIji_QXQusWbVYutuek/edit?usp=sharing

DaleMcGrew:
Hi @zachmonteith, can we consider this issue complete now?

zachmonteith:
👍

PR #529, opened 2016-10-04
added Helmet for title and description #526, created Credits page

zachmonteith:
1) Added nfl's react-helmet component to most unique route pages, to give each its own title; tried to account for all titles we would need, and give descriptive, attractive titles.  

2) Added descriptions to Candidate, Office, Measure and GuidePositionList as well.  These are placeholder descriptions until Colette composes final ones, but easy to update, and gives a sense of the pattern, so can easily be applied to other appearances of Helmet where we want a page description.

3) created Credits route, took most of content from About and moved to Credits

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,6 +36,7 @@
     "react-burger-menu": "^1.10.4",
     "react-copy-to-clipboard": "^4.2.3",
     "react-dom": "^0.14.3",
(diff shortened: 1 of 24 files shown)


DaleMcGrew:
👍

PR #530, opened 2016-10-05
Split out sign out from VoterActions to VoterSessionActions. Started adding Sign In By Link process. (Still needs work.)

DaleMcGrew:


DaleMcGrew:
👍

Issue #532, opened 2016-10-06
Office Pages stuck in infinite Loading


luojxxx:
### Please describe the issue (What happens? What do you expect?)
The page makes the http request to the api and gets back a response with the office object. However, the page does not load.

The issue may be the way the information is passed through the actions/stores. Researching the issue further.

### Steps to reproduce the problem (1, 2, 3...), including links
1. Go to the wevote main page
2. Type in president in the search bar
3. Click on the first/only drop down from the search bar, it's the president as a quick link (http://wevote.us/office/wv02off1)
4. As you can see the page doesn't load but the page does get http response from the api containing the necessary information.

luojxxx:
Further digging shows that the object is being retrieved successfully in OfficeStore, so it seems to be an issue with it not being passed properly to the Office view. Will try to implement a fix.

luojxxx:
The getOffice function within OfficeStore is trying to
`return office_list[office_we_vote_id]`
However, office_list is only a single object without an ID/key. So currently it errors here and _onOfficeStoreChange receives and undefined variable. So by using
`return office_list`
_onOfficeStoreChange is now able to correctly retrieve the office object and set it's state.

However, for some reason setState does not re-render the page. You must refresh in order to display the information.... Hmmm.....

DaleMcGrew:
Fixed

PR #536, opened 2016-10-07
Office Pages stuck in infinite Loading #532 Hot Fix

luojxxx:
office_list is only a single object value (without a key or ID), which
was causing an error in getOffice resulting in an undefined variable
for _onOfficeStoreChange. Therefore returning the “office_list” itself.

Furthermore after the initial object property is loaded.
componentWillReceiveProps does not perform officeRetrieve when someone
navigates to another office page causing the new data to not be
displayed. Therefore added an additional officeRetrieve in the else
clause as well.

Likely we’ll need to dig into these later on to clean it up a bit more.
Also the Office page is still missing the ballot and candidate data.

### DIFF
--- a/src/js/routes/Ballot/Office.jsx
+++ b/src/js/routes/Ballot/Office.jsx
@@ -34,6 +34,7 @@ export default class Office extends Component {
       this.setState({office_we_vote_id: nextProps.params.office_we_vote_id});
       OfficeActions.officeRetrieve(nextProps.params.office_we_vote_id);
     } else {
+      OfficeActions.officeRetrieve(nextProps.params.office_we_vote_id);
       this.setState({office: office, office_we_vote_id: nextProps.params.office_we_vote_id});
     }
 
--- a/src/js/stores/OfficeStore.js
+++ b/src/js/stores/OfficeStore.js
@@ -7,7 +7,8 @@ class OfficeStore extends FluxMapStore {
     // if (!this.isLoaded()){ return undefined; }
     let office_list = this.getState().offices;
     if (office_list) {
-      return office_list[office_we_vote_id];
+      // return office_list[office_we_vote_id];
+      return office_list;
     } else {
       return undefined;
     }


DaleMcGrew:
Thank you @luojxxx -- I look forward to reviewing your fix in the next few days. (I am crunching to get ready for our launch tomorrow right now.)

luojxxx:
No problem, let me know if you have questions!

zachmonteith:
Hey @luojxxx, I was curious about these changes; you have changed the OfficeStore to return the entire office_list, instead of only the office in the list located at the key of the office ID.  I could be mistaken, but I think the office_list here should contain all offices appearing on the user's ballot (or more accurately, all of the offices that are stored in state when we assign office_list on line 8), so we need to key into it in order to only retrieve the office we want to see, and only display the candidates running for that office.  @DaleMcGrew, can you offer any clarity here?

Your change in Office.jsx seems right, although rather than including the same "OfficeActions.officeRetrieve(nextProps.params.office_we_vote_id);" in each branch of the conditional, I would just move it outside of the if/else (in this version, that would be line 32, but if you pull the latest version from upstream I believe it becomes line 34 or 35): both ways work, but one is less redundant.

For me, this seems to fix the infinite loading bug in some cases, but not others; I will keep playing with this, and encourage you to do the same!!

luojxxx:
@zachmonteith I've went to a few offices and they all had the same issue with only a single object within the Officelist. It's also appears to have some kind of sub-object as one of the properties. It's an odd problem, and would be better suited for someone slightly more familiar with the code as it appears to touch a variety of things. Otherwise I'll work with you guys on it the next time I come in, most likely next week.

And I wasn't sure about the idea of putting that line of code outside the conditional branch since that would affect the order of execution for the if-branch. Not quit sure if it matters, but some of the syntax/methods are unfamiliar, still need to catch up.

zachmonteith:
oh, I see your point about order of operations, I missed that on my first pass.  And no worries, its definitely an odd problem and I wouldn't want to discourage you from contributing!

DaleMcGrew:
Hi @luojxxx and @zachmonteith I solved this problem in Pull Request #552.

PR #538, opened 2016-10-08
Content and formatting for about page, including image badges

zachmonteith:
About page now includes text supplied by Colette and bios, with circular images for each person profiled.


DaleMcGrew:
👍

Issue #454, opened 2016-09-09
Mobile (Safari): Can we add "momentum" to scroll?


DaleMcGrew:
### Please describe the issue (What happens? What do you expect?)
When I am trying to scroll down the ballot while using iPhone Safari, or a list of candidates, there isn't any scrolling momentum. (It doesn't seem to be a problem with iPhone Chrome.) It would feel better if you could flick up, and have the page keep scrolling and then slow to a stop.
### Steps to reproduce the problem (1, 2, 3...), including links
1. Set your address to somewhere in the San Francisco Bay Area.
2. Go to your ballot
3. Try to scroll down the ballot quickly with your finger
4. The scrolling feels sluggish, and the scrolling stops as soon as you let go

ixley:
@DaleMcGrew I think this might just be a browser thing. Scrolling has the momentum you describe in Chrome, but not in Safari – seems specific to the (mobile) browser, as opposed to the website.

ixley:
*I haven't dug into it but `-webkit-overflow-scrolling: touch` might do the trick.

ixley:
I found another factor. Seems like triggering the react burger menu is a/the cause of this.

DaleMcGrew:
Hi Jeff, good catch, I can confirm that the scrolling slows down on iPhone Safari only AFTER you open the menu. I haven't tested since the new update. I'll do that shortly.

DaleMcGrew:
Verified! Woohoo!

PR #541, opened 2016-10-10
update react-burger-menu

### LINKED ISSUE #450: react-burger-menu breaks react-headroom
[react-burger-menu](https://github.com/negomi/react-burger-menu) currently breaks [react-headroom](https://github.com/KyleAMathews/react-headroom) functionality after the menu has been triggered.

react-burger-menu injects `overflow-x: hidden` into the `<html>` and `<body>` tags when the menu is triggered, but does not remove them on close.

Investigate whether the problem can be solved on our end (implementation fix) or whether the fault is with the react-burger-menu code, in which case, please [file an issue](https://github.com/negomi/react-burger-menu/issues) with them and report it here.

ixley:
updates react-burger-menu to address a bug that fixes #450 and #454


DaleMcGrew:
👍

PR #543, opened 2016-10-13
minor improvements to about page, added credits to hamburger menu

zachmonteith:
Changes to About page reflect notes emailed to me by Colette:

- more space between images and names
- links to Mozilla and Code for SF open in new tabs, and contain new tab icon
- sample PositionPublicToggle defaults to Friends Only to more closely mirror typical behavior
- typo fixes: organization/organizations, Americaand/America and, who've/who have.
- removed <p> tag that was causing errors in console

Also added credits page to hamburger menu.  Didn't change "Suggestions?" to "What would make We Vote better?" (this is what appears in desktop mode) as I wasn't sure that whole string would fit in the hamburger menu, or look good; however, I did force it to open in a new tab, and to display the new tab icon.

### DIFF
--- a/src/js/components/Navigation/HeaderBar.jsx
+++ b/src/js/components/Navigation/HeaderBar.jsx
@@ -121,9 +121,9 @@ export default class HeaderBar extends Component {
               </Link>
             </li>
             <li>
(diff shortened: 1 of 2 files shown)


DaleMcGrew:
👍

PR #547, opened 2016-10-14
Cleaned up the routes/Process/FacebookSignInProcess to be a lot more stable.

DaleMcGrew:
Cleaned up the routes/Process/FacebookSignInProcess to be a lot more stable. Added BrowserPushMessage to more pages. Added Loading wheel immediately following browserHistory.push to prevent code from continuing to run while redirect is happening. Did a number of eslint cleanups.


DaleMcGrew:
👍

Issue #553, opened 2016-10-17
Candidates don't show the first time you click office on new ballot


DaleMcGrew:
### Steps to reproduce the problem (1, 2, 3...), including links
1. Start with a new account.
2. Enter a new address, and save.
3. Go to your ballot and click on any office. (Make sure you have not refreshed your browser since starting this process.)
4. You will not see any candidate data (see photo below).
5. When you refresh the browser, candidate data appears.

![screen shot 2016-10-16 at 8 33 56 pm](https://cloud.githubusercontent.com/assets/7756031/19424832/93788c64-93e1-11e6-9a91-942c878dc6a8.png)

DaleMcGrew:
Verified

PR #556, opened 2016-10-18
Added BallotItemsRetrieve to AddressSave, resolves Issue #553

zachmonteith:
Added a voterBallotItemsRetrieve invocation to the case for voterAddressSave in the VoterStore; we weren't retrieving all the ballot items when users saved their address, necessitating a browser refresh to trigger this action and populate the user's ballot.


DaleMcGrew:
👍

Issue #559, opened 2016-10-18
measures should display on which ballot (city, state, election) they appear: phase 2


zachmonteith:
See issue #558 
phase 2: API call should return what City (or other jurisdiction, whatever is relevant/accurate) ballot measures belong to, for display on that measure's page.

zachmonteith:
copied from PR #561 comments:
TODO: #559 Measure API should be updated to include election_display_name, regional_display_name, state_display_name.  sample format:
election_display_name = "November 2016 General Election"
regional_display_name = "Oakland"
state_display_name = "California"

Issue #558, opened 2016-10-18
measures should display on which ballot (city, state, election) they appear: phase 1


zachmonteith:
Measure pages should indicate their state, city if relevant, and in which election they'll appear.  This is especially critical if a user searches for and finds a measure that isn't on their personal ballot, as it won't be clear where and when this measure will be voted on. 

Phase 1 will be to include the info already being passed in from the API call: State and Election ID.

DaleMcGrew:
Verified. Great work @zachmonteith!

Issue #557, opened 2016-10-18
Measures should display measure_text with ReadMore truncation


zachmonteith:
measures have expanded text attribute, we should display this on the measure page, but with the ReadMore truncation to keep it neat and not take up too much vertical real estate.

DaleMcGrew:
Verified.

PR #561, opened 2016-10-18
added measure_text and election + location info to MeasureItem #557 #558

zachmonteith:
Updates to MeasureItem: 
1) now displays measure_text inside a ReadMore, in our mid-gray color; 
2) now displays the correct election for the measure, conditionally on the following props passed from API call: election_display_name, regional_display_name, state_display_name.  

TODO: #559 Measure API should be updated to include election_display_name, regional_display_name, state_display_name.  sample format:
election_display_name = "November 2016 General Election"
regional_display_name = "Oakland"
state_display_name = "California"

if there is no election_display_name, measures will display "Appearing on the ballot in", and if no state_display_name, capitalizes state_code that already lives in the measure object.

### DIFF
--- a/src/js/components/Ballot/MeasureItem.jsx
+++ b/src/js/components/Ballot/MeasureItem.jsx
@@ -1,9 +1,9 @@
 import React, { Component, PropTypes } from "react";
 import { Link, browserHistory } from "react-router";
-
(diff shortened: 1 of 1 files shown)


DaleMcGrew:
👍

Issue #567, opened 2016-10-19
Add "Edit" mode on Friends listing page


DaleMcGrew:
Problem -- it is too easy to click "Remove Friend" when browsing through friends.

On the "Friends" page: https://wevote.us/friends, emulate the way the Edit/Done Editing works from this page: https://wevote.us/opinions_followed

DaleMcGrew:
Verified!

PR #568, opened 2016-10-19
Added edit mode for Friends page #567, improvements to CopyLinkModal #487

zachmonteith:
In this PR:
1) added editMode state to Friends page, which gets passed as a prop to the FriendDisplayForList, through the FriendList.  Toggling edit mode will toggle whether the Friend/Unfriend toggle buttons appear next to each friend in a user's list, similar to the edit mode pattern established for organizations a user is following.

2) removed Select All button from CopyLinkModal: it was only being used on mobile browsers, and not all mobile browsers supported it.  Also changed the field containing the URL from a read-only input to a read-only textarea, so that users can more easily see the entire URL string and highlight/select it themselves, to account for cases where the select all script wasn't working properly (most mobile browsers).  Modified appearance of Copy button to match new text area.

left structure in place for future device testing: CopyLinkModal contains a variable named browser_supports_CopyToClipboard, if truthy, copy button will always appear regardless of window size. Default behavior for now will be to only display copy button on max nav browser windows, as it seems to work properly in all desktop browsers.

### DIFF
--- a/src/js/components/Friends/FriendDisplayForList.jsx
+++ b/src/js/components/Friends/FriendDisplayForList.jsx
@@ -14,7 +14,8 @@ export default class FriendDisplayForList extends Component {
     voter_display_name: PropTypes.string,
     voter_twitter_handle: PropTypes.string,
     voter_twitter_description: PropTypes.string,
-    voter_twitter_followers_count: PropTypes.number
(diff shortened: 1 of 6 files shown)


DaleMcGrew:
👍

PR #569, opened 2016-10-19
Cleaned up how we see Friend Requests. Updated sign out routine. Fixed problem with TwitterSignInStart.

DaleMcGrew:


DaleMcGrew:
👍

PR #571, opened 2016-10-20
Develop

luojxxx:
Includes basic core coverage of website. Going through three separate
tests: 1. Ballots supporting, bookmarking, remaining, location 2.
Requests, connect, activity, following, following search. 3. Search bar

There is an issue with getting the automated script to wait properly
without refreshing, will troubleshoot. And then will start looping it
through various platforms/browsers.

Pulling to put mostly ready testing scripts into the repository for others to play with.

### DIFF
--- a/src/js/routes/Ballot/Office.jsx
+++ b/src/js/routes/Ballot/Office.jsx
@@ -34,6 +34,7 @@ export default class Office extends Component {
       this.setState({office_we_vote_id: nextProps.params.office_we_vote_id});
       OfficeActions.officeRetrieve(nextProps.params.office_we_vote_id);
     } else {
+      OfficeActions.officeRetrieve(nextProps.params.office_we_vote_id);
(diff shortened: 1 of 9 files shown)


luojxxx:
Whoops it might be still trying to push through the office pages commits.

luojxxx:
Oh fixed the refresh problem. By setting an element variable to time.sleep().

luojxxx:
Oh never mind, I was just copying the syntax from the selenium example exactly to try to make this work and was wondering maybe for some reason it was looking for this specifically.

So fyi the truth is, in Saucelabs if you press Play through the commands, there's a long refresh between step, which appears during the time.sleep. However, this only occurs the first time you watch it. What was happening was that every time that occurred, I tried running another test, to try something different to fix it, so I was always watching the first play through, which was probably caching the image or something.

Also it seems if you just watch the video itself, there's no problem.

Issue #565, opened 2016-10-19
Instant change: Public/Private toggle


DaleMcGrew:
Currently when you click the public/private toggle, it does not change until we get a response back from the API server.  Update the click handler to instantly change the public/private toggle so we don't have to wait until the API server response.

DaleMcGrew:
Verified

Issue #564, opened 2016-10-19
Instant change: Follow org


DaleMcGrew:
Currently when you click the follow button, it does not change until we get a response back from the API server.  Update the click handler to instantly change the follow button to following so we don't have to wait until the API server response.

DaleMcGrew:
Verified

PR #576, opened 2016-10-20
instant toggle changes #564 #565

zachmonteith:
in PositionPublicToggle and FollowToggle, we now invert the variable used to determine the appearance of the toggle before making our API call.  this way, the button will change appearance as soon as the user clicks it.  this variable is still reassigned once the response comes from the server, so it will still accurately reflect the state once the cycle is complete, but will respond instantly to click so users don't get frustrated and start mashing the button multiple times.

also eliminated the "transitioning" state from PositionPublicToggle as this was preventing the toggle from rendering until the transition was complete.


DaleMcGrew:
👍

PR #578, opened 2016-10-20
updated fix for #564, handles closure without breaking button

zachmonteith:
last PR had some issues with scope and closure of FollowToggle, this fix reassigns is_following variable and executes follow/stopfollow functions without modifying them or the scope in which they reside.


DaleMcGrew:
👍

PR #590, opened 2016-11-03
Entered possible fix to not being able to make certain voter guide entries public. Changed sign in message text.

DaleMcGrew:


DaleMcGrew:
👍

Issue #227, opened 2016-04-23
Accessibility: Add ability to tab to bookmarks and support/oppose


DaleMcGrew:
1. Go to the ballot
2. Tab through to the page and you find that you skip past bookmarks and support/oppose
   A person should be able to use the keyboard to get to all interface elements.

wendy-c:
works on chrome, still skipping support/oppose on safari

zachmonteith:
update: hard coding in a tab index through dev tools does not cause firefox to highlight candidate name, so that theory is debunked

ixley:
@DaleMcGrew This looks like it has been addressed. On a related note – tabbing typically only jumps to links and buttons, so text that has a click handler attached, but has no `href` will not work. On /ballot, notice that tabbing catches the 'office' text links but not the candidate ones.

DaleMcGrew:
Hi @ixley, @nickelser is still working on this. We got it working on all browsers except Safari, and Nick tracked down a possible fix that he is working on.

ixley:
ah okay - sorry about that.

On Sun, Apr 30, 2017 at 10:01 AM, Dale John McGrew <notifications@github.com

nickelser:
the problem actually stems from the default safari behavior -- which doesn't match any of the other browsers.

this SO answer describes the solution: https://stackoverflow.com/a/1914496

we could notify safari users that they should change their settings, but i assume that anyone who has accessibility needs and uses safari would already have it enabled.

i recommend actually not changing any code or HTML and instead trusting the users 😄

PR #593, opened 2016-11-17
fix for #227, bookmark can now be tabbed to

zachmonteith:
Thanks to Adit for pairing with me at Code for SF to solve this one!  in case the commit name wasn't clear, this allows users to use the tab key to highlight the bookmark button for the ballot item they are viewing, and to press the enter or space key to bookmark the item once it is highlighted.

*EDIT*
added similar functionality to all buttons, links, and other clickable objects I could find that weren't previously focusable with tab and didn't trigger on spacebar or enter.


DaleMcGrew:
👍  Great work @zachmonteith and Adit!

PR #609, opened 2017-01-06
Onboarding animation setup with animation story components and slider…

sarah-clements:
Onboarding animation setup with animation story components and slider; starting using timeline with GreenSock; Had to add slick-carousel CSS links rather than rely on npm install react-slick, since react-slick requires higher version of react.


DaleMcGrew:
👍

PR #610, opened 2017-01-07
Update React to 15.4.2 along with other Packages

pertrai1:

All of the packages have been updated to the latest version minus Flux.
Flux has breaking changes that will have to be addressed in more depth.

There are a couple app specific errors that are happening away from the
update:

* Warning about ReactPlayer on about
* Share -> Copy has warning

* Update to all packages to latest minus Flux
* Update to `input` fields that need to use `defaultValue` rather than `value`


DaleMcGrew:
Thank you Rob! I pulled this version to my local machine, and made some fixes and additional updates which I will commit in another Pull Request. 👍

PR #614, opened 2017-01-13
Onboarding animation of first sequences 

sarah-clements:
Onboarding animation of first sequences (FaceBook and Twitter signin components added) with assets; created media queries, added x(close) component with a redirect to sample ballot route and added all assets for first animation section; Tried react-slick nextSlide method but it creates an infinite loop - usage TBD; react-slick CSS CDN replaced with slick-carousel package.


sarah-clements:
@ixley wanted to look at my CSS overrides (in theatre.scss) before the pull request is approved.

ixley:
@sarah-clements I was somehow able to push directly to your branch (which was not my intention!) I made some sample edits to go along with my comments/questions. My changes (which I meant to push to my own fork) are meant as reference for some of the suggestions/comments I made, but don't feel like you have to integrate them. It also is not meant to be integrated 'as is' since I didn't address everything I commented on. 

I don't want to further muck about on your branch so I'm not going to try to roll it back, but feel free to use or not use what you want. Full comments below. Do let me know if you have any questions.

**Design**
I’m not going to comment too much on design since it appears that it’s not really finished, but here are a couple thoughts/questions.
* The content design still looks like a work in progress… Should all the content assume to be placeholders?
    * The sign in buttons, for instance, are not laid out responsively (they overlap on small screens) and also probably shouldn’t be on the first screen of the intro/tour. 
    * Feels strange that text shows up and then disappears on its own (rather quickly).
* The swipe interaction is quite appropriate for touch devices, but is less appropriate for desktop browsers. Consider adding more appropriate navigation UI—like (arrow) buttons—for non-touch devices.

**Implementation**
In general, try to use (or extend) existing conventions, when possible.
* Use the existing ‘breakpoints’ mixin (along with existing $breakpoints maps) for media queries. The mixin is a useful helper (allowing you to nest media query styles in context) and using standardized variables when possible helps reduce unnecessary custom values. (426px seems like a pretty random value, for example.)
* Don’t apply styles to element tags (.intro-story h2) – only style with appropriate classes. (This is just good practice for CSS.)
* Try to use default or existing styles as much as possible. I’m not clear what purpose .example-header and .example-header-2 serve, especially with placeholder content. Use Bootstrap heading classes for now (.h1, .h2, etc.) It’s on my list to add additional custom heading classes soon too.
* Avoid nesting selectors when possible. Use BEM naming conventions to apply custom styles and keep specificity low.
* I’m not familiar with the plugin being used, but it seems strange that there are 3 separate components needed for the different screens. Is it not possible to add screen content with the same component? Again, I’m not familiar—it just struck me as a strange setup. Disregard if this comment is out of place.
* Too many .row elements. You only need .row to setup column width layouts—not to wrap content otherwise. The ones in AnimationStory are essentially unnecessary. See Bootstrap documentation for usage.
* Some comments on the [x] close button: 
    * Consider using Bootstrap’s .close button styling (or something similar). I realize that this implementation calls for a white/light close ‘x’ instead, so consider creating a variant based off the Bootstrap style.
    * Or, if you want to stick with the image, if you are going to set it up as a global element, set the styles up as global as well and don’t nest them in a component specific partial. You should basically decide whether this element/component is scoped to only this section or whether it’s global and stay consistent with how it’s named, made accessible, organized, etc.
* As with the x-close button, try to stay consistent with naming and file structure. It’s confusing that the sass partial is named _theater.scss and the main style within is .intro-story. A component or section based partial should typically be named the same as it’s primary component or route name.

sarah-clements:
Pull request #621 addresses the design/stylesheet issues mentioned in the comments.

Issue #608, opened 2017-01-04
On "About We Vote" page, create grid with team photos


DaleMcGrew:
- [x] Add photos from both c4 (www.WeVoteUSA.org) and c3 (www.WeVoteEducation.org) boards, to the "About We Vote" page. Photos that are not on either of these sites will be placed here: https://drive.google.com/drive/u/1/folders/0B3O7mejbtJ3qc0M2WHh4dFZHdmc
- [x] In desktop the photos should be 4 across
- [x] In mobile the photos should be 3 across
- [x] Name and title of each person should be beneath photo. 
- [ ] There should be a "Read more" link that opens up a Div that floats above the page with the photo, title, and full bio from the person.

We need two sections, with these people left-to-right, top-to-bottom:
We Vote Boards
- Jenifer Fernandez Ancona, c4 Board Chair
- Tiana Epps-Johnson, c3 Board Chair
- Debra Cleaver, c3 Board Member
- Tory Gavito, c4 Board Member
- Lawrence Grodeska, c3 Board Member
- Dale John McGrew, c4 & c3 Board Member
- Anat Shenker-Osorio, c4 Board Member
- William Winters, c4 Board Member

We Vote Staff
- Sarah Clements, Engineering Intern
- Irene Florez, Engineering Intern
- Jeff French, Lead Designer
- Anisha Jain, Engineering Intern
- Judy Johnson, Operations
- Neelam Joshi, Engineering Intern
- Ciero Kilpatrick, User Experience Intern
- Dale John McGrew, Executive Director
- Colette Phair, Manager of Communications and Community Outreach

DaleMcGrew:
Great work @mnneelam!

PR #616, opened 2017-01-18
Added photos for WeVote board and team members under img/global.

mnneelam:
Modified About page with following changes:
1) Separate sections for WeVote Board and WeVote Staff - Name, Title, Brief description
2) Responsive display based on different breakpoints - 3 photos for small screen, 4 for large screen
3) For small screens, hide description

Iteration on issue #608


DaleMcGrew:
Thank you Neelam! I will make some changes to the content before we post it to the live site. 👍

Issue #613, opened 2017-01-12
Add Mailchimp email newsletter signup form to About page


ColettePhair:
Create email newsletter form fields that lets users sign up for our 501(c4) nationwide email list on Mailchimp, to add to We Vote About page

DaleMcGrew:
Verified. Thank you @zachmonteith!

PR #618, opened 2017-01-19
created email newsletter signup form button #613

zachmonteith:
Created a basic form using Mailchimp's tools, lives on their server.  created button on more/about page that opens that form in a new tab.  Also added a light border to medium button so that it matches the other buttons.


DaleMcGrew:
Thank you @zachmonteith and Randy! 👍

PR #619, opened 2017-01-19
About page fixes for col-sm and visible-sm-block and visible-lg-block.

DaleMcGrew:


DaleMcGrew:
@mnneelam it turns out we weren't including conditions for "sm" or "lg" sizes. I added these and it is working in all screen sizes. 👍

Issue #615, opened 2017-01-17
Create basic interaction for an 'intro tour' to We Vote


ixley:
Add functionality for a basic 'intro tour' for first-time We Vote users to orient them to the benefits and basic features of We Vote. In the absence of design or content, focus on basic interactions and features agnostic to content.

The goal of this feature will be to give first-time visitors/users a basic overview of what high-level benefits/abilities to expect. It should _not_ attempt to orient them to specific functionality.

## Overview
- Tour will be a multi(3-4)-screen 'app tour'
- Should be responsive (work on all screen sizes and accommodate both touch devices and desktop browsers)

### UI/Interaction
- Swipe left to advance on touch/mobile devices
- Dots to indicate position/progress
- Include more prominent clickable arrow(s) for non-touch devices
- Make full-screen on smaller screens, and constrain to a reasonable TBD size on larger screens (single breakpoint for responsive design)

### Additional considerations
- Don't worry about content but assume a single call to action (registration) on (only) the last screen. That is to say, focus primarily on the content containers for the time being. If the plugin supports things like nice transitions within a screen, that's fine, but I wouldn't prioritize it.
- If selecting a plugin, ideally focus on something well-supported (popular), content agnostic, and easy to customize (something with bare bones and editable out-of-the-box styling and potentially customizable behavior)

## Reference
- [Basic design brief](https://docs.google.com/document/d/1QEzIu8GIuxcXkcJQQlUYl7BNQc9AeedwZu3Ktjl8f7g/edit#)
- [In-progress prototype](https://app.moqups.com/DaleMcGrew/Xa9gsWkvx0/view/page/a0dc9d6cd)

sarah-clements:
Per another of Jeff's suggestions, there should be a clear sense of what happens above a certain screen size (ideally determined and conveyed by the designers).

ericogawa:
Something along these lines, perhaps? (from the Patreon responsive website) (Attempting to upload screen shot)
![screen shot 2017-01-18 at 12 00 32 pm](https://cloud.githubusercontent.com/assets/13991758/22080807/ce828ad4-dd75-11e6-9ea7-d787ffa15012.png)

ericogawa:
I suspect with the proper image file(s), we could resize all the way to full screen on mobile. We'd have to find some images that are aesthetically functional across a wide range of "croppings"

sarah-clements:
The desktop snapshot looks great with the arrows. For mobile, I've gotten
[...]

PR #621, opened 2017-01-19
Onboarding animation of first sequences REVISED

sarah-clements:
Onboarding animation of first sequences (FaceBook and Twitter signin components added) - timing of animations and styling will be finalized with real content/assets; created media queries, added x(close) component with a redirect to sample ballot route; Tried react-slick nextSlide method with animation onComplete method but it creates an infinite loop - usage TBD; react-slick CSS CDN replaced with slick-carousel dependency; arrow menu navigation added for desktop; cleaned up stylesheet.

Addresses comments in issue #614 and design spec issue #615


DaleMcGrew:
This is a rework of an earlier pull request. It is not ready to go live, but is a good framework for the on-boarding experience. This is related to Issue #615. Once these files get pushed to the live server, we can test on a variety of mobile devices. Thank you @sarah-clements and @ixley! 👍

src/sass/components/_intro-story.scss
```diff
@@ -0,0 +1,79 @@
+.intro-story {
+  background-color: $gray-dark;
+  color: #fff;
+  text-align: center;
+  padding: 0;
+  height: 500px;
+  width: 100%;
+  @include breakpoints (small large){
```
ixley: @sarah-clements FYI, this works, but consider the following...
This query scopes to min-width: [small], max-width: [large], which in this case is excessive. While you are unlikely to have a device/screen narrower than `small` (320px), there is no reason to constrain the media query to that min-width. You could simplify by simply declaring `large` (959px) to be the max-width (`breakpoints(max large)`)
...but also...
consider forming media queries (by default) so that the smaller screens get the default styles and then they are overridden by styles for larger screens. This is not important in this specific line, but the practice has other implications that make it a good practice.
So instead of setting `width: 500px;` as the default, consider dong this instead:
```
width: 100vh;
@include breakpoints(large){ // min-width: 960px
  width: 500px;
}
```
Make (at least some) sense?

src/sass/components/_intro-story.scss
```diff
@@ -0,0 +1,79 @@
...
+    height: 100vh;
+  }
+}
+
+.intro-story__background-image {
+    //Image for desktop has to be a very specific size otherwise it will tile or wrap
+    // onto the next slide (due to div width of react-slick carousel) - 1118 width works fine
+    background-image: url("/img/global/animations/3.1/flag_background_lg.png");
```
ixley: @sarah-clements This is actually a case where the media query scoping I mentioned above (styling for smaller screens as the default and then scoping styles for larger screens in queries) becomes relevant. As it's currently set up, a small screen (or device) will only render the smaller image but still likely download both! Not what you want, especially if you're targeting mobile devices that are likely to have lower bandwidth!
Check out this for examples: https://timkadlec.com/2012/04/media-query-asset-downloading-results/

src/sass/components/_intro-story.scss
sarah-clements: Thanks for the feedback - noted!

src/sass/components/_intro-story.scss
sarah-clements: I will look into this - thanks!

PR #625, opened 2017-01-26
positions in network bar now shows red and green

zachmonteith:
After design talks with Cara and Dale, this PR makes the following change to the "positions in your network" color bar:
if there are people in your network who oppose AND who support a ballot item, the bar will show the percentage who support in green and who oppose in red (e.g. if 2 support and one oppose, the bar will be 2/3 green and 1/3 red)
if there are no positions in the network, it will remain the default light gray, and if all members of your network support or oppose the entire bar will still show the appropriate color.


DaleMcGrew:
👍 Great work Zach!

Issue #507, opened 2016-09-29
Use react patterns instead of searching the DOM


nickbarry:
There are many cases when we're searching the DOM for elements, instead of using proper React patterns to modify the DOM. This is slow. It would also be confusing for other React developers (including our future selves, looking at the code and trying to figure out how it works!).

For example, see the `src/js/utils/search-functions.js` file.

nickbarry:
I've assigned this to myself, but it's probably not super high priority. If someone else comes across this issue, and I'm not clearly in the middle of it (i.e., I haven't left recent comments on the issue), feel free to work on it yourself.

DaleMcGrew:
Thank you @nickelser!

PR #626, opened 2017-01-26
No dom for search

nickelser:

https://github.com/wevote/WebApp/issues/507

Address #507 by moving some global actions to the store.
This required quite a bit more heavy lifting than I would have liked, but this also added:

* keyboard support for the search box (
* more react-y state and transitions within the search box
* more react-y CSS/transitions


DaleMcGrew:
Thank you @nickelser, I verified this locally. Great work! 👍

Issue #624, opened 2017-01-26
Ballot page is missing margins/padding on the left & right


flobotbear:
### Please describe the issue (What happens? What do you expect?)
When you make the viewport more narrow (less than 692px -- or whenever the left sidebar is hidden), the right column with "US 2016 General Election" tex and the candidate cards lose padding/space on the left and right sides.

![ballot_-_we_vote-nomargins](https://cloud.githubusercontent.com/assets/823927/22319691/ec217bda-e339-11e6-8f19-acfb37d0b332.png)

### Steps to reproduce the problem (1, 2, 3...), including links
1. Go to https://wevote.us/ballot
2. Make your viewport narrow enough to lose the left sidebar

This has been observed on the following browsers:
- Chrome
- Firefox
- Safari
- Mobile Chrome

Interestingly the screen before you put in the zipcode for where you are registered to vote, the padding/spacing is still there.

flobotbear:
Dale helped me pinpoint the issue -- it seems to be coming from ``` './components/intro-story';``` bc when we comment that out, the spacing is back. Will make sure the fixes don't end up messing up the work for the intro

DaleMcGrew:
Resolved!

PR #629, opened 2017-01-28
Fix padding problem on some views

pertrai1:

https://github.com/wevote/WebApp/issues/624

Looks like the intro story was overriding the container-main padding so
there is an introduction to a class that is added for the intro story
view on the body that encapsulates the container-main and then removes
that class on any other view so that the container-main will inherit the
regular padding that it has.


DaleMcGrew:
Thank you Rob! This is a very elegant solution. 👍

PR #630, opened 2017-01-28
Fixed a Facebook sign in loop. A few language changes. Update to sign in page to get it closer to the new design.

DaleMcGrew:


DaleMcGrew:
👍

PR #631, opened 2017-01-28
Fixed bug with signout. Added code to create a new voter_device_id if one is used that is no longer recognized by the server. Updated left navigation by removing excess options.

DaleMcGrew:


DaleMcGrew:
👍

Issue #574, opened 2016-10-20
Instant Search: Voter entering address


DaleMcGrew:
Upgrade the https://wevote.us/settings/location "Enter address where you are registered to vote" input box to suggest address options as you type, similar to how we suggest search results in the "Search We Vote..." box.

DaleMcGrew:
Verified. Thank you @sarah-clements! 👍

PR #644, opened 2017-02-07
Autocomplete address

sarah-clements:
Addresses issue #574 using the Google Autocomplete API and fixes a bug in the Ballot component where the user gets stuck on the loading screen if user doesn't enter an address but presses 'save' or 'enter'.


DaleMcGrew:
Looks great @sarah-clements! 👍

PR #646, opened 2017-02-09
This commit includes work-in-progress code, especially in OfficeItemCompressed, so I can get help with some styling. Please look in config-template.js for a new setting. Adding the ability to see voter guides to follow next to candidates on the ballot page.

DaleMcGrew:


DaleMcGrew:
👍

PR #666, opened 2017-02-18
Fixing facebook loging popup blocker and Share option for ballot link on facebook

jainanisha90:


DaleMcGrew:
👍 Thank you Anisha!

PR #673, opened 2017-02-21
Compact position layout

### LINKED ISSUE #654: Cleanup new 'summary' view styling on /ballot
Clean up styling of summary position view on /ballot (on develop branch). The reference design is here: https://app.moqups.com/DaleMcGrew/2vW41QMVXE/view/page/a211a4810

1) For candidates, keep all of these elements on one line:
Name + "follow options" or "positions in your network" + support/oppose

2) For measures, I think the measure description should display above the line that has these two elements:
"follow options" or "positions in your network" + support/oppose

- [x] Use flexbox instead of tables for layout
- [x] break apart styles assigned by context
- [ ] alternate layout for measures: display description above network indicator and buttons
- [ ] ~~alternate style when position is selected?? (still displaying both options)~~

### Pending questions:
- When should text be truncated, and by how much?
    - image
    - At this time I don't think we need to truncate the measure text

- Should we introduce a new pattern for persistently displaying both (support/oppose) options even after a selection has been made?
    - image
    - At this time I don't think we need to alter the support/oppose display standard (we can discuss and create another Issue if needed)

ixley:
**This PR:**
- Cleans up layout for 'compressed' office and measure views
- Switches grid system over to the Bootstrap 4 version using flexbox. Even though this wasn't ultimately used for this layout update, it will provide some new layout options moving forward.

**To note:**
- spacing is still not ideal on small screens due to the button width of selected [support/oppose] position dropdowns
- Measure descriptions are not set up to span the full width on smaller screens. We can address this separately.
- For additional layout/design updates, better design specifications would be useful.

resolves #654

### DIFF
--- a/Gulpfile.js
+++ b/Gulpfile.js
@@ -13,7 +13,8 @@ const del = require("del");
 const server = require("./server");
 
 const config = {
-    bootstrapDir: './node_modules/bootstrap-sass'
+    bootstrapDir: './node_modules/bootstrap-sass',
+    bootstrap4Dir: './node_modules/bootstrap'
 };
 
 const PRODUCTION = process.env.NODE_ENV === "production";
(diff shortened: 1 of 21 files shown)


ixley:
@DaleMcGrew I was pretty careful when merging, but the diffs were not that smart (big blocks), so please double check to make sure any recent edits were not overwritten.

DaleMcGrew:
Thank you @ixley, I am excited to check it out! I'll pull it locally and do a careful compare.

DaleMcGrew:
Hi @ixley this looks great and there isn't any lost code in the merge. I think you are right that on small screens we will need to put the "Positions in your Network" bar and the support/oppose on their own line (below candidate name or measure desription). 👍

Issue #676, opened 2017-02-23
Make the positions popup a bit friendlier


nickelser:
### Please describe the issue (What happens? What do you expect?)
- [x] In mobile (iPhone Safari) the single click on a TinyOpinionToFollow is still opening the organizational popover, instead of opening the modal.
- [x] On https://wevote.us/ballot when looking at Kamala Harris, hover over "California National Organization for Women". In the popover you are not able to click the Follow button (the left-most 1/2 of it at least).
- [x] In desktop mode I noticed that multiple popovers can open:
<img width="463" alt="screen shot 2017-03-01 at 7 37 22 pm" src="https://cloud.githubusercontent.com/assets/7756031/23492033/eef33d2e-feb6-11e6-81a9-196e0404665b.png">

![screen shot 2017-02-22 at 7 17 31 pm](https://cloud.githubusercontent.com/assets/367553/23243417/bd8b8eea-f933-11e6-8f7b-6ce3cd28986d.png)

The positions popup is a little wonky in formatting (some overflow), has some minor usability issues (cannot hover over and click anything).

### Steps to reproduce the problem (1, 2, 3...), including links

Hover over the organization.

nickelser:
Looks like the multiple popovers bug is still present -- reopening.

DaleMcGrew:
Great work!

PR #677, opened 2017-02-23
Fix the hover problems with org popovers.

nickelser:

https://github.com/wevote/WebApp/issues/676

Some JS changes to the hover behavior (that if needed again, should probably be extracted out into a HOC), and a teensy-tiny CSS change to let it fit more content.


DaleMcGrew:
Thank you @nickelser! 👍

PR #678, opened 2017-02-23
ready to vote filter

emilyliza:
- "Ready to Vote" Filter added to Ballot page
- Offices and measures display support/oppose text and symbols to the right of Candidate names.


DaleMcGrew:
Thank you Emily! 👍

PR #679, opened 2017-02-23
Multiple Ballots for one Voter

sarah-clements:
Fixed the bug with the loading wheel and empty ballot being returned (updated voterAddressSave callback with a google_civic_election_id return for simple_save) and set up stylesheet for SelectBallot route. This is not final and needs testing; more style changes to come.


DaleMcGrew:
Thank you @sarah-clements! 👍

Issue #101, opened 2016-02-23
Multiple ballots for one voter: Allow single-click switching


DaleMcGrew:
Let voters save and switch between multiple ballots, whether they be from this or prior elections. We can discuss the interface when it is time to implement. @lisamcho has updated the back end to work with an array of elections already.

DaleMcGrew:
First version done, thank you @sarah-clements and @KIVA7777!

PR #693, opened 2017-02-28
Final interface design for multiple ballot selection

sarah-clements:
For issue #101 , the UI was changed to a modal instead of a separate route.


DaleMcGrew:
Looks great @sarah-clements! 👍

PR #694, opened 2017-03-01
Clean up Ready To Vote pages

emilyliza:
- OfficeItemReadyToVote and MeasureItemReadyToVote pages uses flex containers.

*Note: the text for "Your network is undecided" in the OfficeItemReadyToVote page still uses a "pull-right" instead of flex. This needs to be updated but I ran out of time today.


DaleMcGrew:
Thank you Emily! 👍

Issue #707, opened 2017-03-03
Implement CSS for printable "Ready to Vote" page


DaleMcGrew:
### Please describe the issue (What happens? What do you expect?)
When I use my browser's "Print" function, the styles have not been set up to make the result look good in PDF format. This Issue is about creating print-specific styles that make the PDF output look good for printing your ballot to take to your polling place.
![screen shot 2017-03-03 at 9 40 27 am](https://cloud.githubusercontent.com/assets/7756031/23562005/88034be6-fff5-11e6-9c70-22edb87eeee4.png)
- [ ] Remove the We Vote top navigation
- [ ] Make the election name slightly larger
- [ ] Remove the settings gear
- [ ] Remove the "edit" link for the address
- [ ] Remove the "All items", "Remaining" and "Ready to Vote" buttons
- [ ] Make the office name smaller
- [ ] do not show the link path (ex/ "/office/wv02off1016")

### Steps to reproduce the problem (1, 2, 3...), including links
1. Go to the Ready to Vote page in desktop mode: https://wevote.us/ballot?type=filterReadyToVote
2. In your browser, print

DaleMcGrew:
Great work @ixley!

Issue #704, opened 2017-03-03
Ballot items list photos: visual clean up


DaleMcGrew:
### Please describe the issue (What happens? What do you expect?)
On the main ballot page, the small (24 x 24 in mobile mode) and larger (48x48 in desktop mode) images need better spacing.

### Steps to reproduce the problem (1, 2, 3...), including links
1. Go to http://localhost:3000/ballot (not on live server yet)
2. Enter address "Oakland, CA"
3. Notice desktop spacing around candidate photos AND mobile spacing

ixley:
@DaleMcGrew I'm getting a little confused about what the intended display is supposed to look like. I've added spacers between the candidate avatars and their names to the right, but a lot of the code in here feels like it's adding unnecessary complexity without a clear reason (that I can tell at least). There are also more custom classes that have been added that again, seem like they are adding complexity without a clear justification.

For example, why are there redundant elements for the candidate name and photo that get hidden and shown at different breakpoints? Am I missing something?

DaleMcGrew:
Hi Jeff, feel free to clean up the code and CSS. The goal is to have  different sized photos in mobile versus desktop, and keep the photo, Position bar, and Support or Oppose buttons on the same line if possible, but wrap if not. Thank you!

ixley:
I did some light cleanup but didn't want to get into the weeds since a lot of this feels somewhat fragile. What we really want to target layout changes like this is element queries (instead of media queries). They are not here for CSS yet, but there is some promise with React plugins. Anyway, that's a whole other endeavor but worth looking into eventually.

DaleMcGrew:
Awesome work! Thank you!

PR #714, opened 2017-03-05
Add basic print formatting (for ballot) and light cleanup for Ballot Item 

ixley:
- Addresses #707 - Adds basic print formatting cleanup for `/ballot`. 
  - Removing high level app/layout sections can be targeted in `_print.scss`. Smaller individual elements can be hidden for print with `.hidden-print`
- Addresses #704 - I tried to find a good balance of doing some cleanup but not getting too deep in the weeds. 
- Also resolved a few random linter warnings


DaleMcGrew:
Thank you @ixley, I really appreciate your layout and style work! 👍

Issue #706, opened 2017-03-03
Submitting empty search shows error in browser console


DaleMcGrew:
### Please describe the issue (What happens? What do you expect?)
When you submit an empty search from the top navigation bar, there is an error in the browser console:
![screen shot 2017-03-03 at 8 35 43 am](https://cloud.githubusercontent.com/assets/7756031/23559658/8e8fe914-ffec-11e6-87e5-d0bb471a87f0.png)

### Steps to reproduce the problem (1, 2, 3...), including links
1. Go to https://WeVote.US/ballot
2. Click in the search bar
3. Press enter on the keyboard, or click on the magnifying glass
4. The above error appears in the browser console log.

DaleMcGrew:
Fixed by @bigethan -- thank you!

PR #716, opened 2017-03-06
Empty search submitted will no longer error

bigethan:
Hi, I'm new to this project, so I'm very open to suggestions / advice. Contributing to app projects can be weird without a greater product context, but I wanted to give it a try.

Addresses issue #706

1. Added a method `searchHasContent` to `SearchAllBox.js` and used that method in `onSearchFormSubmit` - this prevents the search from trying to process without any content.

2. Removed the `onFocus` method from the search button.  I moved the logic for that to `onSearchFormSubmit` and only had it trigger if the search field wasn't already open. It's a little more predictable now to use the search icon as a search toggle.


DaleMcGrew:
Thank you @bigethan, I think your solution is clean. I tested it locally and it works very well. Welcome to the team, and it was great to meet you yesterday! 👍

PR #718, opened 2017-03-07
URLs play videos on snippets pages

emilyliza:
- Added React player to `PositionInformationOnlyStatement.jsx` and `PositionSupportOpposeSnippet.jsx`
- Minor clean up

NOTE: The code looks for URLS in both the `more_info_url` and `statement_text`. Is there reason for a user to add a URL in both places for the same measure/offfice? If so, this code should be probably be modified to account for that.


DaleMcGrew:
Thank you @emilyliza! 👍

PR #721, opened 2017-03-08
clearing all lint errors

bigethan:
None, just cleanup.  Should there always be an issue?
Removing some unused variables, declaring some globals as global, and adding some mediocre prop types.


DaleMcGrew:
An issue isn't required for a pull request -- we aren't big on procedure for procedure's sake. Thank you for doing eslint cleanup work!

PR #722, opened 2017-03-09
removing unused underscore dependency

bigethan:
Removed underscore from `package.json` as it's not used in the project.


DaleMcGrew:
Thank you! 👍

PR #728, opened 2017-03-15
Stripe client library setup

sarah-clements:
Setup the stripe library in a component that provides a donation form and returns a token to be used server-side for processing charges.


DaleMcGrew:
👍

PR #733, opened 2017-03-16
"Add another invitation" button is removed after 5 invite inputs

emilyliza:
- "Add another invitation" button is hidden after 5 input fields are displayed.
 - This button is no longer disabled after 5 input fields.


DaleMcGrew:
👍

PR #754, opened 2017-03-23
Skip around the popover problem on mobile by not rendering it.

nickelser:

Fixes https://github.com/wevote/WebApp/issues/676.

Also, fix links being underlined even with the `u-no-underline` class applied due to an CSS import ordering issue.


DaleMcGrew:
👍

PR #767, opened 2017-03-30
Keep track of popover state per-org, to ensure the popovers actually close.

nickelser:
Last item remaining in https://github.com/wevote/WebApp/issues/676.

Instead of having a global popover state, which could be cleared if you quickly transitioned in some ways between different popovers, keep state per-popover (including the important close timer).


DaleMcGrew:
This looks great @nickelser, thank you! 👍

PR #765, opened 2017-03-30
Index work

Anoninnyc:
### What github.com/wevote/WebApp/issues does this fix?
#746
### Changes included this pull request?
Defer loading of css links

PR #772, opened 2017-04-02
Prioritize loading styles for loading screen

ixley:
### Issue:
Deferring all stylesheets in #765 caused some wonky text flashing:
we-vote-loading-without-styling

### Changes
This PR lets index.html load a dedicated stylesheet for loading screen styles, along with redundant inline styling to avoid the flashes of unstyled text.

BTW - why is `bundle.js` so huge?! Is there a way this can be broken up to get the essentials for the app loaded first and the let everything else load progressively?

FYI: @Anoninnyc

### DIFF
--- a/Gulpfile.js
+++ b/Gulpfile.js
@@ -89,7 +89,7 @@ gulp.task('compile-bootstrap', function() {
 
 // Compile main stylesheet and copy to Build directory
 gulp.task("sass", function () {
-  return gulp.src("./src/sass/main.scss")
+  return gulp.src("./src/sass/{main,loading-screen}.scss")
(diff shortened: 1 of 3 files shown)


DaleMcGrew:
Thank you Jeff! 
I agree that it would be ideal for the bundle to be smaller. To clarify, the production (minimized) version is 2.6 Mb, while the development version we see on our development machines is 9 Mb. Were you calling 2.6 Mb huge?

ixley:
@DaleMcGrew Yeah, I was watching the 9 Mb dev version, but 2.6 on production also seems pretty high for a single blocking file. Just something to be mindful of, I guess.

PR #785, opened 2017-04-12
Stripe donation - error handling

sarah-clements:
Created a DonationError component and set up DonateStore listener in Donate.jsx to display error messages when a donation fails; error message will be dynamically created in the server based on specific type of exception (card failure, internal failure, etc)


DaleMcGrew:
Thank you Sarah! 👍

PR #789, opened 2017-04-20
Revert some behavior around popovers, and improve tap target.

nickelser:
Addresses issue #676.

- Fixes the mobile behavior to open the popover again.
- Makes the tap target for the org tiny images a little bigger (we will need to use slightly larger images).


DaleMcGrew:
Thank you Nick! 👍

PR #796, opened 2017-04-21
Donation interface change

sarah-clements:
Changed monthly donation check-box to radio buttons with a default of "monthly".


DaleMcGrew:
👍

Issue #797, opened 2017-04-22
Change Support/Oppose "at" state in ItemActionBar


DaleMcGrew:
In /js/components/Widgets/ItemActionBar.jsx, after a voter has selected "Support" or "Oppose", we show just the selected position:
<img width="123" alt="screen shot 2017-04-22 at 9 58 26 am" src="https://cloud.githubusercontent.com/assets/7756031/25306461/5b22ff30-2742-11e7-9c3b-d515900cd01f.png">

- [ ] Instead, leave both Support/Oppose options visible, but with the selected state highlighted.
- [ ] New functionality: If Support is already chosen, and you click it again, remove Support. (Same for Oppose.)

Anoninnyc:
Hi @DaleMcGrew, so I think I have a working solution to this which  I'm going to refine before I submit PR, let me know if [this video](https://youtu.be/utvEZsh_qPE) demonstrates desired behavior,  (obviously not desired highlighting, I'm sure).

I also wanted to clarify some things:
1. You want the PositionDropdown component removed, correct?
2. If you have a preference, how would you like me to properly highlight the thumbs?

Anoninnyc:
also... you'll notice that my computer is very slow.

DaleMcGrew:
Thank you Chris, for now you can just make the support thumb icon take the green, and the oppose take the red.

DaleMcGrew:
And please leave the PositionDropdown component in the repo for now.

Anoninnyc:
@DaleMcGrew, - got it, thanks

ixley:
I appreciate the simplicity of the updated component but I'm unclear on the intent here. My original understanding was that once a user selects a position, we aren't trying to encourage them frequently change their position. This updated approach seems to remove friction for changing your position, but I'm unclear on the value it provides in doing so. @DaleMcGrew could you clarify this for me?

If the goal, however, is just to make the selected state more compact, we can do that by removing the label text but without keeping both options exposed, no?

DaleMcGrew:
Hi Jeff, thank you for your question. Cara, Samantha and Randy all argued for the benefits of simplifying the support/oppose selection process on the ballot page. This implementation needs more visual design, but until we finish that design, this pull request gives us a chance to see how it feels.

DaleMcGrew:
[...]

PR #799, opened 2017-04-23
Item action bar

Anoninnyc:
#797
Change position/stance toggling to different UX.

### DIFF
--- a/src/js/components/Widgets/ItemActionBar.jsx
+++ b/src/js/components/Widgets/ItemActionBar.jsx
@@ -28,19 +28,24 @@ export default class ItemActionBar extends Component {
     this.setState({transitioning: false});
   }
 
-  supportItem () {
+  supportItem (is_support) {
+    console.log("is_support", is_support);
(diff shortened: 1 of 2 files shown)


Anoninnyc:
Hello:
[This](https://youtu.be/Ybl6Neu8wIA) is the use case I considered for when making the changes you see.

1.  I removed the  PositionDropdown component, along with all the variables it consumed, from the .jsx file you see changed, thought I didn't remove the component from the repo... as per @DaleMcGrew.
2. Aside from that, I'm adding/removing classes from the thumbs' buttons according to whether or not the user supports/ opposes them. I aliased a lengthy className, but all else is the same. I kept the commentButtonHide and shareButtonHide functionalities in tact.
3. For SASS, I nested the classNames that trigger highlighting in a selector that... "... is for when the item-actionbar is added to a line with other elements (like for a candidate on ballot page)", as per original author.

DaleMcGrew:
Thank you @Anoninnyc! 👍

Issue #787, opened 2017-04-19
Replace react-headroom React package because it makes the bundle too big


DaleMcGrew:
react-headroom (14.8M and 7 children) is the biggest package. It's only required on Application.jsx to hide the header when the user scrolls. I think we can switch to a lighter package, such as the original headroom.js (not specific to React, but easy to readapt), or some other similar;

More notes here:
https://docs.google.com/spreadsheets/d/1mINBClQJ8CBd8e4rceE8ZvXPJXZE4d6dxa9Tp5OcCo0/edit#gid=0

DaleMcGrew:
(I just assigned it to you Eric because I saw your note on the other case where you said you could look at other options for react-headroom. Thank you!)

clapinton:
I'm testing the original headroom.js package. The folder is 70KB. It needs a bit more work on it, as it only assigns specific class names to a target element, so it's up to us to specify the CSS rules that hides/shows the header bar. I'm working on implementing the same behavior as we have now.

DaleMcGrew:
Great work Eric! Fixed.

PR #803, opened 2017-04-26
Implementing headroom.js in lieu of react-headroom

clapinton:
Fix for issue #787 

I removed the react-headroom package from package.json and added [headroom.js](https://www.npmjs.com/package/headroom.js). The new package only assigns classes to an element of your choosing in an up or down scroll event, with offset and tolerance configs. Therefore, we must write our own hide/show CSS rules.

This prompted a couple of changes to the Application.jsx file. Basically, when the component updates, it configures a new headroom object, if one does not exist already. This happens inside `componentDidUpdate`, since we need to stylize an HTML element which is not yet defined on the DOM when `componentDidMount` gets called.

Also, added a couple of CSS rules to account for the hide/show animation. I've used the same `transform: translate` property as was used before, which results in the header to jump momentarily when scrolling up. This can be fixed by changing `transform: translate` to the `top` property, but that ends up taking a toll on CPU performance.


DaleMcGrew:
Thank you Eric! I tested it, and everything is working very well. Thank you for tackling this important optimization! 👍

Issue #804, opened 2017-04-27
Headroom miscalculates header height (due to dropdowns)


ixley:
### The issue
It appears that the replaced Headroom.js (via #787) introduces issues with detecting the header height. Loading or refreshing the page locally (not currently on production) commonly miscalculates the header height and assigns an inappropriately large value, creating a big spacing gap. This appears to be related to including the height of the dropdown menus in its calculation (perhaps having to do with the dropdown menus being rendered in the layout before they're hidden?)

![image](https://cloud.githubusercontent.com/assets/160662/25462764/134fb448-2aa6-11e7-9c53-e4f4a9aad1fb.png)

### Steps to reproduce
Do a force refresh of the page (locally) until you notice the spacing bug. (Normal page refreshes don't seem to trigger this bug.)

_I'm not sure which is the bigger culprit - how headroom.js is configured or how the dropdown menus are set up and positioned, but since this bug seems to have been introduced by switching from react-headroom to headroom.js, I'll assume the former is more directly relevant to look at for this._

FYI @clapinton

ixley:
looks good on my end.

PR #805, opened 2017-04-27
Fix the top margin issue with headroom (and do some cleanup).

nickelser:
https://github.com/wevote/WebApp/issues/804

The changes were actually CSS only as far as I can tell, but I also went and cleaned up the headroom stuff.


DaleMcGrew:
👍

PR #813, opened 2017-04-30
fix up position card layout/wrapping

### LINKED ISSUE #784: Update ballot to use inline styling to avoid position bar overflow
In this screenshot we show that position bars wrap to the next line in cases where they could fit on the same line as the candidate name.
screen shot 2017-04-11 at 1 40 03 pm

ixley:
Addresses layout wrapping for wrapping on the position-item card (Fixes #784)

@DaleMcGrew BTW - try to use `span`s only to wrap text or otherwise inline positioned content. For block layout elements that need a wrapper, use `div`s (as a general rule).


DaleMcGrew:
Thank you @ixley! 👍

PR #814, opened 2017-04-30
Fixed issue with support/oppose getting pushed to the right of the ballot column on main ballot page. Fixed up "learn more" and added to MeasureItemCompressed.

DaleMcGrew:
Fixed issue with support/oppose getting pushed to the right of the ballot column on main ballot page. Fixed up "learn more" and added to MeasureItemCompressed.

### DIFF
--- a/src/sass/components/_card.scss
+++ b/src/sass/components/_card.scss
@@ -161,10 +161,22 @@
     }
 
     &__candidate-read-more-link {
-      left: 0;
       margin-left: 10px;
       top: 0;
-      font-weight: lighter;
+      font-family: $body-font-stack;
+      font-size: 14px;
+      font-weight: 100 !important;
+      color: $link-color;
+      width: 100%;
+      height: 100%;
+      overflow: hidden;
+    }
+
+    &__measure-read-more-link {
+      margin-left: 10px;
+      font-family: $body-font-stack;
+      font-size: 14px;
+      font-weight: 100 !important;
       color: $link-color;
       width: 100%;
       height: 100%;
@@ -172,7 +184,8 @@
     }
 
     &__office-read-more-link {
-      margin-left: -13px;
+      margin-left: 10px;
+      font-family: $body-font-stack;
       font-size: 14px;
       font-weight: 100 !important;
       color: $link-color;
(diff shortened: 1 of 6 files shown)


src/sass/components/_card.scss
```diff
@@ -161,18 +161,31 @@
     }
 
     &__candidate-read-more-link {
-      left: 0;
       margin-left: 10px;
       top: 0;
-      font-weight: lighter;
+      font-family: $body-font-stack;
+      font-size: 14px;
+      font-weight: 100 !important;
```
ixley: @DaleMcGrew be mindful of this extremely light weight. I would actually recommend against using such a light weight for small text like this. There is the even possibility you are not even seeing the actual weight as it doesn't render the 100 weight for me on desktop – I only noticed it on my iPhone.

src/sass/components/_card.scss
```diff
@@ -161,18 +161,31 @@
...
+      color: $link-color;
```
ixley: @DaleMcGrew side comment: I'm wondering what an appropriate default way to style link text on the site is since I'm noticing js handlers seem to be used to capture most click events... The simplest would be to still wrap links in `<a>` tags. Does that seem reasonable?

src/sass/components/_card.scss
DaleMcGrew: Hi @ixley can we discuss when we are looking at screens together? I would like these "learn more" links to be extremely light to minimize visual clutter.

src/sass/components/_card.scss
DaleMcGrew: @ixley it seems fine to me to wrap the links we want to show an underline on hover to use the a tag.

src/sass/components/_card.scss
ixley: yes, we are agreed on the goal of wanting to minimize clutter. I know you have said some people seem to not know they can click on things for more context, but IMO 'read/learn more' links should be used much more sparingly and not peppered all over. Part of the clutter is unavoidable when there are this many of them and changing the text styling does help differentiate them, but does not necessarily reduce clutter. We can discuss other ways to better address this later...

PR #815, opened 2017-04-30
New style linter rule (border-radius)

ixley:
Introduces style linter to require variables for border-radius values. (Additional properties to be added incrementally.)


DaleMcGrew:
👍

PR #818, opened 2017-05-02
Add linter to require variables for color properties

ixley:
Adds another stylelint rule to require variables for color properties. 

All the rules to require variables for style properties are intended to reduce the introduction of custom values and to work from a set of established values with scoped names.

Variables should typically be added to the _variables.scss partial. If they happen to be specific to a specific component or element, they can be scoped to that component (ideally nested within the most encompassing context), but this sort of customization should still be done sparingly.


DaleMcGrew:
👍  Thank you @ixley for this very important maintenance and upgrade work!

PR #819, opened 2017-05-02
small adjustment to fix #812

### LINKED ISSUE #812: In Modal popup for organization on Ballot, "Follow" button cannot be clicked
On the ballot page (Address: Oakland, CA 94602), prior to signing in, there are a number of organizational icons. When you hover over them, some of them have a problem with the Follow button. The Follow button on some of them cannot be clicked along the left side of the button. The clickable area is obscured for some organizations.

### Steps to reproduce the problem (1, 2, 3...), including links
1. Sign out
2. Go to ballot page: https://wevote.us/ballot
3. Set your address to "Oakland, CA 94602"
4. Find one of these two organizations. There is something about their names that causes the clickable area of the Follow button on the left side to be "obscured":
screen shot 2017-04-29 at 8 01 10 pm
screen shot 2017-04-29 at 8 01 59 pm
5. I think this code is in "organizationPopover" within WebApp/src/js/components/VoterGuide/ItemTinyOpinionsToFollow.jsx

ixley:
Removes unnecessary padding elements for an indirect fix. Not worth spending more time on without reworking the actual component...
fixes #812

### DIFF
--- a/src/js/components/VoterGuide/ItemTinyOpinionsToFollow.jsx
+++ b/src/js/components/VoterGuide/ItemTinyOpinionsToFollow.jsx
@@ -123,12 +123,8 @@ export default class ItemTinyOpinionsToFollow extends Component {
             onMouseOver={() => this.onTriggerEnter(org_id)}
             onMouseOut={() => this.onTriggerLeave(org_id)}
             className="card-popover">
-            <div className="card">
-              <div className="card-main">
-                <FollowToggle we_vote_id={one_organization.organization_we_vote_id} />
-                <OrganizationCard organization={one_organization_for_organization_card} />
-              </div>
-            </div>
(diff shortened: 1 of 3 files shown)


src/js/components/VoterGuide/ItemTinyOpinionsToFollow.jsx
```diff
@@ -123,12 +123,8 @@ export default class ItemTinyOpinionsToFollow extends Component {
             onMouseOver={() => this.onTriggerEnter(org_id)}
             onMouseOut={() => this.onTriggerLeave(org_id)}
             className="card-popover">
-            <div className="card">
-              <div className="card-main">
```
ixley: this is just an incidental fix. By removing some padding, the popover has a bit more space so the text container is less likely to overlap part of the Follow button.

DaleMcGrew:
👍

PR #824, opened 2017-05-08
New DonationList component, some refactors.

SailingSteve:
Donation directory.
Removed list generation from DonationForm, refactored into new
DonationList component.
Created new DonationList component, that provides the list of past
donations and current subscriptions.  Current subscriptions tab is
currently disabled.  Styles are currently hard coded, will improve this
week.
ProcessingDonation was throwing warnings... it was trying to change
state in the render(), so I refactored.


src/js/routes/More/ProcessingDonation.jsx
```diff
@@ -23,28 +23,24 @@ export default class ProcessingDonation extends Component {
...
   }
 
   _onDonateStoreChange () {
-     this.setState({donation_response_received: DonateStore.donation_response_received(), donation_success: DonateStore.donation_success()});
-    }
-
-
-  render () {
+    this.setState({ donation_response_received: DonateStore.donation_response_received(),
+      donation_success: DonateStore.donation_success(), });
```
DaleMcGrew: Hi Steve, when you call this.setState, the value isn't accessible until a re-render. Moving "if (this.state.donation_response_received)" into _onDonateStoreChange will not behave the way it looks like you are seeking. It is more common to put the state-dependent reactions in the "render".

DaleMcGrew:
I am approving, but we will want to discuss before I push the code to the live server. Thank you Steve!

PR #826, opened 2017-05-11
Changess to DonationList and DonationListForm to handle the display of subscription donation history.

SailingSteve:
DonationList component, Dale was ok with inline react-bootstrap style
settings, so I removed some className and some bsStyle props.  Now shows
all of the subscriptions, (which until I make the sync feature, all show
as active).
DonationListForm now is suppressed if there is no data to show.
ProcessingDonation: Improvement on React lifecycle usage.


DaleMcGrew:
👍

PR #837, opened 2017-05-15
remove leftover code from ‘React Burger Menu’

### LINKED ISSUE #816: Remove React Burger Menu
Is Burger Menu still being used for anything or can we remove it?
cc @DaleMcGrew

ixley:
basic cleanup removing leftover code from the React Burger Menu plugin (which looks like it was removed previously).

I'm not sure what a good way to codify the process would be, but when we remove plugins or custom code, we should also be making sure to remove and/or clean up related code as basic good housekeeping. Consistent naming and commenting and isolating related files also goes a long way towards supporting this.

Resolves #816


src/js/components/Navigation/HeaderBar.jsx
```diff
@@ -297,71 +297,69 @@ export default class HeaderBar extends Component {
 
     return (
       <header className="page-header">
-        <div className="page-header__content">
```
ixley: This diff looks horrible but the only change is that the `page-header__content` div was removed and indentation was updated for nested content.

DaleMcGrew:
Thank you Jeff for doing this cleanup work! 👍

PR #846, opened 2017-05-17
Made the donation history window scrollable if the list gets long.

SailingSteve:
(Made it reverse time ordered on the server side).


DaleMcGrew:
Looks great, thank you! 👍

PR #848, opened 2017-05-18
Added a div with fixed height with two columns and vertical scroll bar for facebook friends list

jainanisha90:


DaleMcGrew:
👍 Thank you Anisha!

PR #853, opened 2017-05-22
Cleared some JavaScript console warnings

SailingSteve:
Use css dimension with units, and clear a "whitespace console warning"  in DonationList.
In DonationListForm, make sure the data is there, before trying to determine its length.


DaleMcGrew:
👍

PR #867, opened 2017-05-26
Changes to get the most current donation to show up in history list

SailingSteve:
In VoterActions, add voterRefreshDonations
In DonationList, add listeners for changes to VoterStore, and on that
change make a call to VoterActions.voterRefreshDonations();
In DonationListForm same types of changes as in DonationList
In Ballot, add a logline to help diagnose setState issue and distinguish
this one from the other similar warnings
In Donate forced the float donation amount to int to avoid a roundoff
error.
In DonateThankYou added the DonationListForm component to display the
donation you just made.
In ProcessingDonation moved the browserPush out of the render, to avoid
an anti-pattern warning.
In VoterStore added a place to store the voter_donation_history_list
In Application.jsx, added /more/donate_thank_you" to the list of paths
so that the components render in the center of the page.


DaleMcGrew:
👍

PR #876, opened 2017-06-01
Changes to allow the cancellation of subscriptions, and support for Webhooks from Stripe.

SailingSteve:
In DonateActions support for the new endpoints for CancelSubscription and Refunds.
New class DonationCancelSubscription.jsx
In DonationList, show "waiting" for subscriptions that have been
created but have not received their card info from the Webhook. Don't
show the info button on one-time donations if the donation is more than
30 days old.  30 days is \set on the server side in an environment
variable "STRIPE_REFUND_DAYS".
In DonationList -- the Webhook that adds the card info to the
subscription (that is in the DB, but is incomplete) takes 5 to 30
seconds. So we can't just block for all that time, so everytime you
enter the subscriptions tab, we do a
oterActions.voterRefreshDonations() to update the data. Simple, but
should be good enough.
In DonateStore add support for the new endpoints.

### DIFF
--- a/src/js/actions/DonateActions.jsx
+++ b/src/js/actions/DonateActions.jsx
@@ -2,8 +2,17 @@ import Dispatcher from "../dispatcher/Dispatcher";
 
 module.exports = {
 
+  donationCancelSubscriptionAction: function (subscription_id) {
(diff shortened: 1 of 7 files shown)


DaleMcGrew:
👍

PR #882, opened 2017-06-05
Icon to SVG for person avatar and organization

bharathdn:
Icon to SVG for person avatar and organization


src/js/components/ImageHandler.jsx
```diff
@@ -25,26 +25,29 @@ export default class ImageHandler extends Component {
...
+      place_holder_image_url = "/img/global/svg-icons/organization-icon.svg";
     }
 
     if (show_placeholder_if_image_missing) {
       // This branch is for situations where we want to show a placeholder image in case the image is broken or missing
-      const image_or_placeholder = this.state.error || this.props.imageUrl === "" ?
-        <i className={sizeClassName + replacementClass}/> :
-        <img className={sizeClassName + this_class} src={this.props.imageUrl} alt={alt}
+      const image_or_placeholder = this.state.error || !this.props.imageUrl || this.props.imageUrl === "" ?
+        <img className={sizeClassName + " " + replacementClass} src={place_holder_image_url}/> :
```
pertrai1: You can use es6 template strings for these concatenations:

```
{`${sizeClassName} ${replacementClass}`}
```

src/sass/base/_images.scss
```diff
@@ -0,0 +1,11 @@
+.image-person-placeholder {
```
pertrai1: I would combine these together so you only have to use the property once:

```
.image {
    &-person, &-organization {
        &-placeholder {
            vertical-align: middle;
        }
    }
}
```
I think that is how you would do the Sass. Have not worked with Sass in awhile but the point would be to have both classes share the property and value in one rule

DaleMcGrew:
👍

src/sass/base/_images.scss
bharathdn: Hi Rob, 
Thanks for suggesting these changes, 
I'm new to sass as well. I have made the changes and will check them in my next commit.

Issue #872, opened 2017-05-30
Add share buttons to Welcome page (Facebook, Twitter, Email)


DaleMcGrew:
- [x] On the new welcome page, create a share default popup for Facebook & Twitter (provided by Facebook and Twitter) for sharing We Vote.
- [x] On the new welcome page, create a "mailto" for the Email share button.
Suggested text: "Check out https://WeVote.US! View your ballot. Learn from friends. Share your vision. @WeVote #Voting #WeVote"

ixley:
* I've commented out the buttons until someone makes them functional

DaleMcGrew:
Hi @jainanisha90, on this page when I click the "share by email" link, it opens the mailto in another browser tab using an "a" tag:
https://www.aclu.org/bio/faiz-shakir

<a href="mailto:?body=https%3A%2F%2Fwww.aclu.org%2Fbio%2Ffaiz-shakir%0D%0A%0D%0AFaiz Shakir joined the American Civil Liberties Union (ACLU) in January 2017. As National Political Director, he oversees the ACLU’s National Political Advocacy Department, which houses the organization’s Washington Legislative Office and State Advocacy and Policy departments. In his role, Shakir develops and implements strategies to advance the organization’s priorities at the federal and state levels.&amp;subject=Faiz%20Shakir%20%7C%20American%20Civil%20Liberties%20Union" target="_blank" title="Submit this to Email">Email</a>

That is what I suggest we try.

DaleMcGrew:
Verified.

PR #883, opened 2017-06-05
Added Facebook/twitter sharing option on welcome page

jainanisha90:

### DIFF
--- a/src/js/routes/Welcome.jsx
+++ b/src/js/routes/Welcome.jsx
@@ -6,6 +6,7 @@ import VoterActions from "../actions/VoterActions";
 import VoterConstants from "../constants/VoterConstants";
 import VoterStore from "../stores/VoterStore";
 
+const web_app_config = require("../config");
 
 export default class Intro extends Component {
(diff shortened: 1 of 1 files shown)


src/js/routes/Welcome.jsx
```diff
@@ -60,6 +61,32 @@ export default class Intro extends Component {
     this.setState({loading: true});
   }
 
+  shareToFacebookButton () {
```
pertrai1: These 2 buttons could be extracted out into social buttons component that could then make them more reusable around the app. Then, at the same time, if more social buttons are added, you won't have to go into files to add or update the buttons that are already in any files. Just a thought.

pertrai1:
This PR goes along with #872

DaleMcGrew:
👍

src/js/routes/Welcome.jsx
jainanisha90: Hi Rob,
Thank you for suggestion.
I will try to make a separate social button component so that it will become more reusable.

PR #884, opened 2017-06-06
Updates for DonationList updating on change to store

SailingSteve:
In DonationCancelOrRefund, changed some on screen text and fixes.
In DonationList, now listening on DonateStore for updates to refresh
the list of donations/subscriptions after a server side change.
In VoterStore show the cookie value for voter_device_id in an error
condition where the device_id is not in the db.


SailingSteve:
I don't know how these  changes got lost.  Yesterday was a rough day with git.  Got some help from Dale.  The app is broken without these.

DaleMcGrew:
Hi @SailingSteve, you will need to create a new pull request with the additional changes since yesterday, because we had already merged your changes yesterday. Once the pull request is merged, additional changes to it don't get included in develop. *Before* it is merged you can add additional code to a pull request.

PR #886, opened 2017-06-07
Cleaned out some un-needed logging for donations, and some un-needed refreshes of the donation list from the server.

SailingSteve:
In DonationList, cleaned out some logging.  Removed some un-needed
refresh of the donations list from the server.
In DonationListForm, removed an unneeded donation list refresh.


DaleMcGrew:
👍

Issue #869, opened 2017-05-26
Add overlay with the names of all supporters (or opposers)


DaleMcGrew:
When you are on the candidate page and you have a position bar with supporters or opposers, we want to add an overlay when you hover over the “+9” (or click in mobile mode), showing who else supports the person. So this overlay should show *all* supporters (or opposers).
<img width="248" alt="screen shot 2017-05-26 at 4 29 57 pm" src="https://cloud.githubusercontent.com/assets/7756031/26515614/286acfbe-4231-11e7-98bb-23988187a81e.png">

DaleMcGrew:
Hi Anisha, I updated the live site and am not seeing this popover. Can you double-check? https://wevote.us/candidate/wv02cand3679
Just a note that the goal is to show *all* of the supporters (or opposers), including the ones that have an icon on the candidate page.

jainanisha90:
Hi Dale, 
I misunderstood #869 and added overlay on ballot page instead of candidate page. Sorry about that.
Created a new PR for overlay at candidate page. For now click on +9 is not working.
I wanted to confirm like What do we wants to display on click of "+9"? Do we want to show all positions for that candidate?

DaleMcGrew:
Verified. 👍

PR #892, opened 2017-06-08
#869 Added overlay with the names of all supporters or op-posers at candidate page

jainanisha90:
#869


DaleMcGrew:
👍

PR #897, opened 2017-06-17
Do not silently fail when the voter address does not find a close by ballot.

SailingSteve:
The server had already been sending back ballotCaveat ... the reason
why the address was not accepted from /settings/location.  Now we store
ballotCaveat in the BallotStore, and listen for changes to it in the
AddressBox, and display the reason why we keep displaying the address
box.  Previously we were silently failing.


DaleMcGrew:
👍

PR #898, opened 2017-06-17
Voter tab guide iphone alignment

### LINKED ISSUE #881: Voter guide "tab" width adjustment in mobile
When viewing an organization's voter guide in mobile mode (iPhone 5 or smaller):
https://wevote.us/ACLU
...the tabs wrap to a second vertical line:
screen shot 2017-06-05 at 8 57 53 am

Please adjust the html/css so that they don't wrap on very small screens. Code here:
WebApp/src/js/routes/Guide/OrganizationVoterGuideTabs.jsx and the style is "tabbed_voter_guide_details"

ixley:
Used a different fix for fitting those tabs on small screens.
fixes #881


ixley:
@dalemcgrew yes I did check but it's nowhere near fitting. If this matters to you I would suggest keeping the same short labels or considering a a very style that looks better when wrapping. I personally don't think it's a significant issue in this context.

Issue #863, opened 2017-05-25
Add list of ballot items to js/Components/Navigation/BallotLeft.jsx


DaleMcGrew:
To the left of the Ballot (in desktop mode) we want to show a list of all of the offices and measures (no candidates) that are on the ballot so that voters can jump quickly down to an office or measure. Please include an "anchor tag" when you click the office name or measure name.

The code powering this left component is js/Components/Navigation/BallotLeft.jsx

DaleMcGrew:
Thank you Steve for your first version. It looks like you are making the quick links go to another page (the Office or Measure pages). Our goal is to jump the voter down the Ballot page to the particular item (the same page the voter is already on), with an anchor tag.

SailingSteve:
TODO:

1. Figure out why we have to sometimes click twice to get the page to scroll to the anchor
2. Leave in the subtitle, but shorten it to fit on the same line as the title, and use css to make it fit
3. Make the fonts and white boxes match the mock Dale sent me
4. In the mobile layout (responsive html5 iphone5 as reference size) make the "Summary of Ballot Items" a modal pop-up, like SelectAddressModal in ballot.

DaleMcGrew:
The first version of this looks great @SailingSteve. Thank you! I look forward to seeing the mobile modal version and the styling updates.

DaleMcGrew:
Thank you Steve for your most recent upgrades to the BallotSideBar (I renamed the "BallotLeft" component). I made some style changes (make sure to grab before continuing work), and we can ask Jeff or another CSS expert for advice on getting the ellipsis working. So the only thing remaining is the mobile modal.

SailingSteve:
Hi Dale,
   My new Mac is transferring 1TB of data so I am computerless until Sunday.  I'm glad you renamed those files!  
   The ellipses were working, it was the responsive sizing that wasn't working due to that width value.  Did you try it first?  No worries in any case, I'll take a look Monday.

Thanks,
Steve

DaleMcGrew:
Great work @SailingSteve! I am marking this as finished. 👍

PR #913, opened 2017-06-30
More work for issue #863

SailingSteve:
In MeasureItemCompressed and OfficeItemCompressed added an anchor 
(the we_vote_id) so that the BallotLeft can link to the right.  Now 
follows the link with one click.
In BallotLeft, changed from a link to another page, to a href to the same page.  
In MenuLink did some preliminary shortening of the subtitle, the final will have to be in css.
In DonationList (unrelated to #863) removed some columns that Dale 
thought were un-necessary.  Added the word "canceled" in the columns
with the cancel subscription buttons.  Some early for for a 
"Last Charged" column, but I'm waiting for a response from Stripe re the best way to do that.


DaleMcGrew:
👍

PR #927, opened 2017-07-07
Styling changes for donation tables in mobile mode, for DonationForm,

SailingSteve:
DonationList, and Donate.  Only show 3 columns when mobile.


DaleMcGrew:
👍

PR #929, opened 2017-07-10
DonationList now displays last_charged for subscriptions.

SailingSteve:
Donate lint cleanup
_buttons.scss, minor style change for donation value buttons


DaleMcGrew:
👍

Issue #934, opened 2017-07-12
Pressing Enter in the custom amount text field, causes a full page refresh and loses context


SailingSteve:
The Donate.js money entry field, causes a page reload and looses context (any button choice, custom amount, etc.) if you press Enter with the cursor in the input field.

This is nasty because you are unsure if the donation you intended went through (even though you never saw the Stripe pop-up). 

To reproduce, select Donate (http://localhost:3000/more/donate) and then press the "Other Amount" button, type in a number, and press Enter.

SailingSteve:
Tested on live server.  Looks good.

PR #935, opened 2017-07-12
Fix for issue #934, page reload on Enter in money input for donation

SailingSteve:
Donate.jsx render() mixed original bootstrap with react-bootstrap, so
I redid it all in react-bootstrap to see if that solved the problem.
There were a number of Stack overflow about similar issues, and the
solution was to just swallow the Enter if in the input field.  The fix
is in _handleKeyPress in Donate.jsx.


DaleMcGrew:
Thank you for catching and fixing this @SailingSteve! 👍

Issue #868, opened 2017-05-26
Clicking friends/public toggle before sign in should show "Please sign in" Modal


DaleMcGrew:
When you are looking at a measure or candidate you will see a switch that allows you share your position publicly. We want to make this change: if you click the friends/public toggle before you are signed in, do not change the switch, but instead show a modal encouraging you to sign in. This modal should continue to appear every time you click the friends/public toggle until you sign in.
<img width="201" alt="screen shot 2017-05-26 at 4 26 29 pm" src="https://cloud.githubusercontent.com/assets/7756031/26515507/3ad7e098-4230-11e7-9a44-278f9399c531.png">

Add a modal in src/js/components/Widgets/PositionPublicToggle.jsx

Example code can be found in js/components/Widgets/ItemActionbar.jsx, starting with the function "supportItem".

Messages to use
Not signed in: Clicking this toggle will make your position visible to anyone on We Vote. In order to change this toggle you need to sign in first.

Signed in and first time: By clicking this toggle, you have just made your position visible to anyone on We Vote. If you do NOT want to share your position publicly, you may click the toggle again to restrict visibility to We Vote friends only.

DaleMcGrew:
Excellent work @edward-ly! Feature added.

PR #942, opened 2017-07-13
Add modal for public position toggle

edward-ly:
Added a modal for when the user attempts to change setting when not signed in

#868


DaleMcGrew:
This looks great @edward-ly, thank you for your first pull request with We Vote! 👍

Issue #861, opened 2017-05-25
Bookmark explanation modal


DaleMcGrew:
On the ballot page (in a component we can place everywhere a voter sees a bookmark), create a modal that explains what a bookmark is the first time you click it.

<img width="610" alt="13_friends_committee_on_national_legislation_-_we_vote" src="https://cloud.githubusercontent.com/assets/7756031/26434864/19376222-40c0-11e7-8158-c4f9ebdb5312.png">

This can be implemented using the "VoterStore.getInterfaceFlagState" system in the file js/components/Widgets/BookmarkAction.jsx

Example code can be found in js/components/Widgets/ItemActionbar.jsx, starting with the function "supportItem".

DaleMcGrew:
Thank you @edward-ly, this works very well.

PR #945, opened 2017-07-17
Added bookmark explanation modal

edward-ly:
Added a modal to explain bookmarking to the user. The exact text inside the modal is yet to be decided, but it will still change depending on whether or not the user is signed in.

#861


DaleMcGrew:
Hello @edward-ly this looks great. I found only two modifications:
1) we let people bookmark ballot items (candidates, offices, or measures) even when they are not signed in.
2) the modal can use the words "ballot item" instead of "candidate"
Once you have made these changes I'll approve the pull request and do more testing. Thank you!

DaleMcGrew:
Great work @edward-ly! 👍

PR #946, opened 2017-07-17
Refresh the HeaderBar's display of voter_photo_url_medium on the very

SailingSteve:
first login of a new voter, so that their picture shows up right away.
For:  Twitter photos not retrieved on first sign in #377


DaleMcGrew:
👍

Issue #864, opened 2017-05-25
Limit number of candidates (per office) to 3 on Ballot page


DaleMcGrew:
On the Ballot.jsx page, we only want to show the first three candidates at first because vertical screen real estate is so precious. If there are more than 3 candidates in a particular office, show a "Show More Candidates" link (in a horizontal bar that spans the width of the Office card) that "unfurls" all of the remaining candidates. This display is in WebApp/src/js/components/Ballot/OfficeItemCompressed.jsx

pertrai1:
@DaleMcGrew Is this something that will just have the rest of the candidates hidden or will only 3 be on the first fetch and then the rest on another request after clicking the "Show More Candidates"?

DaleMcGrew:
Hi @pertrai1, the way we retrieve the ballot/candidate data won't change. This issue only affects what is displayed in WebApp.

pertrai1:
@DaleMcGrew sounds easy enough. Let me know if there is nobody else that you have that can handle this for you.

pertrai1:
@DaleMcGrew we will need to look at this together because from what I am seeing is that there are numerous returns being made from OfficeItemCompressed.jsx that are not needed. Code review when you can with me

pertrai1:
@DaleMcGrew I added Discussion Needed label to keep track that I would like to talk about making smaller components for this.

DaleMcGrew:
This is what we would like the horizontal bar to look like and what it should say.
![ballotpage-clicktoseemorecandidates](https://user-images.githubusercontent.com/7756031/28330889-12a4858e-6ba4-11e7-8f99-69600f2feb07.png)

DaleMcGrew:
This is working with one issue entered as Issue #962. Thank you @bharathdn!

PR #948, opened 2017-07-19
Limit number of candidates (per office) to 3 on Ballot page

bharathdn:
Limit number of candidates (per office) to 3 on Ballot page

#864


DaleMcGrew:
👍

PR #961, opened 2017-07-21
Limit number of candidates to 3: Add url link to scroll to section header

bharathdn:
Limit number of candidates to 3: Add url link to scroll to section header

#864


DaleMcGrew:
👍

Issue #958, opened 2017-07-20
Refactor Ballot Modals into separate components


DaleMcGrew:
Move all of the Ballot Modals (WebApp/src/js/routes/Ballot/Ballot.jsx) and put them in their own components, in WebApp/src/js/components/Ballot, starting with the title "BallotModal..."

DaleMcGrew:
Awesome work @SailingSteve! This issue is fixed.

PR #964, opened 2017-07-23
Refactor Ballot Modals into separate components #958, first step

SailingSteve:
Pulled BallotIntroModal out of Ballot.jsx 
BallotIntroFriends had a wrong classname (downside of transpiling) 
BallotIntroPositionBar was importing a propType cloned from some other file, and did not exist.


DaleMcGrew:
👍

Issue #955, opened 2017-07-20
Address alert: "Not a full address" overlay


DaleMcGrew:
### Please describe the issue (What happens? What do you expect?)
We Vote has an IP address-based method of "guessing" about the voter's city, state, ZIP. We then return a ballot based on this information. This can present problems in big cities (like New York), and cause the wrong ballot items to appear for that voter. To solve this, we want to make it very clear when the voter has NOT entered their full address (when we are using a nearby ballot), and encourage the voter to enter their full address. See this design:
<img width="375" alt="screen shot 2017-07-20 at 10 05 01 am" src="https://user-images.githubusercontent.com/7756031/28430186-72ec1f92-6d34-11e7-9645-7d4138d62d15.png">

The code where the alert icon and "This is our best guess" overlay can be added is:
WebApp/src/js/routes/Ballot/Ballot.jsx (search for voter_address around line 556)

The variables that indicate that this isn't an address that we used to retrieve your ballot directly from Google Civic are highlighted below. I think we can start by checking to see if voter_address has "normalized_line1".
The "Enter Address >>" button can open the overlay with "_toggleSelectAddressModal"

**1) http://localhost:8001/apis/v1/voterAddressRetrieve**
{
normalized_city: "",
address_found: true,
google_civic_election_id: 5000,
status: "VOTER_ADDRESS_RETRIEVE-ADDRESS_FOUND",
longitude: "",
text_for_map_search: "Oakland, CA",
voter_device_id: "t0a95zIVLS0TOtVx4SXMhybU61ygwyVGc6EP73sY5Y3uVaNCCYbvyqNgAuaBS0LhFfgAOEnysiXdxW6QW4iXza16",
success: true,
guess_if_no_address_saved: true,
normalized_zip: "",
**normalized_line1**: "",
address_type: "B",
normalized_line2: "",
latitude: "",
normalized_state: ""
}

**2) http://localhost:8001/apis/v1/voterBallotItemsRetrieve/**
{
is_from_test_ballot: false,
election_name: "US 2016 General Election",
polling_location_we_vote_id_source: "wv01ploc7871",
[...]

PR #970, opened 2017-07-26
Add address alert: "Not a full address" overlay

edward-ly:
A new address alert overlay has been added in `src/js/components/Widgets/EditAddressPopover.jsx`. You can show/hide this overlay in the ballot page by clicking on the red icon between the current address and the "Edit" link. This icon should be shown only when the current address is incomplete, but this is subject to further testing.

#955


src/js/routes/Ballot/Ballot.jsx
```diff
@@ -349,7 +354,7 @@ export default class Ballot extends Component {
             </h1>
           </OverlayTrigger> :
           null }
-        <EditAddress address={voter_address} _toggleSelectAddressModal={this._toggleSelectAddressModal} />
+        <EditAddress address={voter_address} isNormalized={is_normalized} _toggleSelectAddressModal={this._toggleSelectAddressModal} />
```
DaleMcGrew: Hi Edward, could you please refactor EditAddress to take in the address={address_object}, so you don't need to calculate "isNormalized" external to the EditAddress component? (I am going to approve this pull request as-is, so there is no rush.)

DaleMcGrew:
Great work @edward-ly! 👍

PR #971, opened 2017-07-27
Refactor EditAddress component

edward-ly:
The properties of the EditAddress component have been modified so that the necessary computations are done inside the component. The computational steps themselves have also been updated and abbreviated where possible.


DaleMcGrew:
👍

Issue #989, opened 2017-08-01
convert About in HeaderBar to no longer be Link


selva-oscura:
### Please describe the issue (What happens? What do you expect?)
About in HeaderBar was mistakenly made a Link.  (It actually is an onClick trigger for toggleAboutMenu which will display actual Link elements that are surrounded by the spurious Link on About.)  Since About's Link surrounds other Links, we end up with malformed html and an error in console: 
  validateDOMNesting(...): <a> cannot appear as a descendant of <a>. See HeaderBar > Link > a > ... > Link > a.
### Steps to reproduce the problem (1, 2, 3...), including links
1 - go to website.
2 - look at console errors

DaleMcGrew:
Resolved and merged. :)

PR #990, opened 2017-08-01
Issue #989, converted About in HeaderBar from Link to span; collapsed…

selva-oscura:
converted About in HeaderBar from Link to span; 
collapsed We Vote nav__label span down to one line;

replaced Link around About with span
src/js/Navigation/HeaderBar.jsx


DaleMcGrew:
Good catch @selva-oscura! 👍

Issue #957, opened 2017-07-20
New ballot header design to be implemented


DaleMcGrew:
**Our OLD ballot header looks like this (Desktop):**
![screen shot 2017-07-20 at 10 59 43 am](https://user-images.githubusercontent.com/7756031/28431846-a295600e-6d3a-11e7-9b28-e6dd80ed4fcf.png)

**We would like it to look like this (Desktop):**
<img width="1267" alt="screen shot 2017-07-20 at 11 00 38 am" src="https://user-images.githubusercontent.com/7756031/28431876-b69af726-6d3a-11e7-8c80-c3c174bb6886.png">

**This is the OLD Mobile design:**
![screen shot 2017-07-20 at 11 01 37 am](https://user-images.githubusercontent.com/7756031/28431919-e5783162-6d3a-11e7-95b5-ea42aab8c946.png)

**...and the NEW Mobile design:**
<img width="377" alt="screen shot 2017-07-20 at 11 02 58 am" src="https://user-images.githubusercontent.com/7756031/28431952-0919f6b4-6d3b-11e7-88b4-af28ace96e33.png">

The code can be found in WebApp/src/js/routes/Ballot/Ballot.jsx

DaleMcGrew:
This looks AWESOME @edward-ly. And thank you to Yuan for the beautiful designs.

PR #1000, opened 2017-08-03
Implement new ballot header design

edward-ly:
A new layout design has been implemented to the ballot header, so some components and HTML have been moved and modified to fit the new layout. A new file `src/sass/components/_ballot.scss` has also been added to implement all of the styling changes (except for `BallotItemsSummary` which is in `_ballotList.scss`). Also included is a bit of refactoring concerning ES6 syntax and variable usage.

#957


DaleMcGrew:
Excellent work @edward-ly. 👍

PR #1002, opened 2017-08-04
Changes for facebook cover display

SailingSteve:
Added a new organizationRetrieveWithBanner() in OrganizationActions to
send display_facebook_banner boolean to server.  Call that new function
from OrganizationVoterGuide.


DaleMcGrew:
👍

Issue #999, opened 2017-08-03
Tab overflow on mobile devices to be implemented


edward-ly:
### Please describe the issue (What happens? What do you expect?)

As a follow-up to #957, we want to implement the BallotFilter tabs on the ballot page so that, on mobile devices, they overflow to the right and the user can scroll between them. Currently, the ballot look like this:

![screenshot 2017-08-03 14 57 06](https://user-images.githubusercontent.com/11434331/28945725-7b4b3172-785c-11e7-9aca-cc9bb53ce996.png)

### Steps to reproduce the problem (1, 2, 3...), including links

1. Go to http://localhost:3000/ballot
2. Open your browser's developer tools (Ctrl + Shift + I on Windows)
3. Enable the toggle device toolbar (Ctrl + Shift + M on Windows)
4. Select "iPhone 5" or a similar display in the menu above the web page

DaleMcGrew:
Hi @edward-ly, this is a great step forward. Thank you. I have two requests:
1) Prevent scrolling when all of the tabs are narrower than the space they are in (so the horizontal scrolling wouldn't happen in desktop mode). 
2) Only scroll to the right so that you can see all of the text that is off-screen to the right. Currently you can scroll to the right so that the tabs are completely gone.

Thanks!

DaleMcGrew:
Excellent work @edward-ly -- this issue is now closed.

PR #1015, opened 2017-08-09
Implemented overflow for BallotFilter tabs

edward-ly:
Using only SCSS (and rearranging some HTML), the ballot filter tabs now remain on one line and have the ability to horizontally scroll through them (in both desktop and moble versions). The only two minor issues right now are that the scrollbar below the tabs (which is supposed to be hidden) may still be visible if you change the zoom level in your browser, and there is a ton of extra space to the right of the tabs that you can scroll through.

#999


DaleMcGrew:
Thank you @edward-ly! 👍

Issue #1006, opened 2017-08-06
Add "Close search" x to search boxes


DaleMcGrew:
### Please describe the issue (What happens? What do you expect?)
In the main search box we have added an "x" that clears the search text from the search box:
![screen shot 2017-08-06 at 9 13 54 am](https://user-images.githubusercontent.com/7756031/29005049-ffef5084-7a87-11e7-8950-65dfc7949d5d.png)

Please add this "x" to any other search box in We Vote. Some places:
- [ ] Your own voter guide, "Who You are Following"
- [ ] Network > All Friends: http://localhost:3000/friends
- [ ] Network > Organizations to Follow: http://localhost:3000/more/network
- [ ] Others?

josh-southern:
I started working on this tonight. Going to work on abstracting the search bar with close and button options tomorrow. Is it ok to self assign?

DaleMcGrew:
Thank you! Yes, please feel free to self assign. One request--I think we may want to avoid doing any refactoring on the main header search box right now, but abstracting for all of the other search boxes sounds great!

josh-southern:
Dropped the new component into /more/network, but I'm unsure how to see the /friends search box. Do I need the backend to add friends and test that out?

DaleMcGrew:
Hi @josh-southern, at this time adding friends is a little hard to test, so I think it is fine to consider this issue closed by putting it on the network page. I can migrate it to the harder-to-test pages. Thank you!

DaleMcGrew:
Hi @josh-southern, I just got a chance to do some testing. This is mostly in great shape with one request. On all pages that use the new search box, could you please add the magnifying glass icon in the resting state (when the search box is empty) like you do on the scratchpad ( http://localhost:3000/testing/scratchpad )?

Pages where I don't see the magnifying glass:
http://localhost:3000/more/network
http://localhost:3000/opinions

josh-southern:
Just flipped the flag. Should be showing on both pages now.

DaleMcGrew:
Great work @josh-southern -- this issue is now complete.

PR #1017, opened 2017-08-10
search bar with options and scratchpad used for testing

josh-southern:
https://github.com/wevote/WebApp/issues/1006
In addition to the issue mentioned it does include a testing route and component. Review on the SCSS include placement would be appreciated.

I'm not entire happy with this `this.updateInputValue({ target: { value: "" }});` shim I used for clearing with the passed in update function, but I haven't really come up with a better version of that besides revoking access to the `event` object entirely and just passing the update function the input value.


DaleMcGrew:
Thank you @josh-southern -- I love learning from reading your code. I really like the idea behind setting up a scratchpad too! 👍

Issue #982, opened 2017-07-29
Navigation bar icons too large in mobile


DaleMcGrew:
### Please describe the issue (What happens? What do you expect?)
When looking at the secondary navigation bar on a mobile device, the icons (above "Issues", "Organizations", "Positions" etc.) look a little to big. (They are fine for desktop.)
<img width="396" alt="screen shot 2017-07-28 at 9 16 11 pm" src="https://user-images.githubusercontent.com/7756031/28742046-27af0e48-73da-11e7-870e-e813639bf767.png">

### Steps to reproduce the problem (1, 2, 3...), including links
1. Put your browser in phone mode
2. Go to ballot

DaleMcGrew:
Thank you @edward-ly, your fix resolves this.

PR #1018, opened 2017-08-11
Fixed overflow behavior for BallotFilter tabs

edward-ly:
Fixed an issue where there is extra space to the right of the ballot filter tabs that you can scroll through (and an issue where the ballot header background does not extend the entire width of the page). The unnecessary ability to scroll when the width of the page is longer than the width of the tabs has subsequently been removed.

#999 

APPEND: The latest commits also resolve issue #982.


DaleMcGrew:
These changes are a big step forward @edward-ly, thank you! 👍

Issue #993, opened 2017-08-01
Ballot item summary sidebar lingers when address is removed


edward-ly:
### Please describe the issue (What happens? What do you expect?)

When you edit your ballot address by clearing it, the ballot item summary sidebar should also be hidden, but it still lingers.

![screenshot](https://user-images.githubusercontent.com/11434331/28850721-34925600-76d3-11e7-8ee9-67ae648b9ce0.png)

### Steps to reproduce the problem (1, 2, 3...), including links

1. Go to http://localhost:3000/ballot
2. Enter an address to view a ballot and ballot summary if you have not already
3. Click on "Edit" to show the SelectAddress modal
4. Clear the address the address box and click "Save"

jpnelson:
For this issue, would it be better if it were not possible to have no address? Or is it supported to have no address?

jpnelson:
We clarified that it should not, since you can get into this state in other ways

DaleMcGrew:
Thank you @jpnelson!

PR #1024, opened 2017-08-17
fixes 993

jpnelson:
#993


DaleMcGrew:
Thank you @jpnelson, that fixed it! 👍

Issue #972, opened 2017-07-27
New design: candidate display on ballot page


DaleMcGrew:
We would like to change the display of candidates to this (DESKTOP):
<img width="800" alt="screen shot 2017-07-26 at 5 56 34 pm" src="https://user-images.githubusercontent.com/7756031/28649746-e47ffb14-722b-11e7-96f7-527177784602.png">

...and this for MOBILE:
<img width="356" alt="screen shot 2017-07-26 at 5 58 08 pm" src="https://user-images.githubusercontent.com/7756031/28649761-09b9f7d6-722c-11e7-9e11-d13035dcdd5a.png">

Notes:
- We want the position bar (green & red bar - ItemSupportOpposeCounts) to always be left aligned under the candidate's name, to the right of the photo.
- The voter guides a person might want to follow (ItemTinyOpinionsToFollow) should be centered on desktop. In mobile mode, ItemTinyOpinionsToFollow can be displayed if the position bar doesn't have any values.
- The candidate's name should not exceed one line in mobile mode, and can "fade" to "..."
- The description line (Party + twitter_description) can go to the right of the candidate's name in desktop, and on the next line in mobile mode.

Once you click "more", unfurl the description text to its full length, and add a "learn more" link at the end, which takes you to the candidate page.
<img width="358" alt="screen shot 2017-07-26 at 6 31 27 pm" src="https://user-images.githubusercontent.com/7756031/28650426-c5067164-7230-11e7-89d0-7163eac24a12.png">

Please start in WebApp/js/components/Ballot/OfficeItemCompressed.jsx

DaleMcGrew:
Great work @edward-ly!

PR #1025, opened 2017-08-18
Redesigned candidate display on ballot page

edward-ly:
The layout of each ballot item in the ballot page has been redesigned to fit the mock-up as closely as possible (except for the candidate name, which is still currently on a separate line). I also added some more custom styling and JavaScript implementation to handle what components are displayed in mobile mode. A new LearnMore component has been created from ReadMore so that there is a "learn more" link instead of a "less" link.

#972


DaleMcGrew:
👍  This is an amazing upgrade @edward-ly -- thank you!

PR #1030, opened 2017-08-21
Update styling on network and voter guide tabs

edward-ly:
The CSS styling of the tabs in the network and voter guide pages has been updated to be consistent with the styling on the ballot page. A new component file `_tabs.scss`  has been created to make these tabs more reusable.


DaleMcGrew:
Great work @edward-ly! 👍

Issue #1032, opened 2017-08-22
Network buttons layout improvement


n0f3:
### Please describe the issue (What happens? What do you expect?)
The layout for the network buttons is currently fairly static and doesn't utilize the negative space very well. I would love to improve on that and provide a more scalable and dynamic solution

### Steps to reproduce the problem (1, 2, 3...), including links
https://wevote.us/more/network/, take a look at the build your network section

DaleMcGrew:
Thank you @n0f3 for these improvements! I found these bugs:
**Bug A) Network buttons break onto two lines in Desktop mode**
1) Refresh browser on http://localhost:3000/more/network (when you are not signed in) -- layout is fine.
2) Slowly reduce the width of the browser window and you will see this around 829px width -- all three buttons should remain on one horizontal line:
![screen shot 2017-08-22 at 6 13 17 am](https://user-images.githubusercontent.com/7756031/29567117-12ec9180-8701-11e7-9781-481184694174.png)

**Bug B) Descriptions go away in desktop**
3) Keep narrowing the window until the buttons go into mobile mode, then widen the window until they jump back to desktop mode. The descriptions go away:
![screen shot 2017-08-22 at 6 14 09 am](https://user-images.githubusercontent.com/7756031/29567203-4a1154ca-8701-11e7-9bc7-2666fbc0af54.png)

**Bug C) Network buttons should fit on one horizontal line in mobile mode (iPhone 5).**
Removing the current margins might be enough space. If that doesn't work we could take the text down in font size.
![screen shot 2017-08-22 at 6 16 30 am](https://user-images.githubusercontent.com/7756031/29567289-945f2f98-8701-11e7-81dd-05a75a060cbb.png)

n0f3:
Sounds good, I'll make sure to fix them up!

DaleMcGrew:
Hi @n0f3, are you available to come to Code for SF tonight? I'd love to get your changes checked in.

n0f3:
@DaleMcGrew I've fixed the issues relative to spacing and style. The TextTruncate component is what's causing the bugs with the text going away and it's due to it detecting that the parent div's width is 0, but never updating itself. 

There is no way that I can see to easily detect changes in styles in the page and force and update to the component.

I think issues with TextTruncate not updating itself should be filed as a different bug, as it will affect other components that rely on display: none to hide/show elements depending on resolution

DaleMcGrew:
Thank you @n0f3, your changes look great!

[...]

PR #1033, opened 2017-08-22
Issue1032

n0f3:
#1032 
Updated network stylesheet, included container for buttons that uses flex display, fixed warning in stylelint related to depracated rule


DaleMcGrew:
👍

src/sass/components/_network.scss
```diff
@@ -1,12 +1,44 @@
...
+$button-margin-horizontal: 10px;
+$description-margin-top: 10px;
+
+.buttons-container {
+  display: flex;
+  flex-flow: row wrap;
+  justify-content: space-around;
+}
+
+@media (min-width: 768px) {
```
ixley: *Use the breakpoints mixin with existing breakpoint variables instead of @media queries w/ raw values.

src/sass/components/_network.scss
```diff
@@ -1,12 +1,44 @@
...
+  .network-btn {
+    display: inline-block;
+    max-width: $button-div-max-width;
+    margin-right: $button-margin-horizontal;
+    margin-left: $button-margin-horizontal;
+  }
+}
+
+@media (max-width: 767px) {
```
ixley: *Use the breakpoints mixin with existing breakpoint variables instead of @media queries w/ raw values.

ixley: *Use the breakpoints mixin with existing breakpoint variables instead of @media queries w/ raw values.

src/sass/components/_network.scss
ixley: *Use the breakpoints mixin with existing breakpoint variables instead of @media queries w/ raw values.

src/sass/components/_network.scss
```diff
@@ -1,12 +1,44 @@
-.network-btn {
-  display: inline-block;
-  max-width: 240px;
-  margin-right: $space-sm;
-  margin-left: $space-sm;
+$full-width: 100%;
```
ixley: Just FYI, part of the point of requiring variables for style declarations is to try and avoid so many custom values. There are a number of things here that seem unnecessarily customized or are overly specific (in a way that looks very brittle). See what you can use of the existing spacing variables with a goal of removing as many of these custom variables as possible. Feel free to ping me if you have questions.

src/sass/components/_network.scss
```diff
@@ -1,12 +1,44 @@
...
+@media (max-width: 767px) {
+  .network-btn {
+    display: inline-block;
+    margin-left: $button-margin-horizontal / 2;
+  }
 }
 
 .social-btn-description {
-  margin-top: $space-sm;
+  margin-top: $description-margin-top;
```
ixley: again, this seems unnecessarily specific. Why did this need a unique custom value?

Issue #981, opened 2017-07-29
Edit Voter Guide on Mobile: Remove word "Support" or "Oppose" so the thumbs-up/down icons appear on same line


DaleMcGrew:
### Please describe the issue (What happens? What do you expect?)
On your voter guide, the ItemActionBar (where you choose support or oppose) gets broken into two lines. We would like to turn off the word "Support" or "Oppose" so there is room for the component to render on one horizontal line.
![voter guide editing-itemactionbar](https://user-images.githubusercontent.com/7756031/28741882-94911ca4-73d5-11e7-8691-40257a04da75.PNG)

### Steps to reproduce the problem (1, 2, 3...), including links
1. Resize your browser to phone size, then sign in.
2. Go to the ballot and support or oppose one candidate.
3. Click on your profile photo on the top right of the navigation bar, and go to "Your Voter Guide"
4. Click "Edit Positions"
5. You should see the visual glitch shown above.

WebApp/src/js/components/VoterGuide/OrganizationPositionItem.jsx (which contains ItemActionBar) is called from WebApp/src/js/components/VoterGuide/VoterGuidePositions.jsx

DaleMcGrew:
Nailed it! Thank you @edward-ly.

Issue #1029, opened 2017-08-21
Measures on Ballot Page: Same upgrade as was done for Candidates


DaleMcGrew:
### Please describe the issue (What happens? What do you expect?)
Similar to the closed Issue #972, we would like to implement the same "Position Bar + organizations + support/oppose buttons" layout for Measures as was done for Candidates.

DaleMcGrew:
Awesome work @edward-ly. 👍  Issue fixed.

PR #1034, opened 2017-08-22
Update measures layout on ballot page, remove support/oppose labels in mobile view

edward-ly:
The layout of each measure item on the ballot page has been updated to match that of each office item (except for the measure description, which may be updated as a separate issue if needed).

#1029 

UPDATE: the latest commit also resolves issue #981.


DaleMcGrew:
👍

Issue #1036, opened 2017-08-24
Place cursor in input field when you edit address


DaleMcGrew:
### Please describe the issue (What happens? What do you expect?)
When a voter goes to change their address, you need to click on the "Enter address where you are registered to vote" box before you can start typing.

### Steps to reproduce the problem (1, 2, 3...), including links
1. Go to Ballot page
2. Click "Edit" next to the address field
3. A modal popup appears
We want the cursor to be in the input field so a person can start typing without having to click in the input field first

edward-ly:
I found two possible solutions for this one. One would just be to add the `autofocus` attribute to the text `<input />`. Another would be to add `this.refs.autocomplete.focus();` inside the `componentDidMount` function (this is all in `src/js/components/AddressBox.jsx`, by the way). Let me know which implementation you prefer.

DaleMcGrew:
If adding autofocus to the input works, that sounds simplest!

edward-ly:
OK, I can include it in my next pull request (or if you want to add it in yourself, be my guest).

DaleMcGrew:
Solved! Thank you @edward-ly.

Issue #974, opened 2017-07-27
Next buttons on Modals covering text


DaleMcGrew:
### Please describe the issue (What happens? What do you expect?)
On some of the modals, we are seeing the "Next" button covering text. We need to resize the content or make some other adjustments. I have also seen this on mobile devices.
![ballotintromission-smaller screen](https://user-images.githubusercontent.com/7756031/28653873-eeaafd6e-7244-11e7-9514-763ed1bb98d5.png)

### Steps to reproduce the problem (1, 2, 3...), including links
1. Resize your browser to a small size like this: 775 x 440
2. Navigate to the ballot page
3. Click "(show intro)"

The 4 components that make up this Ballot intro modal are:
WebApp/js/components/Ballot/BallotIntroMission.jsx
WebApp/js/components/Ballot/BallotIntroFollowIssues.jsx
WebApp/js/components/Ballot/BallotIntroFollowAdvisers.jsx
WebApp/js/components/Ballot/BallotIntroPositionBar.jsx

DaleMcGrew:
Resolved -- thank you @edward-ly!

PR #1038, opened 2017-08-24
Fix layout of ballot intro modals, add autofocus to address input

edward-ly:

#974, #1036 

Fixes the layout of the modals shown in the ballot intro. Adds the autofocus attribute to the edit address text input so users can start typing their address without clicking on it first.


DaleMcGrew:
👍 Thank you Edward!

PR #1037, opened 2017-08-24
next button fix, ballot follow countdown

josh-southern:
### What github.com/wevote/WebApp/issues does this fix?
https://github.com/wevote/WebApp/issues/1023
### Changes included this pull request?
Changed the follow issue ballot modal to do a count down when they hit the follow buttons.

PR #1039, opened 2017-08-24
Fix modal dots bug, update SCSS styling

edward-ly:
Fixed an issue where the previous pull request (#1037) caused the modal dots to appear in the wrong place (solution: deleted duplicate class definitions). Updated SCSS styling by removing some local variables (in favor of using `// stylelint-disable-line sh-waqar/declaration-use-variable` instead). Update `_tabs.scss` to use LF line feed (for consistency).


DaleMcGrew:
Hi @edward-ly, I see where you are going with this, but I am concerned about moving away from using style variables on many of these lines. I am cc'ing @ixley so we can get his thoughts. I see the value in using direct values when it is a  style that is one-of-a-kind, but some of the variable names you are replacing seemed useful.

edward-ly:
Perhaps I could bring some variables back, but I saw @ixley's comments about custom variables in #1033, so I thought I'd follow that example.

DaleMcGrew:
Thank you @edward-ly, maybe we can wait until we hear what Jeff thinks, and decide once he has weighed in.

ixley:
@edward-ly Sorry if I was unclear about the customized variables. Variables are valuable when they are able to convey context for the values, particularly when they are reusable. If there is a single custom value, it doesn't normally warrant creating a unique variable because the added layer of abstraction is not useful. If a variable gets used in more than a couple places, even if it's within a single component, it makes more sense. 

What I was actually trying to call out was not using custom values or custom variables, but to try not to use custom values in general. For example, there are spacing variables that have already been set up and there is a place where you are listing out the exact same values as raw values instead of using the spacing variables. In other cases, there are specific values but it's not clear to me whether the specificity of a few pixels warrants using custom values as opposed to trying to conform to the existing defaults. To be fair, the system is not fully built out and so I'm not saying there is no reason to add new values or custom ones in specific contexts, but I am urging you to try hard not to unless there's a really good reason. Does that make a bit more sense? Have you read some of the styling [conventions/guidelines](https://github.com/wevote/WebApp/blob/develop/docs/working/STYLING.md)?

edward-ly:
Thanks for the feedback, @ixley. Now, I used the spacing variables wherever possible, and I re-added all of the variables that I believe add context to the styling (all while making sure there are still no lint warnings).

DaleMcGrew:
Thank you @edward-ly and @ixley!

Issue #1044, opened 2017-08-25
Modal buttons should be full width


DaleMcGrew:
### Please describe the issue (What happens? What do you expect?)
In WebApp/src/js/components/Ballot/BallotIntroFollowIssues.jsx the buttons should be full width like this:
![screen shot 2017-08-24 at 5 49 58 pm](https://user-images.githubusercontent.com/7756031/29695005-e09ea5b0-88f4-11e7-8762-a7e0dc4047ce.png)

![screen shot 2017-08-24 at 5 50 26 pm](https://user-images.githubusercontent.com/7756031/29695003-dcd5373c-88f4-11e7-94a7-1411387f86e9.png)

On iPhone they are not, and they look like this:
![img_4179 png](https://user-images.githubusercontent.com/7756031/29695172-1523470e-88f6-11e7-9f51-2e46c2799f94.png)

We tried to style this with width 20% in WebApp/src/sass/components/_intro-modal.scss but that hack falls apart on an actual mobile device:
`  &__button {
    position: absolute;
    bottom: 42px;
    width: 20%;
    @include breakpoints (max mid-small) {
      bottom: 0;
    }
  }

  &__button-follow {
    width: 20%;
  }

  &__button-disabled {
    width: 20%;
  }
`

DaleMcGrew:
Fixed.

Issue #860, opened 2017-05-25
Choose election modal: buttons should be same width


DaleMcGrew:
Upcoming election(s) and Prior election(s) buttons should all be the same width, with text centered.
<img width="638" alt="screen shot 2017-05-24 at 7 46 44 pm" src="https://cloud.githubusercontent.com/assets/7756031/26434326/ff4c0208-40bc-11e7-8852-ea2c493093ca.png">

This component is in WebApp/src/js/components/Ballot/BallotElectionList.jsx

DaleMcGrew:
Fixed.

PR #1049, opened 2017-08-29
Make intro modal buttons fixed width, election modal buttons full width, text truncated and centered

edward-ly:
Set the width of the intro modal buttons to a constant 180 pixels to fix the appearance of the buttons on moble (#1044). Fixed the grammar of the messages in the candidate list and issues modal when count is at 1. Changed the election modal buttons so that it covers the full width of the modal, all text is centered, and text is hard-limited to a set number of characters (#860). Fixed an issue with the slick dots appearing in the wrong place for certain modals.


DaleMcGrew:
👍

PR #1052, opened 2017-08-30
Fix TextTruncate in ballot item office name, intro modal minimum height

edward-ly:
Fixed an issue where the ballot item office name was not truncating properly on mobile devices. Fixed a bug where the navigation button may still cover some text in the organizations intro modal by forcing a separate minimum height in desktop view.


DaleMcGrew:
👍 Great work.

Issue #1054, opened 2017-08-30
Add "Get Started / Create Voter Guide" bar on Welcome page


DaleMcGrew:
We want to add a horizontal bar that is split in half. On the left should be "Get Started" and on the right should be "Create Voter Guide". When you hover over either side it should turn dark blue. "Get Started" should click through to ballot. "Create Voter Guide" should go to https://wevote.us/more/sign_in for now (I am working on a new destination page.)

Please reorient the text above this text per the designs in desktop. Mobile can remain the same.
View your ballot.
Learn from friends.
Share your vision.

![getstarted-desk-voterguide-bar1](https://user-images.githubusercontent.com/7756031/29894137-30ad343e-8d89-11e7-9253-4e288f47a268.png)

![getstarted-mobile-voterguide-bar1](https://user-images.githubusercontent.com/7756031/29894145-36983b78-8d89-11e7-92fd-9034ba450991.png)

![getstarted-desk-voterguide-bar2-hover](https://user-images.githubusercontent.com/7756031/29894157-3c2a6bb0-8d89-11e7-8842-747c437af42b.png)

DaleMcGrew:
Verified!

Issue #953, opened 2017-07-20
Welcome page: Add descriptive text under icons


DaleMcGrew:
### Please describe the issue (What happens? What do you expect?)
When you go to the welcome page, there are 6 icons. We would like to add descriptive text (in one way for desktop, and another way mobile).

In **Desktop** mode, this is the desired size/layout of the text under each item. Please see actual text to use at the bottom of this issue. Please do not change the visual relationship between the icon and the title.
<img width="789" alt="screen shot 2017-07-20 at 8 08 30 am" src="https://user-images.githubusercontent.com/7756031/28424572-cf72dd80-6d22-11e7-880b-3b4b719cc2c3.png">

In **Mobile** mode, the text should not show until the icon or title is clicked. When the icon or title is clicked, the icon should go away, the title should be moved up to where the top of the icon was, and the text added below that. Clicking the title or text should "flip" the display back to the icon plus title. We are trying to minimize the shifting of the other icons.

**Descriptive text to add**
View Your Ballot
You will see your actual ballot, including candidates and measures.

Learn From Organizations
Follow the voter guides of groups you trust. See what they support or oppose.

See Your Network's Positions
See how many (in your network) support or oppose each candidate or measure.

Invite Friends to Your We Vote Network
Talk politics with friends who share your values. Avoid flame wars!

Share Your Vision
Empower other voters with information and help your friends.

Decide & Vote
Bring We Vote on your phone or a print-out to cast your vote with confidence.

DaleMcGrew:
@emilymorgado we miss you! Let me know if you would like to find a time during the week to get on a hangout if you'd like help with a pull request. I can also check in the files if you'd like to email them to me: dale.mcgrew@WeVote.US

DaleMcGrew:
Resolved!

PR #1063, opened 2017-09-05
Welcome Page: Add descriptive text and "Get Started / Create Voter Guide" bar

edward-ly:
Added a horizontal bar splitting the Welcome page that contains links to the ballot and sign-in pages, and re-oriented the the header text and input form above (#1054). Added descriptive text under each feature item with different behavior depending on display size. On desktop, the font size of these descriptions change when clicked. On mobile, either the image or the description is shown, and clicking on the feature item toggles between them (#953). Animations between each state are currently not supported (since `visibility` does not have a gradual transition), but may be added later with further refactoring.


DaleMcGrew:
This is a great upgrade @edward-ly, thank you! 👍

PR #1065, opened 2017-09-06
Server side issue 431, Persist and use facebook background offsets.

SailingSteve:
Zero can be a valid value for facebook offsets, so handle data structure
accordingly.
Also remove the old facebookFriendsAction that asked facebook to send
all your fb friends via the "friends" API, which no longer worked for
our purposes.  Now we call the "games" API from the client to get a list
of your fb friends.


DaleMcGrew:
👍

Issue #1062, opened 2017-09-05
Copy / UI tweaks to onboarding screens


mshmsh5000:
Some suggestions intended to make the onboarding experience a little bit snappier and less formal, give the copy a sense of momentum, and tweak the layout for easier scanning.

## Screen 1: Issues

- Title: **What do you care about?**
- Sub: _Pick five issues to see matching organizations!_
- Context language all above selection
- Bigger grid window

Reference image: https://cl.ly/1i3I191e271d

## Screen 2: Organizations

- Title: **Follow organizations**
- Move helper language above selection window, & shorten to _Great work! Next, follow any of these organizations to see where they stand._
- Increase `max-height` of selection window to `50vh` (currently at `30vh`)

Reference image: https://cl.ly/2N3u1o082B3U

## Screen 2A: Orgs / voter guides

This screen is not explicitly navigable: the user needs to select all issue-related orgs to follow before the screen appears. If the user simply clicks "Next" after selecting a subset of orgs, they'll skip this screen. This is true of all org listing screens in this step.

- Recommended: Aggregate all issue-based orgs into a single view. _Con:_ Conflates orgs and voter guide providers (but IMO this isn't a necessary distinction during onboarding)
- Alternate 1: Remove, as it's confusing. _Con:_ This opens the problem of how to introduce this concept later in the experience
- Alternate 2: Make this an explicit third screen, so that there are four total. _Con:_ Adds another onboarding step before showing value 
  - Reference image: https://cl.ly/2q3K2D1C081m

## Screen 3: Position bar

- Some suggestions for less formal copy.

Reference image: https://cl.ly/0G0Y0G1l0p3r

aliciaprevost:
On Screen 1, I like Matt's suggestion to reframe as a question. Maybe after "What do you care about?" the subheading could be: "Tell us what issues are important to you and we'll suggest organizations that are working on those issues." Maybe too long... but trying to get across the point that the organizations we suggest are working on solving the issues you care about...

My question: why does it need to be 5? 

Also - if we are framing this as "what issues you care about" (which I think we should) then I think we should eliminate some of the issue headings - the ones that aren't really issues (including: conservative values, christian living, progressive values, etc.) Also don't want to lead with the political parties - I think the value of this issue frame is that it circumvents parties (since people who just want to vote straight party ticket don't need a voter guide anyway)

DaleMcGrew:
[...]

PR #1074, opened 2017-09-11
Adjust size and position of various intro modal elements, fix intro modal button width and placement

edward-ly:
Increased the height and width of the intro modal, as well as the number of issues shown per row, on desktop displays. Increased the height of the vertical scroll container to `40vh`-`42vh` (not quite `50vh` as suggested in #1062, but I think it looks big enough the way it is now). Adjusted the size and positioning of various text elements in the modal. Fixed an issue where the bottom navigation button is not positioned correctly on some displays, and adjusted the button's width. Added breakpoint variables that align with Bootstrap's breakpoints.


DaleMcGrew:
Thank you Edward! 👍

Issue #1072, opened 2017-09-09
Prevent welcome page form submission without valid email


DaleMcGrew:
### Please describe the issue (What happens? What do you expect?)
I was showing someone WeVote.US on their mobile phone and they were part way through entering their email address on the welcome page when something they pressed caused the form to submit.

### Steps to reproduce the problem (1, 2, 3...), including links
1. Go to https://WeVote.US/welcome
2. Click "Sign Up"
3. The form submits and it shouldn't.

Please make the Sign Up button disabled until a valid email has been entered. Hitting "Enter" before a valid email has been entered should also prevent submission. Please make sure the behavior on this page matches, including using the same email validity-checking function: http://localhost:3000/more/sign_in

DaleMcGrew:
Fixed! Thank you @bharathdn.

PR #1075, opened 2017-09-12
Prevent welcome page form submission without valid email

bharathdn:
Prevent welcome page form submission without valid email #1072


DaleMcGrew:
👍

PR #1076, opened 2017-09-12
Fix vertical scroll container height

edward-ly:
Fixed an issue where the issues grid container has an incorrect height depending on the height of the display.


DaleMcGrew:
👍

PR #1077, opened 2017-09-12
Increase modal and vertical scroll container height on desktop

edward-ly:
Increased the height of the modal and the issues grid container on desktop displays. Prevented the webpage from scrolling while the modal is visible.


DaleMcGrew:
👍

Issue #1079, opened 2017-09-12
Bookmarks page needs to be updated with new Ballot page header


DaleMcGrew:
Migrate changes made in this component:
WebApp/src/js/routes/Ballot/Ballot.jsx

...over to this component:
WebApp/src/js/components/Bookmarks/Bookmarks.jsx

biomassives:
This message was created automatically by mail delivery software.

A message that you sent could not be delivered to one or more of its
recipients. This is a temporary error. The following address(es) deferred:

  acmeideal@gmail.com
    Domain biomassiv.es has exceeded the max emails per hour (110/100 (110%)) allowed.  Message will be reattempted later

------- This is a copy of the message, including all the headers. ------
Received: from github-smtp2-ext6.iad.github.net ([192.30.252.197]:46046 helo=github-smtp2b-ext-cp1-prd.iad.github.net)
	by chi-server32.websitehostserver.net with esmtps (TLSv1.2:ECDHE-RSA-AES256-GCM-SHA384:256)
	(Exim 4.89)
	(envelope-from <noreply@github.com>)
	id 1drqP9-002Wia-QY
	for greg@biomassiv.es; Tue, 12 Sep 2017 13:59:27 -0500
Date: Tue, 12 Sep 2017 11:58:47 -0700
DKIM-Signature: v=1; a=rsa-sha256; c=relaxed/relaxed; d=github.com;
	s=pf2014; t=1505242727;
	bh=4U/xeFCVSFDUo2sDeXb4zHoekgbGjOGVIlBc6TFLjBY=;
	h=From:Reply-To:To:Cc:Subject:List-ID:List-Archive:List-Post:
	 List-Unsubscribe:From;
	b=cE75Jr84+BtM0m5aAhg3yw5Lz0MLJWqxRKM76ZEUTor/85V/85Yzircfuzj3GOhIl
	 N6/hOxJf01l3vJJJnxpCtVbks+7hCfPRfT+RM2VkpvJk7hGDcVEtIrf8HwbnEmIeeo
	 5x1kWOpW9AotNvxFMr/8e5ejmIKHKxFCT9zP33sQ=

DaleMcGrew:
Verified!

PR #1082, opened 2017-09-13
Update Bookmarks page with new Ballot page header

bharathdn:
Update Bookmarks page with new Ballot page header #1079


DaleMcGrew:
👍

PR #1086, opened 2017-09-14
Fix ballot header width on bookmarks page

edward-ly:
Fixed a bug where the width of the ballot header on the bookmarks page is not extending the entire width of the page.

#1079


DaleMcGrew:
👍

Issue #951, opened 2017-07-20
Choose Facebook friends in mobile mode: Style problem


DaleMcGrew:
### Please describe the issue (What happens? What do you expect?)
In mobile mode, (iPhone 5 or similar small view), the picture + name of friend wraps and does not look good:
![img_4068](https://user-images.githubusercontent.com/7756031/28399221-ae360820-6cc0-11e7-80fc-9754b530aab5.PNG)

The code that powers the list of friends is here:
WebApp/src/js/routes/FacebookInvitableFriends.jsx and each "friend checkbox" can be found by searching in this file by: facebook_friend_list_for_display
...and WebApp/src/js/components/Connect/CheckBox.jsx

DaleMcGrew:
Verified.

PR #1091, opened 2017-09-15
Redesign Facebook friends list, modify issues text size and styling

edward-ly:
Display the Facebook friends to invite list as a grid similar to that of issues to follow (#951). Add and style default text inside each grid whenever there are no items in them. Fix the font size of the issues grid text in the network page.


DaleMcGrew:
👍

Issue #1090, opened 2017-09-15
Add popover for Position bar


DaleMcGrew:
Add popover on the Ballot page when you hover over the position bar. If support or oppose positions exist, text should say:
This is a summary of the "support" and "oppose" positions from your network. Click to see who supports and who opposes this ballot item, and any other organizations you can follow.

If NO support or oppose positions exist, text should say:
This will show a summary of the "support" and "oppose" positions from your network once you have followed any voter guides that have positions on this ballot item. Click to see organizations you can follow that have positions on this ballot item.

DaleMcGrew:
Please turn off the "Click to see more detail." text when the position bar is not clickable.
![screen shot 2017-09-19 at 10 04 30 am](https://user-images.githubusercontent.com/7756031/30605246-38c9bf04-9d22-11e7-9058-65ef9f7812cb.png)

DaleMcGrew:
Verified!

PR #1097, opened 2017-09-19
turn off the "Click to see more detail." when Position bar is not clickable

bharathdn:
turn off the "Click to see more detail." when Position bar is not clickable

#1090


DaleMcGrew:
👍

PR #1104, opened 2017-09-22
Redesign organizations to follow as grid, add descriptions to each organization and issue square,  fix ESLint styling

edward-ly:
Redesigned the list of organizations in the ballot intro modal as a grid similar to the issues grid. Added an icon to each issue or organization square that reveals the issue or organization description on hover (if there is one). Removed the remaining ESLint warnings.


DaleMcGrew:
Thank you for helping to make We Vote so beautiful! 👍

PR #1116, opened 2017-09-29
Redesign welcome page header image and sign up form

edward-ly:
Updated the design of the sign up form on the welcome page. Fixed the size of the welcome header image so that the scaling does not exceed its actual size (and added a new header background color for large displays).


DaleMcGrew:
Thank you @edward-ly! 👍

Issue #1089, opened 2017-09-15
iPhone Modal buttons formatting


DaleMcGrew:
On the iPhone we still have some display problems that we are hoping a developer with a Mac can fix. Some of these issues can be reproduced in Safari on the Mac.

- [ ] ISSUE 1: Next button and carousel dots show "below the fold":

![img_4201 png](https://user-images.githubusercontent.com/7756031/30495531-2e5ed686-9a01-11e7-80ba-08c97b2cff0a.png)

- [ ] ISSUE 2: Buttons at bottom of the modal are shifted to left or right in Safari/iPhone only

![buttons to the right](https://user-images.githubusercontent.com/7756031/30495638-a71c38e8-9a01-11e7-8da9-7466f051ff39.png)

![img_4205 png](https://user-images.githubusercontent.com/7756031/30495683-cf7ea4ce-9a01-11e7-8b07-4dce32b84c13.png)

Steps to reproduce:
1) Go to your ballot
2) Click on the Issues button

The modal is opened from:
WebApp/src/js/components/Navigation/HeaderGettingStartedBar.jsx

This is the first component:
WebApp/src/js/components/Ballot/BallotIntroFollowIssues.jsx

ixley:
@DaleMcGrew I poked into this further and I'm no longer seeing this rendering bug. I have the latest versions of Safari on Mac and iOS so perhaps this resolved itself?

DaleMcGrew:
It is still there on my iPhone 6.

ixley:
This is how it renders for me on Safari / iOS 11 / iPhone 6
![img_1081](https://user-images.githubusercontent.com/160662/30774725-2798bf6a-a03c-11e7-9651-59d7cdb28798.PNG)

DaleMcGrew:
@josh-southern your solution works really well, thank you!

PR #1118, opened 2017-09-29
Issu1089

josh-southern:
https://github.com/wevote/WebApp/issues/1089

Some positioning rejigging on the small media breakpoint taking into account Safari's rendering of heights. I did investigate some Safari on iPhone specific targeting, but I am hesitant to implement as they are extremely hacky and likely prone to future breakages.


DaleMcGrew:
Thank you Josh! 👍

Issue #1112, opened 2017-09-28
we_vote_branding_off incoming URL variable should change header bar


DaleMcGrew:
### Please describe the issue (What happens? What do you expect?)
Partner organizations can put We Vote in a frame and pass in variables to change how We Vote is displayed. ([Full description here](https://docs.google.com/document/d/1VACOrXGKzBvgqlGAG2RnFu89NM7T57ZPeMsWZkK4PNM/edit#).)

- [x] Store a cookie ("we_vote_branding_off") that expires after 1 day
- [x] If "we_vote_branding_off" is set, then change the display of the header these ways:
Header will will show with “Ballot” button, “Network button” and search bar.
Header will NOT show We Vote, About or Donate.
Mobile profile drop down will NOT show Getting Started, About, or Donate.
- [x] When we set "we_vote_branding_off", also set "show_full_navigation" cookie -- I think with the branding off we always want to show_full_navigation.
- [x] On the welcome page, turn off "Sign up for updates..."
- [x] On the welcome page, turn off "Our Network" section
- [x] On the welcome page, turn off "Please share or donate" block, including the links starting with "About", "Our Vision", etc. -- please leave everything below "WeVote.US is brought to you..."
- [x] Please change the color of the header nav bar to a dark gray (instead of dark blue) when "we_vote_branding_off" is set

DaleMcGrew:
Issue complete!

Issue #1106, opened 2017-09-24
IntroModal appears and then disappears quickly on sign in


DaleMcGrew:
### Please describe the issue (What happens? What do you expect?)
When you sign in, the IntroModal with all of the Issue pictures flashes on screen for a split second, and then disappears. This is disorienting.

### Steps to reproduce the problem (1, 2, 3...), including links
1. Sign out of WebApp
2. Visit the ballot. You should see the IntroModal. Close it.
3. Sign In
4. While the sign in is being processed, the IntroModal will flash on screen. Then when the new voter is retrieved (voterRetrieve), we get the data from this in /js/routes/Ballot/Ballot.jsx:
VoterStore.getInterfaceFlagState(VoterConstants.BALLOT_INTRO_MODAL_SHOWN)
...which tells us that the signed in voter record has already seen the modal. So the modal quickly closes.

Possible Solution: 
During the sign in process, disable the setting of "this.state.showBallotIntroModal" while we are waiting for the response to voterRetrieve. 

- [x] Figure out a way around "Facebook login failed..." message
- [x] I also got a "Twitter sign in failed. Please try again." message that stopped the sign in process.
- [x] Once the signed in voter data is retrieve, redirect the browser to "/ballot" so that the url variable "?wait_until_voter_sign_in_completes=1" is removed.

DaleMcGrew:
Your changes work very well @jainanisha90!

PR #1119, opened 2017-09-29
#1106 Do not display Facebook login failed when successful and once sign in completes redirect to /ballot  #1112 Fixed remaining things when we_vote_branding_off is set

jainanisha90:


DaleMcGrew:
👍

Issue #1132, opened 2017-10-11
Ballot Location Buttons: Add "x more" button in mobile mode


DaleMcGrew:
### Please describe the issue (What happens? What do you expect?)
Currently we show 5-6 ballot location buttons in desktop. In mobile, we want to show 2 ballot location buttons and then a "x more" button that unfurls all of the other options. It would be ideal if these three buttons could fit on 1 line. This is what we currently show in mobile:
![img_4237](https://user-images.githubusercontent.com/7756031/31421953-7aeee83a-adff-11e7-9937-f567d013e109.jpeg)

The code that powers these buttons is here:
WebApp/src/js/components/Navigation/BallotLocationChoices.jsx

You can find an election with multiple ballot locations here:
https://wevote.us/ballot/arlington_virginia_november_2017

josh-southern:
Got a ok version of this working. Not sure I completely like the structure right now, but will push next week sometime regardless of what state it's in.

DaleMcGrew:
Thank you for letting me know Josh!

DaleMcGrew:
Resolved! Thank you @josh-southern.

PR #1161, opened 2017-11-03
Issu1132 b

josh-southern:
https://github.com/wevote/WebApp/issues/1132
Roughish version of this solution. Will work on it more next Wednesday.

### DIFF
--- a/src/js/components/Navigation/BallotLocationButton.jsx
+++ b/src/js/components/Navigation/BallotLocationButton.jsx
@@ -1,0 +1,54 @@
+import React, { PropTypes, Component } from "react";
+import { Button } from "react-bootstrap";
+import BallotStore from "../../stores/BallotStore";
+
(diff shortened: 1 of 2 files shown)


DaleMcGrew:
Hi Josh, I see what you are trying to do with the onMobile function. In the spirit of responsive design, I think there is a more elegant way to do this, specifically with the use of html that is displayed based on viewport size. For example:
{/* Desktop */}
<span className="hidden-xs">Remaining Decisions ({this.props.length_remaining})</span>
{/* Mobile */}
<span className="visible-xs-block">Decisions ({this.props.length_remaining})</span>

We also don't want to use logic like "this.state.on_mobile" -- instead we create the html for both cases and have the browser show or hide spans or divs based on the viewport size.

I'm happy to talk it through with you this coming Wednesday night. Thank you!

Issue #1168, opened 2017-11-08
Add scroll bar to Issue and Organization Modals


DaleMcGrew:
### Please describe the issue (What happens? What do you expect?)
In the Issues and Organization Modals, we would like a scroll bar to show up when in desktop mode. (Not necessary in mobile mode.) Voter testing showed that it wasn't obvious that this area could be scrolled through.

![screen shot 2017-11-08 at 7 24 54 am](https://user-images.githubusercontent.com/7756031/32557547-5800a12e-c457-11e7-8702-8725dd1357ee.png)

To get here:
1. Go to ballot
2. Click on Issues button:
![screen shot 2017-11-08 at 7 47 18 am](https://user-images.githubusercontent.com/7756031/32558235-264f216c-c459-11e7-96cf-d2f7304a9ea1.png)

DaleMcGrew:
Great work @josh-southern! I love how it works now in the Issues modal. Could you please add the changes that you made to the Issues modal over to the Organization modal?

josh-southern:
Hey fixed this bug, and applied it to the Organizations modal, but there's still quite a bit of extra space in the modals. I will likely tweak it more with breakpoints next week.

DaleMcGrew:
Ok thank you @josh-southern! If you'd like your current code merged with develop could you please create a pull request? Or you can do that next week after your additional changes.

DaleMcGrew:
Thank you @josh-southern!

PR #1177, opened 2017-11-16
webkit scrollbar fix

josh-southern:
https://github.com/wevote/WebApp/issues/1168
Preliminary fix for the webkit/mac osx default scroll bar not showing issue. Can't seem to position it over the div though. Would like some feedback.

### DIFF
--- a/src/sass/components/_intro-modal.scss
+++ b/src/sass/components/_intro-modal.scss
@@ -511,6 +511,30 @@
     .intro-modal-vertical-scroll-contain {
       height: calc(100vh - 350px);
       max-height: 640px; // Scroll area in modal: 990 - 350 = 640px We set this so that when We Vote is in a frame, the modal isn't too long.
+
+      &::-webkit-scrollbar {
+        width: 9px;
+      }
+
+      &::-webkit-scrollbar-track {
+        -webkit-border-radius:5px;
+        border-radius:5px;
+        background:rgba(255, 255, 255, 0.1);
+      }
+
+      &::-webkit-scrollbar-thumb {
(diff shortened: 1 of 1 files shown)


DaleMcGrew:
Thank you @josh-southern. @ixley, @pertrai1 or @edward-ly do either of you have any comments on Josh's approach?

ixley:
Trying to force a scrollbar, and only for select browsers seems like it overcomplicates the issue and might not be the ideal solution. Is there a reason this issue is unique to desktop browsers? That's a bit surprising since there is likely to be more screen space available. 

One alternative to trying to selectively force a scrollbar to appear would be to add a visible drop-shadow or gradient over the boxes near the bottom of the container. This is a visual cue that can help when it's not clear that there is more content is being cropped off.

pertrai1:
@josh-southern do you happen to have a screenshot of the issue that this is resolving?

josh-southern:
@pertrai1 this is a particularity of mac osx and webkit browsers. Basically scrollbars are not shown until the user attempts to scroll. The issue is it is not apparent to some users that the modal is scrollable at all since it's hidden initially.

josh-southern:
@ixley I like this idea more than fiddling with the native scrollbar like I did in this code. I'd like to get @DaleMcGrew's thoughts on taking this approach instead.

DaleMcGrew:
I just got to Code for SF and @josh-southern and I are sitting side-by-side now. He is going to look into the drop-shadow approach instead of modifying the Mac scroll bars. Thank you @ixley and @pertrai1!

Issue #1149, opened 2017-10-26
Modal Dialog style refactoring and bug fixing


DaleMcGrew:
### Please describe the issue (What happens? What do you expect?)
We need someone to take on some stylesheet clean up related to modal styling, sizing and positioning. Please be cautious about your changes, and test all of the scenarios below when you make changes. This is an advanced css issue.

The main problem is that we have modal styling that has not been restricted to specific modals. The goal is to attach all modal styling to specific modals, with clear naming and descriptions in the style sheets. 

To find the styles that affect modals, search for:
modal-dialog
modal-body
max-height
100vh

### Testing your changes
- [x] On the ballot page we have the Issues and Organizations modals. These are controlled by styles in:
WebApp/src/sass/components/_intro-modal.scss
intro-modal-vertical-scroll-contain
calc-height
modal-dialog
- [x] Click on any position bar on the ballot page
- [x] Click the "Edit" link to the right of your address on the ballot page
- [x] Click on the Gear icon to the right of the Election name
- [x] Clear your voter_device_id cookie and then click on the Thumbs Up or Thumbs Down button on the ballot. There is a modal that opens with an explanation the first time you support or oppose something.

DaleMcGrew:
Great work @edward-ly! Verified.

PR #1180, opened 2017-11-21
Refactored modal styling on ballot page

edward-ly:
Fixed class names of some modals for consistency. Refactored modal-dialog class declarations. Removed manual modal height declarations that were messing with some modals.

#1149


DaleMcGrew:
👍

Issue #1188, opened 2017-11-23
Hovering over some organizations displays two modals


DaleMcGrew:
### Please describe the issue (What happens? What do you expect?)
When you hover over some organizations, their are two informational modals shown:
![screenshot from 2017-11-17 17-20-18](https://user-images.githubusercontent.com/7756031/33190778-791c7756-d065-11e7-9123-40d7b12644b1.png)

DaleMcGrew:
Verified. Thank you @edward-ly.

PR #1192, opened 2017-11-30
Hovering over organizations displays correct overlay

edward-ly:
Fixed an issue causing both desktop and mobile overlays to appear at the same time when hovering over an organization (#1188). Fixed the styling of the overlay in `ItemTinyPositionBreakdownList.jsx`. Various ESLint fixes.


DaleMcGrew:
👍

PR #1193, opened 2017-11-30
div bar blur added to bottom of modal card.

josh-southern:
https://github.com/wevote/WebApp/pull/1177
Instead of the webkit scroll bar code I added in a cosmetic div that's blurred at the bottom.


DaleMcGrew:
👍

Issue #1176, opened 2017-11-16
Convert click on organization logo to "Follow" group instead of jumping to voter guide page


DaleMcGrew:
### Please describe the issue (What happens? What do you expect?)
1) Mobile, Following (groups you are already following): A click should open the mini-modal, a second click should close the mini-modal. In mobile we should center this modal left-right.
2) Mobile, More To Follow (groups you can follow): A click on the icon should open the mini-modal, a second click should close the mini-modal.
3) Desktop, Following: On rollover, the mini-modal should open (as it does now). A click should open or close the mini-modal. Roll-off should also close the mini-modal (as it does now).
4) Desktop, More to Follow: A click on the icon should open the mini-modal, a second click should close the mini-modal.

### Steps to reproduce the problem (1, 2, 3...), including links
1. Go to: http://localhost:3000/ballot/fredericksburg_virginia_november_2017
2. Click on Governor State of Virginia
3. Change the click on a logo (for items 2) and 4) above) to open the group modal.
![screen shot 2017-11-15 at 5 46 04 pm](https://user-images.githubusercontent.com/7756031/32869634-e4009708-ca2c-11e7-9fd6-5bf796f9b7aa.png)

The code for the groups you are following is in "ItemTinyPositionBreakdownList":
WebApp/src/js/components/Position/ItemTinyPositionBreakdownList.jsx

The code for the "More To Follow" icons is in: 
WebApp/src/js/components/Widgets/ItemSupportOpposeCheetah.jsx
Search for "organizations_to_follow_support_desktop" and trace it to "voterGuideLink" around line 170. You should be able to replace this with "OrganizationActions.organizationFollow"

DaleMcGrew:
Done!

PR #1195, opened 2017-12-05
Updated organization logo behavior on ballot page

edward-ly:
Converted the organization logos so that they, when clicked, follow the organization for those not currently following or open/close the mini-modal for organizations already followed (#1176). Various ES6 and style fixes.


DaleMcGrew:
👍

PR #1208, opened 2017-12-11
Upgraded from Flux 2.1.1 to 3.1.3, replaced all FluxMapStore with ReduceStore

SailingSteve:
Extracted all of the boilerplate out of About, Credits, and Team -- now
reads from people.js, which is also now common to WebApp and Native.
Bookmark store does a state.merge() which now requires Immutable.js
with Flux 3.1.3, so added facebook's Immutable.js
Removed package-lock.json from .gitignore, see
https://docs.npmjs.com/files/package-lock.json for the reason.


DaleMcGrew:
Hi @SailingSteve, this is some awesome work. Thank you for doing this critical upgrade! 👍

Issue #1199, opened 2017-12-07
Find your polling location


DaleMcGrew:
On election day, voters need to find their polling location in a hurry. In the weeks before an election, We Vote gets the address for each voter’s polling location from Google Civic. We would like to show the voter where their polling location is, and help them get there.

More details here:
https://docs.google.com/document/d/1PX6cF8mYRRC4rIN6SoL4SXH8GzlkkbnNtJ6fkK_dDAE/edit#

DaleMcGrew:
We will tackle this in 2019.

PR #1210, opened 2017-12-11
#1199 Added Polling Locator to secondary navigation bar to check for polling place location(in progress)

jainanisha90:


DaleMcGrew:
👍  Thank you!

PR #1216, opened 2017-12-20
disabled clicking next when there isn't at least one valid email in A…

chulinguy:
…ddFriendsByEmail

Add Friends by Email: Should require email address

- I tried a couple ways to fix this issue.  In the end I went with the one that had least amount of changes in the code ... I simply disabled the 'next' button when there isn't at least one valid email entered.  With this approach, I don't have to write additional code to make the "no valid email" warning message disappear when user enters a valid email BUT before user clicks the 'next' button.   However, if the warning message approach is preferable, please let me know as well.  

-Also added yarn.lock to .gitignore

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -176,3 +176,4 @@ web_app/build/*
 *.pyc
 staticfiles
 
+yarn.lock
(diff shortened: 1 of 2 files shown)


DaleMcGrew:
Thank you Chi Jen Lu! I'm not sure why the clahub warning is still happening even though you have approved it. 👍

PR #1219, opened 2017-12-21
height fixes with added gulp support for sourcemaps with css minifica…

josh-southern:
…tion via cssnano

https://github.com/wevote/WebApp/issues/1168
Small height fix in the tablet breakpoint. I also added in some support for sourcemaps with gulp. I had to comment out gulp-autoprefixer and replace the existing minifier with a gulp-sourcemaps compatible one. Is gulp-autoprefixer being actively used? It appeared to play no role with the current css/sass files.


DaleMcGrew:
Thank you @josh-southern! @pertrai1, would you be willing to take a look at this check-in since it involves gulp and source maps.

Issue #1203, opened 2017-12-09
Email Ballot to Friend: Ballot is not sent to "test2" friend after "test1" verifies email address


DaleMcGrew:
### Please describe the issue (What happens? What do you expect?)
After verifying the original email sent to test1@wevote.info, I did not receive the ballot link or friend request email that should have been sent to "test2@wevote.info"

### Steps to reproduce the problem (1, 2, 3...), including links
1. Cleared cookies
2. Went to ballot and chose Virginia ballot
3. Used Email Ballot modal, and sent ballot to self (test1@wevote.info)
4. Sent ballot to friend (test2@wevote.info)
5. test1@wevote.info received both emails: one with ballot link, and another "Please verify your email"
6. I clicked the Verify Email button and was correctly signed in.
7. I did not receive the email with the ballot link or friend request email at "test2@wevote.info"

DaleMcGrew:
I have verified that your most recent checkin fixed this! Great work @jainanisha90.

PR #1225, opened 2017-12-26
#1203 Email Ballot to Friend: Ballot is not sent to test2 friend after test1 verifies email address

jainanisha90:


src/js/actions/FriendActions.js
```diff
@@ -30,7 +30,7 @@ module.exports = {
   },
 
   emailBallotData: function ( email_address_array, first_name_array, last_name_array, email_addresses,
-                              invitation_message, ballot_link, sender_email_address) {
+                              invitation_message, ballot_link, sender_email_address, verification_email_sent) {
```
pertrai1: Might be easier to read if this is separated into an interface that is passed in.

src/js/actions/FriendActions.js
```diff
@@ -39,7 +39,8 @@ module.exports = {
         email_addresses_raw: email_addresses,
         invitation_message: invitation_message,
         ballot_link: ballot_link,
-        sender_email_address: sender_email_address
+        sender_email_address: sender_email_address,
```
pertrai1: ES6 allows for shorthand key value. Rather than:

`sender_email_address: sender_email_address`

you can do:

`sender_email_address`

That goes for all of the keys listed here. Might be best to name the args passed in the same as the keys so you can reduce this to just having to use the key name.

src/js/components/Ballot/EmailBallotModal.jsx
```diff
@@ -124,23 +124,28 @@ export default class EmailBallotModal extends Component {
...
     } else {
       success_message = <span>Success! This ballot has been sent to the email address {this.state.sender_email_address} </span>;
     }
 
     FriendActions.emailBallotData("", "", "", this.state.sender_email_address, this.state.email_ballot_message,
-      this.state.ballot_link, this.state.sender_email_address);
-    console.log("ballotEmailSend: email_address: ", this.state.sender_email_address);
+      this.state.ballot_link, this.state.sender_email_address, this.state.verification_email_sent);
+
+    if (!this.hasValidEmail()) {
```
pertrai1: Not sure I understand the naming convention used here. Is this for the checking of a proper email address or an email address that has not been verified yet? I can't determine the condition of the check based on `hasValidEmail` being false. If it is false, to me the name does not state clearly what is not valid about it.

src/js/components/Ballot/EmailBallotModal.jsx
```diff
@@ -124,23 +124,28 @@ export default class EmailBallotModal extends Component {
...
+    }
     // After calling the API, reset the form
     this.setState({
       loading: true,
       sender_email_address_error: false,
       on_enter_email_addresses_step: true,
-      on_collect_email_step: false,
       on_ballot_email_sent_step: true,
       showEmailToFriendsModal: false,
+      verification_email_sent: verification_email_sent,
```
pertrai1: can be changed to `verification_email_sent` shorthand

src/js/actions/FriendActions.js
DaleMcGrew: Thank you @pertrai1. What would be an example of separating this into an interface?

src/js/actions/FriendActions.js
DaleMcGrew: Great suggestion @pertrai1! We will start transitioning to this new shorthand.

src/js/components/Ballot/EmailBallotModal.jsx
DaleMcGrew: I will discuss with @jainanisha90 a better function name. Thank you.

DaleMcGrew:
👍

src/js/actions/FriendActions.js
pertrai1: @DaleMcGrew If there is some type of utility file of some sorts that has generic exports, you can have one such as:

Utilities.js (example name)
```
const userData = {
    email_address_array,
    first_name_array,
    etc...
}

export userData
```

FriendAction.js
```
import { userData } from 'Utilities'

emailBallotData({...userData} = {}) {
    // body here 
    // userData gives the the object that you can then use in the body
}
```

This is not an exact replica of what you have. Give it a shot or I can help if needed.

Issue #1231, opened 2018-01-03
"Add Friends by Email" fails despite valid email


robla:
### Please describe the issue (What happens? What do you expect?)
Adding a friend using /friends/invitebyemail throws an empty email address error despite providing valid email.

### Steps to reproduce the problem (1, 2, 3...), including links
1.  Go to https://wevote.us/friends/invitebyemail
2.  Fill in an email address (and a first+last name)
3.  Click "Send >"
4.  Expected behavior: a celebration of a valid email  :-)
     Actual: An error: "Please enter at least one email address."

chulinguy:
Hi robla, 

I believe I accidentally introduced this bug in one of my commits; I just created a pull request to fix this.  Thanks for letting me know! 

Chi

DaleMcGrew:
This particular issue is now fixed.

PR #1232, opened 2018-01-03
Fix for #1231: fix up setter for email_address_array

robla:
Issue #1231 

Prior to this fix, the app sets this.email_address_array, but then checks for the value in this.state.email_address_array.  Setting them both in the setEmailAddressArray setter works around the problem, but it's possible that this.email_address_array also needs to go away altogether.

### DIFF
--- a/src/js/components/Friends/AddFriendsByEmail.jsx
+++ b/src/js/components/Friends/AddFriendsByEmail.jsx
@@ -278,15 +278,19 @@ export default class AddFriendsByEmail extends Component {
   }
 
   setEmailAddressArray (value) {
+    // do we need both this.email... and this.state.email...?
     this.email_address_array = value;
+    this.state.email_address_array = value;
   }
 
   setFirstNameArray (value) {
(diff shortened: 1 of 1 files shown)


DaleMcGrew:
Thank you Rob! Will you be at Code for SF tonight? I'd love to talk through this.

robla:
> Will you be at Code for SF tonight?

Yup, I will

chulinguy:
Hi Rob, 

Thanks for spotting this. I think whoever originally created AddFriendsByEmail had duplicate arrays (one as a key and one as a state)to track emails/firstname/lastname/etc to circumvent the fact the asynchronous nature of this.setState().  

React Docs mentioned to never mutate state directly ... https://stackoverflow.com/questions/37755997/why-cant-i-directly-modify-a-components-state-really.  

I think it's possible to remove the non-state arrays but it would probably require significant rewriting.

DaleMcGrew:
Thank you @robla for finding this bug and working with me on a fix.

PR #1235, opened 2018-01-04
#621 Display election search results with election name, election date and in past in the search results drop-down

jainanisha90:


DaleMcGrew:
👍

Issue #1207, opened 2017-12-09
Issues & Org Modals: scrolling issues box too short in mobile mode


DaleMcGrew:
### Please describe the issue (What happens? What do you expect?)
On iPhone 6 Safari, the screen area where issues photos show is too short and could be taller. The same problem exists on Organization modal.
![img_4388](https://user-images.githubusercontent.com/7756031/33792866-40661d7a-dc60-11e7-8513-dac4480fea2a.jpeg)

DaleMcGrew:
Resolved. Thank you @josh-southern.

PR #1237, opened 2018-01-04
added shadow to org modals and adjusted height

josh-southern:
https://github.com/wevote/WebApp/issues/1207
Small changes to org component for modal shadows and some SASS changes.

### DIFF
--- a/src/sass/components/_intro-modal.scss
+++ b/src/sass/components/_intro-modal.scss
@@ -434,6 +434,10 @@
   -webkit-overflow-scrolling: auto;
   overflow-x: hidden;
 
+  &.advisers {
+    height: calc(100vh - 345px);
+  }
+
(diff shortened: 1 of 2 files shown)


src/sass/components/_intro-modal.scss
```diff
@@ -434,6 +434,10 @@
   -webkit-overflow-scrolling: auto;
   overflow-x: hidden;
 
+  &.advisers {
```
DaleMcGrew: Hi @josh-southern, could you take a look at the BEM naming conventions, linked to from here? https://github.com/wevote/WebApp/blob/develop/docs/working/STYLING.md

I would recommend naming this &__advisers (instead of &.advisers) so that the class you would use would be: intro-modal-vertical-scroll-contain__advisers (instead of "advisers")

PR #1256, opened 2018-01-12
Steve web app jan12

SailingSteve:
(Did not change the actual data, just the boilerplate)
Removed the wrapper function that could not do the job in Native.
Removed module.exports in favor of ES6 exports.
This is an unimportant change to prove that my Mac is configured so
that node-sass doesn't prevent git commits.


DaleMcGrew:
👍

PR #1257, opened 2018-01-12
Updated candidate positions labels on ballot page

edward-ly:
Fixed a bug where the "Score in Your Network" label is not in the right position on mobile. Added a similar "Positions" label to the left of each (non-empty) list of organizations.


DaleMcGrew:
👍

Issue #1249, opened 2018-01-08
Add left/right scroll to the ItemSupportOpposeRacoon component


DaleMcGrew:
### Please describe the issue (What happens? What do you expect?)
We can have many voter guides for any particular ballot item. Currently we limit the number of voter guide icons shown. Update WebApp/src/js/components/Widgets/ItemSupportOpposeRaccoon.jsx to allow the voter to scroll to see additional voter guide icons.
![screen shot 2018-01-08 at 2 33 11 pm](https://user-images.githubusercontent.com/7756031/34695751-e3c5e92c-f480-11e7-93d8-133b3f791abf.png)

We also want to add an arrow on the right side of the scrollable area if there are voter guide icons that will stretch off the screen.

Later, we will refactor "maximum_organizations_to_show_desktop" and remove it, but lets not do that yet.

DaleMcGrew:
Working!

PR #1276, opened 2018-01-19
Implemented left/right scrolling to ItemSupportOpposeRacoon

edward-ly:
Implemented the first version of the new list of organization icons that is entirely contained on one line and supports left/right scrolling on mobile and desktop (#1249). Includes the first draft of an implementation for scrolling through the organizations list with button clicks.


DaleMcGrew:
👍

PR #1278, opened 2018-01-19
Initial changes to allow the WebApp to run in an Apache Cordova container

SailingSteve:
For now, there are two index.html files, the pre-existing one for WebApp
alone, and the new one in /www for WeVoteCordova.
Removed line length lint warning from .jscrc


DaleMcGrew:
👍 Thank you @SailingSteve, we will need to update the license to MIT from the current one specified in the index file.

PR #1284, opened 2018-01-23
The We Vote WebApp is about 2/3 working in Apache Cordova (b)

SailingSteve:
Files are symlinked between the two We Vote git projects, so there is
almost no customised JavaScript on the WeVoteCordova side, it is all in
the WebApp. The entry point for Cordova is in the new www directory.
The WeVoteCordova side just contains the iOS and Android
specific code and the Cordova code.
There were different paths for images, so a wrapper "cordovaDot" is now
wrapping every image path in the WebApp -- when working on the WebApp
the wrapper does nothing, when in Cordova it adds a '.' to the beginning
of the path. The cordovaDot function is in /utils/cordovaUtils.
The WebApp widely uses browserHistory.push() from "react-router", but
Cordova requires hashHistory.push() from "react-router" so every
instance of browserHistory.push() has been replaced with a wrapper
called historyPush also in /utils/cordovaUtils.
There is a new configuration item in config.js, called IS_CORDOVA which
defaults to false, for the default usage in the WebApp... you will need
to add this to your personal config.js
I have also been clearing JSCS warnings, when I make nearby changes.


DaleMcGrew:
Hi Steve, thank you for all of these changes! I am excited by your progress. 👍

PR #1285, opened 2018-01-23
Lint errors resolved, most warnings resolved from prior pull request

SailingSteve:
Minor fix to text in Your Network Supports popover
and rewrote a comment in index.js
Cleared almost all of the lint warnings and all of the errors.
Clearedmany JSCS warnings in the same files.
JSCS disallowSpaceInFunctionDeclarations (which I like), conflicts with
lint space-before-function-paren, so I left the function defs that had
space-before-function-paren warnings in place.


DaleMcGrew:
Thank you @SailingSteve! 👍

Issue #1241, opened 2018-01-04
Mobile: When thumbs up is toggled on, add "pop down" notification


DaleMcGrew:
### Please describe the issue (What happens? What do you expect?)
Currently in mobile we show overlays when you click the thumbs up or thumbs down. We want to remove these overlays, and instead show a "pop down" notification (research needed to find the best package) with similar text that shows *after* the click. The behavior in desktop should not change.

### Steps to reproduce the problem (1, 2, 3...), including links
1. Go to the ballot page https://WeVote.US/ballot on your mobile device
2. Click on one office until you see candidates
3. Click on the thumbs up, and you will see that the overlay shows. You have to click again to make the thumbs up select. This should be changed so that clicking on the thumbs up causes a "pop down" notification at the top of the screen, and no overlay next to the thumbs up.

WebApp/src/js/components/Widgets/ItemActionBar.jsx
We have "supportButtonPopoverTooltip" which should show in desktop. 
- [x] Don't show "supportButtonPopoverTooltip" (or "opposeButtonPopoverTooltip") in mobile.
- [ ] Add a new "pop down" notification (from the top of the screen) that shows in mobile, but not desktop

uncleoptimus:
Would this toast library do the trick?
https://fkhadra.github.io/react-toastify/

(or something similar)

DaleMcGrew:
@uncleoptimus, I love the package you found! https://fkhadra.github.io/react-toastify/ Thank you for doing the research.

uncleoptimus:
Cool, looks like we can just use the Bootstrap responsive classes to hide the tooltips on mobile 👍 

Couple quik thoughts:

- Should I go ahead and bump up the Bootstrap version to the new official 4.0 release? Btwn v3.3 and official 4.0 there have been like three different naming conventions for the responsive classes (including differences btwn the version currently used on dev branch and the new release). Downside: may break things if other folks have been using stuff from 4.0-beta that has been altered
 
- `package.json`: seems we got some dev packages listed under "dependencies" and vice-versa. Can shift stuff around a bit there if desired

uncleoptimus:
Cool @edward-ly @DaleMcGrew , should still I go ahead and pull down these latest changes and proceed with the toast feature (only on mobile)?

Also, do we already have a programmatic way to determine if the User is on mobile? Else, I added a lil util function based on:
https://stackoverflow.com/a/3540295

Reference screen w/ default styling:
[...]

PR #1286, opened 2018-01-23
Implemented click to scroll organizations list, hide support/oppose button tooltips on mobile

edward-ly:
Implemented the ability to scroll through the organizations lists for each candidate by clicking on the left/right arrows to the right of each list (#1249). The ability to hide these arrows if the list is not scrollable, or hide one at a time if the list cannot scroll further to the left or right, will be a separate issue.

In addition, the supportButtonPopoverTooltip and opposeButtonPopoverTooltip are no longer shown on mobile (step 1 of #1241).


DaleMcGrew:
👍

PR #1287, opened 2018-01-24
Make goToBallotLink() static to avoid unbound function error in two classes.

SailingSteve:
URLs in css/sass are problematic for Cordova, since CSS loads assets
relative its location and js loads assets relative to the HTML location.
So condtionally overrode the css image locations with style settings in
the IntroNetwork* classes and in Welcome.jsx


DaleMcGrew:
👍 Thank you for fixing this @SailingSteve!

PR #1295, opened 2018-01-31
Implemented hiding of left/right scroll buttons based on scroll position

edward-ly:
Implemented the ability to hide the left scroll button if the corresponding organization list cannot scroll any farther left, and similarly for the right scroll button. Moved the left scroll button to the left of each organization list, and updated the button margins to account for this (#1249). Organization list classes have also been renamed to follow BEM naming.


DaleMcGrew:
👍  I love these changes!

PR #1308, opened 2018-02-07
Twitter SignIn working in Cordova

SailingSteve:
Mostly due to new pushHistory in Application.jsx
Lint changes
Still has a minor problem where the logged in icon, does not show up right away.


DaleMcGrew:
Thank you @SailingSteve! 👍

Issue #1305, opened 2018-02-02
Make the comment area on the ballot page wider


DaleMcGrew:
### Please describe the issue (What happens? What do you expect?)
The comment component(s) (With "Why you support...") are narrower than they should be on the ballot.
![screen shot 2018-02-01 at 10 35 54 pm](https://user-images.githubusercontent.com/7756031/35720426-89532e16-07a2-11e8-9afd-1911b1080543.png)

### Steps to reproduce the problem (1, 2, 3...), including links
Look for "comment_display_raccoon_desktop" and "comment_display_raccoon_mobile" within this component:
WebApp/src/js/components/Widgets/ItemSupportOpposeRaccoon.jsx

DaleMcGrew:
Awesome solution @edward-ly! Thank you.

PR #1312, opened 2018-02-08
Changed comment area layout on ballot page

edward-ly:
Removed the extra space to the left of the comment area (#1305). Changed the comment area layout by removing the support/oppose text, moving the public position toggle and post button to the right of the comment area, and changed the size of the post button to align with the toggle.


DaleMcGrew:
👍

Issue #1292, opened 2018-01-30
Create "Register to Vote" page using Vote.org widget


DaleMcGrew:
### Please describe the issue (What happens? What do you expect?)
We want to be able to point voters to a "Register to Vote" page on WeVote.US, that uses the Vote.org registration widget.

DaleMcGrew:
Verified. Thank you @edward-ly!

PR #1314, opened 2018-02-10
Added "Register to Vote" page and more from Vote.org widgets

edward-ly:
Added three additional routes created from three of the four available free tools from [Vote.org](https://www.vote.org/technology/) (#1292). Used the `ReactIframeResizer` component from the `react-iframe-resizer-super` package to automatically resize the height of the iframe based on the height of the widget contents.


DaleMcGrew:
👍 Great work work @edward-ly!

PR #1319, opened 2018-02-14
Suppressed Facebook signin for WeVoteCordova

SailingSteve:
Will backout suppression within a month, all the relevant changes are
tagged with comments.
Cleared some lint warnings.


DaleMcGrew:
👍

PR #1331, opened 2018-02-20
Partial fix #1247 Change to Another Ballot page to show a loading message when switching from an upcoming election to different upcoming election

### LINKED ISSUE #1247: Update the "Change to Another Ballot" page to show a loading message when switching to different election
Currently when you click on one of the bit green buttons you are taken immediately to the ballot page, even before the new ballot information is received. Instead, we want to stay in the modal with a "loading" message until the new ballot has loaded, and then go to the ballot page.
screen shot 2018-01-05 at 3 28 41 pm

This work is in WebApp/src/js/components/Ballot/BallotElectionList.jsx, and we don’t want to close the Modal with this.props.toggleFunction until the new data has loaded.

So we want to add a “Switching ballot data now…” message within the modal, and then when the new election data has arrived (after calling voterBallotItemsRetrieve), we can then close the modal.

This is related to: https://github.com/wevote/WebApp/issues/1088

jainanisha90:


DaleMcGrew:
👍

Issue #1318, opened 2018-02-14
"Opinions about..." overlap of organizational icons


DaleMcGrew:
### Please describe the issue (What happens? What do you expect?)
When showing Opinions to Listen to, we start to see an overlap of organizational icons:
![screen shot 2018-02-13 at 4 44 45 pm](https://user-images.githubusercontent.com/7756031/36183737-8ed7dc6a-10e4-11e8-9822-269b24701ce8.png)

See here: https://wevote.us/ballot/id/wv02ballot2743

DaleMcGrew:
Thank you @edward-ly, this is now solved.

PR #1332, opened 2018-02-21
Fixed overlap of organizational icons

edward-ly:
Fixed an issue that caused some sub-lists of organization icons to overlap another (#1318). Also rounded the width value of the scrollbar so that the scroll buttons are properly shown or hidden.


DaleMcGrew:
👍

PR #1340, opened 2018-02-26
Implemented scrolling behavior for issues list, fixed issue and network score display

edward-ly:
Copied the scrolling behavior of the organizations list into the icons list. Fixed a bug where the issue score and network score are not displayed on the same row. Implemented hiding the issue score on mobile to show the network score instead (if there is one).


DaleMcGrew:
👍

PR #1343, opened 2018-02-27
Cordova changes 1 through 6: new footer bar for Cordova

SailingSteve:
New FooterBarCordova.jsx
Removed some items from HeaderBar and HeaderBarProfilePopUp if Cordova
In Welcome.jsx, changed the text of a label since donation option is not available.
In cordovaUtils.js, added enclosingRectangle logger, since it is harder to find
objects on screen in Cordova while debugging.
In Application.js, display the new footer bar if Cordova.
In _navigator.scss, various changes for the new footer bar and the new
location of the HeaderBarProfilePopUp (bottom of the screen in Cordova).


DaleMcGrew:
👍  Thank you @SailingSteve, and welcome back!

Issue #1317, opened 2018-02-13
We are ready to update our "Free Tools" page


DaleMcGrew:
### Please describe the issue (What happens? What do you expect?)
We want to update our "Free Tools" page to be less intimidating with this improved design.
![toolspage_v7 1](https://user-images.githubusercontent.com/7756031/36180457-60141682-10d5-11e8-8fc5-e25ad205d509.png)

### Steps to reproduce the problem (1, 2, 3...), including links
You can get to this page by clicking the "Tools For Your Website" link at the bottom of the welcome page.

The code is here:
WebApp/src/js/routes/More/ToolsToShareOnOtherWebsites.jsx

DaleMcGrew:
Hi @edward-ly, I updated the live server with this API (and checked in WebApp code too): https://api.wevoteusa.org/apis/v1/docs/organizationSearch/#tryitnow
If you enter "aclu" in organization_twitter_handle and set exact_match to True, you will get a unique result confirming we have their organization in our database.

I also added this action in WebApp you can use OrganizationActions.organizationSearch("", twitter_handle, True).

You can call OrganizationStore.getOrganizationSearchResultsTwitterHandle() to see if you get the same organization_twitter_handle back as a result. (Make sure to do a case insensitive compare, and this WebApp code hasn't been tested.)

DaleMcGrew:
Excellent work!

PR #1346, opened 2018-02-28
More Updates to Tools Page

edward-ly:
Added more changes to the free tools page: down-scaled all images to decrease data usage, added a loading status message and screenshot for the voter guide tool, and changed most images and links to link to example uses (#1317).


DaleMcGrew:
👍

Issue #1273, opened 2018-01-18
Voter Guide Margins (Card display) Needs Visual Cleanup


DaleMcGrew:
### Please describe the issue (What happens? What do you expect?)
In mobile, the cards should stretch to the edges of the screen:
<img width="335" alt="screen shot 2018-01-17 at 7 01 05 pm" src="https://user-images.githubusercontent.com/7756031/35078821-f939bdb2-fbb8-11e7-863f-1ab5a19ed1d7.png">

In desktop there should be some space above the first candidate card. The candidate card should stretch to the same width as the section above:
<img width="582" alt="screen shot 2018-01-17 at 7 04 42 pm" src="https://user-images.githubusercontent.com/7756031/35078879-53880c10-fbb9-11e7-83f4-6f47b429a1c5.png">

KatieStarks:
There's also a missing drop shadow on the ballot header.

DaleMcGrew:
Fixed!

PR #1349, opened 2018-03-01
Fixed Header Margins for Voter Guide

edward-ly:
Updated the header card in the voter guide to be declared as a card rather than enclosed within an additional Bootstrap row (#1273). Additional style fixes.


DaleMcGrew:
👍  This looks much better. :)

PR #1355, opened 2018-03-01
Cordova: Prevent our text from showing under the top line iOS menu

SailingSteve:
The top line of the phone with ATT, the time, and battery status is now protected


DaleMcGrew:
That's awesome! I'm glad you figured out a fix @SailingSteve. 👍

Issue #1353, opened 2018-03-01
Please make the "Issue Score" popover wider, and limit organization name length


DaleMcGrew:
### Please describe the issue (What happens? What do you expect?)
- [x] When a voter clicks on the Issue Score popover, please make the box that opens the same width as the organization popovers.
![img_4595](https://user-images.githubusercontent.com/7756031/36857285-0029c23a-1d2d-11e8-9505-9956b25b5cf9.jpeg)

- [x] Also, when an organization's name is to long to fits in the overlay box on one line, please wrap the text so that the second line of text starts to the right of the thumbs up/thumbs down icon. (So the thumbs up/thumbs down icons are treated a little like bullets.
![screen shot 2018-03-01 at 8 46 55 am](https://user-images.githubusercontent.com/7756031/36857550-a8b8998a-1d2d-11e8-8c14-d72025a76ca4.png)

DaleMcGrew:
Excellent work! Verified.

PR #1357, opened 2018-03-02
Increased "Issue Score" popover width, wrapped organization name to next line

edward-ly:
Fixed an issue where the organization name in the issue score popover does not wrap to the next line. Increased the width of the popover for both desktop and mobile, and the popover now extends across the screen properly on mobile (#1353). Renamed `DonateActions.jsx` to `DonateActions.js` for consistency.


DaleMcGrew:
This is some very nice code @edward-ly. 👍

Issue #1352, opened 2018-03-01
Search results to full width in mobile mode


DaleMcGrew:
### Please describe the issue (What happens? What do you expect?)
When searching in mobile mode, please make the search results drop down stretch the entire width of the view area, instead of being limited to the width of the search box.
![img_4594](https://user-images.githubusercontent.com/7756031/36857067-70809a5a-1d2c-11e8-8ffd-e2f3c6394b63.jpeg)

DaleMcGrew:
Nailed it!

PR #1358, opened 2018-03-02
Search results now full width in mobile

edward-ly:
Added a breakpoint to the "search-container" class so that search results are now full width in mobile (#1352). Updated styling to `SearchAllBox.jsx`.


DaleMcGrew:
👍

Issue #1363, opened 2018-03-02
Fix the "jump" as you scroll right on Issues and Opinions


DaleMcGrew:
### Please describe the issue (What happens? What do you expect?)
When you click the right arrow the first time, there is a visible jump when the left arrow gets displayed. Perhaps we can get rid of this by displaying the left arrow first as the issue images are starting to move?

If you keep your eye on one of the logos after they finish scrolling, the position of the logo also jumps when the left arrow appears. (May not be visible in this animated gif.)
![mthptil6gl](https://user-images.githubusercontent.com/7756031/36918325-69475804-1e0e-11e8-9d33-d90a09210f1d.gif)

DaleMcGrew:
Fixed. Beautiful work @edward-ly!

PR #1364, opened 2018-03-03
Fixed issues and opinions list scroll behavior

edward-ly:
Fixed the "jump" when scrolling through the issues and organizations lists by keeping the viewable width of each list constant and making the scroll buttons invisible when necessary rather than completely hiding them (#1363). Updated the algorithm for determining whether or not a list is scrollable at all.


DaleMcGrew:
I can't wait to play with it! 👍

PR #1372, opened 2018-03-03
Cordova: Multiple fixes and UI changes

SailingSteve:
The MP4 viewer on the About page now works in Cordova
The doubled search menu on GettingStarted is resolved.
    wevote/WeVoteCordova/issues/19
The WeVote Alpha text is removed when in Cordova.
    wevote/WeVoteCordova/issues/20
The Top line "ATT Time Battery" on iPhones now has $page-header-color
background color, white text still to be addressed.
Our header in Cordova, now only consists of a centered search box.
Added some space between candidate name, and candidate description on the card.
Moved the 'X' on the IntroNetwork down and away from the battery indicator if Cordova.


DaleMcGrew:
👍

Issue #1360, opened 2018-03-02
Improve look of election search results


DaleMcGrew:
### Please describe the issue (What happens? What do you expect?)
Perhaps place the election date and "Upcoming Election" on another line, and make the font smaller?
![img_4597](https://user-images.githubusercontent.com/7756031/36907170-beeac236-1dec-11e8-9f5a-22dd32eecd08.jpeg)

DaleMcGrew:
Great work @edward-ly!

PR #1377, opened 2018-03-07
Updated election search results layout

edward-ly:
Updated the layout of each election search result so that the date and "Upcoming Election" remain on the right side and on separate lines, and changed the look of "Upcoming Election" (#1360).


DaleMcGrew:
👍

Issue #1367, opened 2018-03-03
Add "No search results found." display in search dropdown


DaleMcGrew:
### Please describe the issue (What happens? What do you expect?)
If you search for a word that doesn't have any search results, please add the search dropdown with a line that says:
Nothing found for "Boxer".

![fullsizerender](https://user-images.githubusercontent.com/7756031/36937205-ab5a3a5c-1ec4-11e8-8a5c-86d31030dd62.jpg)

DaleMcGrew:
Great solution @edward-ly!

PR #1385, opened 2018-03-08
Added message in search dropdown if no search results

edward-ly:
Added a "Nothing found for {result}." message to the search dropdown when there are no search results (#1367).


DaleMcGrew:
👍

PR #1389, opened 2018-03-12
Intermediate checkin on "slide in" footer profile menu for cordova

SailingSteve:
emailBallotData endpoint now sends a deviceType string so that we can
make the emails have the correct url format for iOS and Android.
HeaderBarProfileSlideIn was refactored to eliminate tons of repetitive
code, now calls the new HeaderBarProfileSlideInRow.
In cordovaUtils, added a new deviceTypeString() that enumerates if
cordova and if ios/andoroid in a string that can be sent to the server.


DaleMcGrew:
👍

PR #1403, opened 2018-03-19
Type error voter guide ballot

uncleoptimus:
Addresses issue #1320 where a type error would show up in console

This pr copies over the missing update function and state var from `Ballot` to the `VoterGuideBallot` component.
Opening/closing a ballot item in either view will now persist between views


DaleMcGrew:
👍

Issue #1354, opened 2018-03-01
Show more Issues on Network page and Ballot modal


DaleMcGrew:
### Please describe the issue (What happens? What do you expect?)
THIS ISSUE Updated Saturday, March 3rd, 2018
- [x] In mobile mode, on the Network page, keep the number of issues at three wide, but increase the font size of the Issue names. Please also reduce the left and right margins so the images can be slightly larger:
![screen shot 2018-03-01 at 9 42 45 am](https://user-images.githubusercontent.com/7756031/36860230-8d55d830-1d35-11e8-9590-06c595771e0b.png)

- [x] In desktop mode, show four issues wide on the Network page (instead of the current 6). Please also reduce the left and right margins so the images can be slightly larger:
![screen shot 2018-03-01 at 9 43 12 am](https://user-images.githubusercontent.com/7756031/36860222-87f5ae24-1d35-11e8-99e5-3d6ceb269fb9.png)

- [x] In mobile mode, keep three issues wide in the Ballot modal, but increase the font size of the Issue names. In desktop mode, the number of issues per row should remain at 6 per row in the ballot modal:
![screen shot 2018-03-01 at 9 46 31 am](https://user-images.githubusercontent.com/7756031/36860211-8375b6e6-1d35-11e8-8b90-d6b05a631912.png)

- [x] I forgot to mention the outside margins on this Network page in mobile mode: https://wevote.us/issues_followed
<img width="350" alt="screen shot 2018-03-16 at 11 58 13 am" src="https://user-images.githubusercontent.com/7756031/37539618-6ce7c1ac-2911-11e8-9df4-08fe47d8f7ea.png">

- [x] And the outside margins on this network page in mobile mode: https://wevote.us/issues_to_follow
<img width="365" alt="screen shot 2018-03-16 at 11 58 29 am" src="https://user-images.githubusercontent.com/7756031/37539614-6692e82c-2911-11e8-9ddc-813e66972f65.png">

DaleMcGrew:
Awesome work @edward-ly! I forgot to mention this one last Network page:
https://wevote.us/issues_followed

DaleMcGrew:
Excellent work!

DaleMcGrew:
@edward-ly I figured out where I saw the problem -- in mobile mode. I updated the description of this issue with screen shots.

DaleMcGrew:
Solved!

PR #1404, opened 2018-03-19
Removed extra padding in network issues pages

edward-ly:
In mobile view, removed the excess left/right padding to the issues lists in the Issues Followed and Issues To Follow pages by removing the surrounding `card-child__list-group` div elements (#1354).


DaleMcGrew:
👍

PR #1409, opened 2018-03-23
Fix to dropdown size, from the SearchAllBox in Cordova, and more

SailingSteve:
Fixes the most recent problem in wevote/WeVoteCordova#32

Dale's suggestion to add "margin-bottom" of $space-xl to the
"BallotList" class (VoterGuideBallot, Ballot, and _ballotList.scss) resolves wevote/WeVoteCordova#26

In .jscrcs, added `"disallowSpacesInFunctionDeclaration": null`
to remove lint error indicator, for example, where
`function startApp () {` had been flagged for the space between "startApp" and "()".  This change
will take effect for all `*.js` files, but was not an issue in `*.jsx` files.

### DIFF
--- a/.jscsrc
+++ b/.jscsrc
@@ -4,4 +4,5 @@
     "excludeFiles": ["build/**", "node_modules"],
     "maximumLineLength": null,
     "maxErrors": 200,
+    "disallowSpacesInFunctionDeclaration": null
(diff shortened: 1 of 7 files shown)


DaleMcGrew:
👍

Issue #1366, opened 2018-03-03
Ballot modal on mobile needs fine-tuning


DaleMcGrew:
### Please describe the issue (What happens? What do you expect?)
- [x] Dragging the area with the issues is choppy -- is there anything we can do to smooth that out?
- [x] Increase the height of the area in which the issue photos display. (It has been configured the way it currently is to work on both mobile and desktop, but is too short on mobile.)
- [ ] When you reach the bottom of the Issues list and try to drag the issues up, the ballot behind the Issues modal drags up, and is distracting and disorienting.

![screen shot 2018-03-03 at 8 29 29 am](https://user-images.githubusercontent.com/7756031/36936717-c52e1e82-1ebd-11e8-97d2-4ee1685b8b1e.png)

DaleMcGrew:
Great work @edward-ly! I moved the body scrolling to a new issue.

PR #1411, opened 2018-03-24
Updated ballot issues modal layout and swipe behavior

edward-ly:
Fixed the responsive layout of the issues modal on the ballot page by using flexbox over calculated heights. Fixed the choppy scroll behavior of the issues list by disabling the ability to swipe left/right in the modal (#1366). Removed the text in issues modal suggesting that the next screen leads to the organizations modal.


DaleMcGrew:
👍

PR #1417, opened 2018-03-25
Refactor to make debugging Cordova more manageable

SailingSteve:
*** Added new constants to config.js (and template). ***

Added a renderLog() line after each render () call for debugging, it is
controlled by new constants in config.js
Copied logging.js from WeVoteReactNative.
Eliminated almost all the require() in favor of ES6 imports.
Made web_app_config into webAppConfig everywhere, and now we import it.
Cleaned up many eslint warnings.
In ItemPostioinStatementActionBar, resolved the garbled characters
portion of wevote/WeVoteCordova/issues/35 by replacing an inline utf8
char with an escaped unicode value.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1751,7 +1751,7 @@
     "base64-js": {
       "version": "1.2.1",
       "resolved": "https://registry.npmjs.org/base64-js/-/base64-js-1.2.1.tgz",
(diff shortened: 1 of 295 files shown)


DaleMcGrew:
👍

Issue #1262, opened 2018-01-14
Signing out should clear data out of FriendStore


DaleMcGrew:
### Please describe the issue (What happens? What do you expect?)
If you have friends attached to your account and then you sign out, the data in FriendStore is not cleared out, so it looks like you have friends when you aren't signed in.

All of the stores to clear out upon sign out:
- [ ] BallotStore
- [ ] BookmarkStore
- [ ] DonateStore
- [ ] FacebookStore
- [ ] FriendStore
- [ ] IssueStore
- [ ] OrganizationStore
- [ ] SupportStore
- [ ] TwitterStore
- [ ] VoterSessionStore
- [ ] VoterStore

chulinguy:
I forgot to ask about this ... are there other stores that you'd like to clear on signing out?  
My best guess would be to also clear VoteGuideStore, VoterSessionStore, VoterStore, TwitterStore, and FacebookStore.  It seems to be a good idea to keep the other stores to make the app load faster for the next user?

DaleMcGrew:
Thank you @chulinguy, I just updated this issue with all of the stores to clear out.

chulinguy:
@DaleMcGrew thanks! Follow-up question:

1) it seems to be that when the voter does a Facebook or Twitter sign out, the app will then load voterRetrieve() so the voter is technically still signed in on WeVote with email instead of Facebook/Twitter?
Is that the right behavior or should any signing out completely signs out the user?  

2) the VoterSessionStore does not seem to be storing any meaningful data? What is the intent to try to sign out with VoterSessionAction if VoterSessionStore is not necessary?

DaleMcGrew:
Hi @chulinguy, 
1. When we get a response from "voterSignOut" it triggers action in VoterStore, and FacebookStore. The action in VoterStore just sets up a new user, which I'm realizing I need to look at. Maybe the answer is to create an entry in every store that responds to "voterSignOut" by clearing the store variables, like we do partially do in FacebookStore.
2. You are correct VoterSessionStore isn't storing data, so I have removed that from the list.

PR #1418, opened 2018-03-26
Issue #1262: Signing out should clear data out of Flux Stores

chulinguy:
-Issue #1262: Signing out should clear data out of Flux Stores
-added a resetState method to each of the Flux store that needs to be cleared out
-each of those Flux store responses to a "voterSignOut" dispatch call by calling its own resetState() 
     - Decoupled calling additional actions needed to create a new user from just voterStore into individual stores (bookmark, friend, ballot)


DaleMcGrew:
Thank you Chi! 👍 

And thank you for your note about voterMergeTwoAccounts. I will work on decoupling the calls it makes, but I want to wait to do that when I can test more thoroughly.

PR #1420, opened 2018-03-27
Scrolling when Cordova keyboard opens after text entry, works acceptably

SailingSteve:
With some compromises, fixes wevote/WeVoteCordova/issues/35
Added bottom padding in various places, seems ok, much better than
before.

It is a little wonky for the address entry, since the address suggestions start up before 
the app has a chance to scroll out of the way of the keyboard for the first time.  But it works.


DaleMcGrew:
👍

PR #1421, opened 2018-03-27
Disabled the Choose via Facebook button in Cordova

SailingSteve:
Since it requires a Facebook sign-in, which we haven't implemented yet
in Cordova.


DaleMcGrew:
👍

PR #1422, opened 2018-03-27
Fix required parameter errors in WebApp and Cordova

SailingSteve:
Marked 5 parameters as not required.
As we use EditAddress today, half of the parameters are not used,
and not needed in EditAddress or EditAddressPopover.  Same with dozens
of lines of code that process the parameters that we don't pass.
Added a TODO indicating the abandoned code, which can be removed if we
have permanently gone a different direction.


DaleMcGrew:
👍

Issue #1361, opened 2018-03-02
Increase font size of "Related Issues" and "Opinions about..." on mobile


DaleMcGrew:
### Please describe the issue (What happens? What do you expect?)
On mobile (not desktop) I would like "Related Issues" and "Opinions about Mark" to be broken onto two lines and the text made larger so that the images can start further to the left. Please make sure to center vertically the "i" icon between the two lines.
![img_4598](https://user-images.githubusercontent.com/7756031/36917130-d767d8d0-1e0a-11e8-8d36-de5b69742d2f.jpeg)

DaleMcGrew:
Verified! Great work as always @edward-ly.

PR #1431, opened 2018-03-30
Condensed support label text on ballot page

edward-ly:
Modified the issues and opinions support label on the ballot page by separating it into two lines (on both desktop and mobile) (#1361).


DaleMcGrew:
👍

PR #1440, opened 2018-04-06
WebApp changes for Cordova Android -- oAuth through Twitter now works

SailingSteve:
Added InAppBrowser support for the oAuth redirect scheme, this relies on
a PR that I made to Cordova to allow schemes
https://github.com/apache/cordova-plugin-inappbrowser/pull/263
which they are likely to approve, but if not we can work off of the fork.
Moved all the iOS twitter scheme processing out of Application.js and into
TwitterSignIn.jsx, and enhanced it for Android.  Some small styling changes
to remove the iOS


DaleMcGrew:
👍

PR #1452, opened 2018-04-14
Styling changes for Cordova Android

SailingSteve:
Reduced the size of the $header-height-cordova for Android.
Removed the Cordova oriented www directory from WebApp, found a better
way to do this completely on the Cordova side -- requires fewer symbolic
links.


DaleMcGrew:
👍

PR #1453, opened 2018-04-16
Allow SignInByEmail to signin when email is read on the phone in Cordova

SailingSteve:
Resolves the iOS portion of wevote/WeVoteCordova#11


DaleMcGrew:
👍

PR #1457, opened 2018-04-18
Fixed issue where the elections button were not forwarding to the election in Cordova

SailingSteve:
We can't use browserHistory.push anymore, must use the wrapper historyPush()


DaleMcGrew:
👍

PR #1459, opened 2018-04-20
Changed "< Return to Ballot" into a Button

SailingSteve:
Tested with Cordova, WebApp, and Mobile WebApp
On the HamburgerMenu, Bookmarks menu item doesn't have a "Return" so I
removed the indent that indicates it is a setting.
BallotStore has a hard to reproduce bug where ballotItemListCandidatesDict
is occasionally is not initialized, so I put in an error log line so we
will have a hint about what went wrong.


DaleMcGrew:
👍

PR #1470, opened 2018-04-24
Changes for Cordova with iPad

SailingSteve:
HeaderBar: suppress Donate button
HeaderGettingStartedBar:  suppress print ballot
SettingsPersonalSideBar: lint warning cleanup
Ballot: Don't show BallotSideBar if Cordova. Anchored hrefs with #, are
relative to a url, and don't work as is in Cordova.
SettingsDashboard: The new dashboard could fit on the iPad size display,
but does not fit well with the HamburgerMenu settings choice, so
suppressed it if Cordova.
Application: Cordova specific console.log was displaying in the webapp
_navigator.scss, fixed HamburgerMenu column width for iPad.

### DIFF
--- a/src/js/Application.jsx
+++ b/src/js/Application.jsx
@@ -59,9 +59,8 @@ export default class Application extends Component {
   }
 
   initCordova () {
-    console.log("Application initCordova ------------ " + __filename);
-
     if (isCordova()) {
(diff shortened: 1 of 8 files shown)


DaleMcGrew:
👍

PR #1491, opened 2018-04-30
Prevent form submit temporarily, Create Your Voter Guide, Step 1 #1465

### LINKED ISSUE #1465: Prevent form submit temporarily, Create Your Voter Guide, Step 1
In Step 1 of "Create Your Voter Guide", we ask a voter to enter their Twitter handle:
https://wevote.us/voterguidegetstarted
In user testing we found that the voter enters their Twitter handle, and then hits return. The problem with this is that it doesn't give the page a chance to receive the response from the server with the Twitter data. I attempted a fix, and in fact changed the button to say "One Moment...", but I wasn't able to disable the form. Please disable the form until we have received a response from the server.
This is only a problem in desktop mode.

### Steps to reproduce the problem (1, 2, 3...), including links
The code for this is here:
WebApp/src/js/routes/VoterGuide/VoterGuideGetStarted.jsx

chulinguy:
This PR fixes #1465

-Hitting "enter" while Twitter has not returned will simply set a state "didUserHitEnter" to true
-The Twitter data fetch logic now also checks the "didUserHitEnter" state, and if that state is yes AND Twitter finds a voter guide
-removing redundant logic to set autofocus state 

##I am actually not sure how to user test this fix, so I'll try to join Google hangout!

### DIFF
--- a/src/js/routes/VoterGuide/VoterGuideGetStarted.jsx
+++ b/src/js/routes/VoterGuide/VoterGuideGetStarted.jsx
@@ -20,6 +20,7 @@ export default class VoterGuideGetStarted extends Component {
       twitterSearchStatus: "",
       isLoadingTwitterData: false,
       didUserPressEnter: false,
+      isTwitterHandleValid: false,
     };
     this.onOrganizationStoreChange = this.onOrganizationStoreChange.bind(this);
     this.resetState = this.resetState.bind(this);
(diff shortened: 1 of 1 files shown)


chulinguy:
OK this is working now ... I removed the onKeyDown event from the input because it would reset states and mess up my checks for historyPush

DaleMcGrew:
👍

PR #1495, opened 2018-05-02
Fixed wevote/WebApp/issues/1494 -- Two search bars on voter guide

SailingSteve:
Started fixing it in Application, then did a refactor of
HeaderGettingStartedBar, then found a better place to fix it in
OrganizationVoterGuide, and backed out most of my changes from
Application, except for the refactor.


DaleMcGrew:
👍

PR #1497, opened 2018-05-03
Retrieve updated Facebook profile picture

jainanisha90:


src/js/actions/FacebookActions.js
```diff
@@ -71,14 +71,16 @@ export default {
     });
   },
 
-  getFacebookProfilePicture: function (userId) {
+  getFacebookProfilePicture: function () {
     if (window.FB) {
```
DaleMcGrew: @jainanisha90, do you think "if (window.FB)" is still necessary? It seemed like (when I tested a few days ago), that code after "if (window.FB)" wasn't getting run. (I have not tested this fix yet, but I will.)

PR #1501, opened 2018-05-04
Some new space below the header: fixes for iOS

SailingSteve:
Made sure that the "Back to Ballot" button showed for Measures in Cordova.


DaleMcGrew:
👍

PR #1505, opened 2018-05-10
fix bug config-template

agheith:
Setting up we vote, I had discovered that `module.exports = ` to `export default`.


DaleMcGrew:
👍 Thank you @agheith for catching this configuration problem that caused a gulp error when you ran "npm start".

Issue #1519, opened 2018-05-16
Change "Ratings from Vote Smart" popover to a "read more" link at the bottom


DaleMcGrew:
Having a modal view in a popover is a no-no. Only show the description of the Org after they click. Don’t give the “Ratings from Vote Smart” so much room -- move it to the bottom of the card.
![screen shot 2018-05-15 at 5 53 45 pm](https://user-images.githubusercontent.com/7756031/40090813-f90d209c-5868-11e8-810f-8582c6e9126e.png)

DaleMcGrew:
Beautiful work @edward-ly. Verified.

Issue #1518, opened 2018-05-16
Update Votesmart/Candidate rating display to include the name of the candidate


DaleMcGrew:
Add text that says ACLU supports candidate in the popover. 96% rating *of Candidate*
Xavier has scored….Candidate name has scored yy in xx year by ww Organization.
![screen shot 2018-05-15 at 5 55 12 pm](https://user-images.githubusercontent.com/7756031/40090847-2f6aa484-5869-11e8-9993-14cd9f38e49c.png)

DaleMcGrew:
Thank you @edward-ly! Verified.

PR #1524, opened 2018-05-17
Updated organization popover content on ballot page

edward-ly:
Updated the VoteSmart rating display so that it includes the candidate name, it is broken up into two lines, and the rating icon is bottom-aligned with the percent rating (#1518). Moved the explanation of the VoteSmart rating to the bottom of the organization popover, with the rating and explanation header implemented to show/hide the text on click (#1519).


DaleMcGrew:
Thank you @edward-ly!

PR #1526, opened 2018-05-18
fix autocomplete on the Ballot page

agheith:
The Google autocomplete to offer suggested addresses on the Ballot page is functional as in the other components `WebApp/src/js/components/Settings/SettingsAddress.jsx`. The issue is that `.pac-container` that contains the list of autocompletes has a lower z-index value than the modal's z-index. Therefore, in the `WebApp/src/sass/components/_ballotModals.scss` I gave the modal a lower z-index value than the `.pac-container` to display the autocompletes.

### DIFF
--- a/src/sass/components/_ballotModals.scss
+++ b/src/sass/components/_ballotModals.scss
@@ -17,6 +17,7 @@
     bottom: 0;
     left: 0;
     overflow: auto;
+    z-index: 1050;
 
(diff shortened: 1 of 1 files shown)


DaleMcGrew:
Wow, I'm impressed you were able to track this down! How obscure. 👍 Thank you @agheith.

PR #1535, opened 2018-05-21
Facebook auth is working for Cordova in iOS and Android

SailingSteve:
FacebookActions relies on the initFacebook() in Application.jsx (if
isWebApp() ), otherwise relies on FacebookConnectPlugin4 if Cordova.
Still have to an unresolved issue with Network/Facebook "Choose" in
Cordova only, it works correctly in the WebApp.


DaleMcGrew:
Hi @SailingSteve, the Facebook "Choose Friends" feature is actually broken right now due to a change by Facebook. @jainanisha90 will be working on a fix for this in the coming weeks.

Issue #1527, opened 2018-05-18
Jump from "List of Offices" doesn't work when candidate not on page


DaleMcGrew:
### Please describe the issue (What happens? What do you expect?)
On the right side of the ballot page in Desktop mode, there is a "List of Offices". After a voter has supported a candidate under one of the offices, the voter is shown these three tab options:
![screen shot 2018-05-18 at 8 37 27 am](https://user-images.githubusercontent.com/7756031/40244097-cef1d450-5a76-11e8-919e-ad45b9dbc6f4.png)

If you go the "Remaining Decisions" tab and then click on an office in the "List of Offices" where you have already supported a candidate, you are not taken to the office. One solution would be to check to see if the voter is on one of these filtered ballot views:
https://wevote.us/ballot?type=filterRemaining
https://wevote.us/ballot?type=filterSupport

And if so, check to see if the Candidate is on either of these filtered views. If they are NOT, then do the anchor link to the "All Items" view.

DaleMcGrew:
Verified. Thank you @jingyiwang1209!

PR #1536, opened 2018-05-21
Made the jump from "List of Offices" go to the correct location when candidate not on page #1527

jingyiwang1209:


DaleMcGrew:
Thank you @jingyiwang1209! 👍

Issue #1523, opened 2018-05-17
Under each office, order the candidates (top-to-bottom) based on their scores


DaleMcGrew:
### Please describe the issue (What happens? What do you expect?)
When I unfurl an office on the ballot page, I would like the candidates to be ordered this way:
1) Highest Network Score
2) Highest Issue Score (a network score of 1 should be higher than any Issue score)

### Steps to reproduce the problem (1, 2, 3...), including links
I think the reordering can happen on this page, near the top of the render. Search for this:
WebApp/src/js/components/Ballot/OfficeItemCompressedRaccoon.jsx
this.state.candidateList.forEach((candidate)

The challenge will be getting the Network scores and Issue scores at the right time in the lifecycle, and preventing "churn".

DaleMcGrew:
Thank you @jingyiwang1209, this is a great improvement!

PR #1545, opened 2018-05-25
Under each office, order the candidates (top-to-bottom) based on their scores #1523

jingyiwang1209:
issue # 1523


DaleMcGrew:
Thank you @jingyiwang1209! 👍

PR #1553, opened 2018-05-29
Applied anchor and jumping features in vote guide page

jingyiwang1209:
Applied anchor and jumping features in vote guide page


DaleMcGrew:
👍

Issue #1567, opened 2018-06-08
Incorrect pointer when hovering over 'About' on top navigation


johnajackson:
**Issue Description**
The 'About' link in the top navigation does not display the correct pointer on hover.

**To Reproduce**
1. Visit any screen within WeVote and
2. Hover your pointer over the 'About' link in the top navigation

DaleMcGrew:
Resolved! Thank you @johnajackson.

PR #1568, opened 2018-06-08
Link Hover Fix

johnajackson:
The 'About' link in the top navigation does not display the correct pointer on hover.

Issue #1567

**Fix**
Enables pointer on hover for 'About' link.


DaleMcGrew:
👍

Issue #1484, opened 2018-04-26
Canceling Facebook login -- deal with going back in browser


DaleMcGrew:
### Please describe the issue (What happens? What do you expect?)
If you try to sign in with Facebook in mobile mode, we open a new window. If you decide you don't want to sign in and go back, you are returned to this page:
https://wevote.us/facebook_invitable_friends

...and all you see is the loading animation.

Please work through all "cancel" scenarios so that a voter can get back to We Vote, ideally the page they just came from.

ndrewr:
@DaleMcGrew is there a specific FACEBOOK_APP_ID we should utilize or should I just create my own and plug that into config temporarily?

ndrewr:
I was getting errors trying to login with facebook signin button.

Rearranged the parameters passed to the `login()` method according to the [docs](https://developers.facebook.com/docs/reference/javascript/FB.login/v3.0#permissions) and that seems to resolve the issue.

Besides that, the issue here seems to be that we don't check the `authResponse` value in the `login` callback. Checking that and preventing any followup dispatches to store may do the trick.

Sidenote:
The `getLoginStatus()` method will cache successful login attempts making it hard to test the flow (since clicking the signin button will just automatically auth you successfully until the token expires). To get around that, there is a second parameter according to the [doc page](https://developers.facebook.com/docs/reference/javascript/FB.getLoginStatus).

But oddly I still had to test in incognito window to prevent the auto-authorization.

DaleMcGrew:
Verified! Thank you @ndrewr.

PR #1569, opened 2018-06-12
Rearrange args sent to facebook login method; Check auth response fro…

ndrewr:
…m login attempt, cancel login flow if null

Addresses issue #1484 

Refactor arguments sent to facebook login method.
Check the response value sent from facebook login request. If null (indicating cancelled/rejected login attempt), return from the function back to the signin page.


DaleMcGrew:
Thank you for finding this fix @ndrewr! 👍

PR #1570, opened 2018-06-12
Disable copy and paste if not in an input field if in Cordova

SailingSteve:
Fixes wevote/WeVoteCordova/issues/37


DaleMcGrew:
👍

PR #1572, opened 2018-06-14
 Performance fixes for WebApp and Cordova

SailingSteve:
Fixes wevote/WeVoteCordova#65
In Ballot.js, some lint cleanup, and don't call
SupportActions.positionsCountForAllBallotItems if not needed.

In Candidate.js, don't call
VoterGuideActions.voterGuidesToFollowRetrieveByBallotItem for the same
candidate if the associated voter guides are already in cache.

There still is a server side issue with
voterGuidesToFollowRetrieveByBallotItem where we spend almost 45 seconds
processing the response, this fix just eliminates one unnecessary call the slow ap,i in this one situation.

### DIFF
--- a/src/js/routes/Ballot/Ballot.jsx
+++ b/src/js/routes/Ballot/Ballot.jsx
@@ -230,7 +230,7 @@ export default class Ballot extends Component {
       wait_until_voter_sign_in_completes: wait_until_voter_sign_in_completes,
     });
 
-    if (this.props.location && this.props.location.hashvoterGuidesForId) {
+    if (this.props.location && this.props.location.hash) {
       // this.hashLinkScroll();
       this.setState({ lastHashUsedInLinkScroll: this.props.location.hash });
     }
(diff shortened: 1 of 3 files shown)


src/js/routes/Ballot/Ballot.jsx
```diff
@@ -214,13 +230,13 @@ export default class Ballot extends Component {
       wait_until_voter_sign_in_completes: wait_until_voter_sign_in_completes,
     });
 
-    if (this.props.location && this.props.location.hash){
+    if (this.props.location && this.props.location.hashvoterGuidesForId) {
```
DaleMcGrew: Hi @SailingSteve, since this is a prop (as opposed to a local state), shouldn't it still be "this.props.location.hash" instead of "this.props.location.hashvoterGuidesForId"?

SailingSteve: yes, that is a complete typo.

SailingSteve:
@DaleMcGrew accepted your suggestion, it was a complete typo on my part.

DaleMcGrew:
👍

PR #1575, opened 2018-06-17
Performance fixes for WebApp and Cordova (issuesRetrieve)

SailingSteve:
This was the next biggest performance killer on the Cordova side, and
also was tested on the WebApp side.
issuesRetrieve was being called 11 times in 3 seconds, when navigating
between Ballot and Candidate, and returning identical data.  Now only
does this once or twice (twice if the first two queries are simultaneous).


DaleMcGrew:
👍

PR #1577, opened 2018-06-20
The Voting Information Project works in Cordova without an iFrame

SailingSteve:
Instead we load it as an external webpage -- actually our webpage
stored in aws s3.

Fixes Locator portion of wevote/WeVoteCordova/issues/49


DaleMcGrew:
👍

Issue #1552, opened 2018-05-29
Make sure measures show up under "Items Decided"


DaleMcGrew:
### Please describe the issue (What happens? What do you expect?)
If you choose to oppose a measure, the measure should show up under the "Items Decided" tab. Currently it does not because only supported items get displayed there. Please note that candidates that opposed should continue to NOT cause an office to show up under "Items Decided".

DaleMcGrew:
Hi @ndrewr I wanted to see if you had made any headway on this issue? If not feel to unassign yourself so someone else can pick up this issue. Thank you!

ndrewr:
Yeah I see where the list is built within `BallotStore`.

https://github.com/wevote/WebApp/blob/develop/src/js/stores/BallotStore.js#L146

It doesn't seem like the `ballot_supported` getter is used anywhere besides this case. We can perhaps just retrieve the decided-opposed items as well and return them under the same "filterType".

If not that, we can create a new filterType value e.g. "filterDecided"

DaleMcGrew:
Thank you @ndrewr we found a solution that leveraged some of your code. Thank you @edward-ly for solving this.

PR #1579, opened 2018-06-22
Return opposed items when displaying "Items Decided" on Ballot

ndrewr:
Addresses Issue #1552 

Have Ballot store getter return opposed items along with supported items for "Items Decided" filter.


DaleMcGrew:
Thank you @ndrewr, we were able to use your suggested change.

PR #1580, opened 2018-06-23
Facebook login was broken for Cordova, now is working again

SailingSteve:
Due to recent changes in FacebookActions
Simplified HeaderSecondaryNavBar
Refactored SecondaryNavBarItem into a component that extends React Component


DaleMcGrew:
👍

PR #1581, opened 2018-06-23
Replace 'I stand with' button with a 'success' message

agheith:
When a user clicks on the "I stand with 'Organization's name'" button, a 'thank you' message and a toast message appears.


DaleMcGrew:
👍

PR #1622, opened 2018-07-26
Fix Ballotpedia links had a glitch in the WebApp

SailingSteve:
Don't try to display the external link, unless the "more" has been clicked


DaleMcGrew:
👍

Issue #1507, opened 2018-05-10
itemActionBar - yes_vote_description formatting


DaleMcGrew:
### Please describe the issue (What happens? What do you expect?)
When a voter is looking at a measure that has text describing what a "Yes vote" and a "No vote" mean, the text is not formatted yet:
![screen shot 2018-05-10 at 3 35 13 pm](https://user-images.githubusercontent.com/7756031/39897594-ca115d30-5467-11e8-8c0d-d50d3c031616.png)

### Steps to reproduce the problem (1, 2, 3...), including links
The code for this is here:
WebApp/src/js/components/Widgets/ItemActionBar.jsx
Search for "yesVoteDescription"

josh-southern:
@DaleMcGrew Hey I've unassigned myself because I've been busy with work and I don't know how soon I can contribute again, so I want to leave this up for someone else to grab if they want to tackle it. Have a good hacking session tonight!

DaleMcGrew:
Looks great @edward-ly, thank you!

PR #1635, opened 2018-08-15
Added flex positioning to measure vote descriptions

edward-ly:
Added some formatting to the measure yes/no vote descriptions by adding Flexbox to position the descriptions to the right of their yes/no buttons as well as vertically align the yes/no buttons (#1507).


DaleMcGrew:
👍 Thank you @edward-ly.

Issue #1639, opened 2018-08-17
Move thumbs up/thumbs down to bottom of card


DaleMcGrew:
### Please describe the issue (What happens? What do you expect?)
We want to move the entire horizontal bar with Thumbs up, thumbs down and the score, to be down below the horizontal list of organizations ("Opinions about Xavier").
![screen shot 2018-08-17 at 1 57 09 pm](https://user-images.githubusercontent.com/7756031/44288545-93164500-a225-11e8-8325-cc6105ecd1ef.png)

DaleMcGrew:
Thank you Edward, this is resolved.

Issue #1638, opened 2018-08-17
Turn off Animation as modals open


DaleMcGrew:
### Please describe the issue (What happens? What do you expect?)
Currently when modals open in We Vote (for example, the one for choosing Issues), they animate in from the top of the screen. Unfortunately the performance of the animation is currently not good. Please turn off the animation so that the modal simply opens on the page.

DaleMcGrew:
Thank you @edward-ly, this is resolved.

Issue #1628, opened 2018-08-02
Remove "Issues" tab from the "Network" interface


DaleMcGrew:
### Please describe the issue (What happens? What do you expect?)
We don't want to delete the files, but we want to remove the management of what Issues you can choose from the "Network section". The default section visible should be "Organizations" as opposed to "Issues"
<img width="624" alt="screen shot 2018-08-02 at 8 59 19 am" src="https://user-images.githubusercontent.com/7756031/43595829-80e6e988-9632-11e8-8abc-415eaca5531a.png">

We can leave "Issues You Are Following" on this page (on the right side of the page in Desktop mode.

DaleMcGrew:
Resolved.

PR #1641, opened 2018-08-21
Updates to ballot page modals, layout, and network page tabs

edward-ly:
Turned off the slide down animation when opening modals (#1638). Moved the row containing the network actions for each candidate to the bottom of the candidate card (#1639). Removed the "Issues to Follow" tab from the Network page (#1628).


DaleMcGrew:
👍

Issue #1643, opened 2018-08-21
Remove issues under each candidate on Ballot (comment out the code)


DaleMcGrew:
### Please describe the issue (What happens? What do you expect?)
Currently we show Related Issues under each office and candidate. On the ballot page, we want to comment these displays out. 
Note: We want to leave Related Issues in place on the candidate page.
ROLLED UP
![screen shot 2018-08-21 at 8 53 46 am](https://user-images.githubusercontent.com/7756031/44413253-c7d51580-a51f-11e8-988a-e38a8480fefc.png)

UNFURLED
![screen shot 2018-08-21 at 8 50 37 am](https://user-images.githubusercontent.com/7756031/44413106-58f7bc80-a51f-11e8-9d41-d9f3ead1a960.png)

You can leave all of the listeners in place, but please comment out the code that generates the Issues display on the Ballot (since it is quite CPU expensive)

DaleMcGrew:
Verified, thank you @edward-ly.

PR #1647, opened 2018-08-22
Removed all instances of issues list on ballot page

edward-ly:
Removed the list of issues under each ballot item and candidate item on the ballot page while keeping the list on each individual candidate page (#1643).


DaleMcGrew:
Thank you @edward-ly! 👍

Issue #1627, opened 2018-08-02
Add checkbox "Show All States" under "Upcoming Elections"


DaleMcGrew:
### Please describe the issue (What happens? What do you expect?)
On the "change address or election" popup, a voter can choose another election in the "Upcoming Election(s)" section:
<img width="250" alt="screen shot 2018-08-02 at 8 48 10 am" src="https://user-images.githubusercontent.com/7756031/43595346-4334a202-9631-11e8-9594-38e860d5d2ee.png">

The list can get pretty overwhelming, so we want to make it easier for most voters to find the elections relevant to them.
- [x] If the voter's address contains a State, please change the default filter to only show upcoming elections in the voter's state.
- [x] Please also filter for prior elections using the same checkbox state (that is, don't add a second checkbox) 
- [x] When the voter checks the "Show All States" checkbox, refresh the view to show all upcoming elections
- [x] I believe the "Upcoming Elections" section goes away if there aren't any upcoming elections. Please make sure the checkbox is visible even if there aren't any upcoming elections (so it can be used for Prior Elections)
- [x] Add a link at the very bottom of the page called "Show prior elections". Do not show entire "Prior Election(s)" section until that link has been clicked. No link to "Hide" previous elections is needed.
- [x] Make sure National elections always appear in Upcoming and Previous elections
- [x] Add state to the header "Upcoming Elections in New York" (as long as a state has been identified)
- [x] Add state to the header "Previous Elections in New York"
- [x] When showing all states, remove " in New York"
- [x] Automatically change "Upcoming Elections" and "Previous Elections" from singular to plural
- [x] Update "Show All States" to "Show all states - 5 more elections" (convert to sentence case, and support the case "1 more election"). Make sure to hide "Show all states..." if there aren't any other state elections.

DaleMcGrew:
This is a great new feature, thank you @edward-ly!

PR #1656, opened 2018-08-25
Implemented show/hide elections by state in BallotElectionList

edward-ly:
Implemented hiding elections outside the user's state (currently based on IP address) in the lists of upcoming and prior elections on the ballot page. Changed the text on the election list headers as appropriate. Completely hidden the prior elections list until toggled on (#1627).


DaleMcGrew:
👍 Thank you @edward-ly. This will be a great feature.

Issue #1644, opened 2018-08-21
Opening Issues modal brings up keyboard on iPhone


DaleMcGrew:
On the ballot page, using an iPhone, when you click on the Issues button:
![screen shot 2018-08-21 at 8 55 56 am](https://user-images.githubusercontent.com/7756031/44413365-0b2f8400-a520-11e8-821d-31865a557322.png)

The modal opens with the iPhone keyboard. We do not want the keyboard to open. I think this is because we added a text input box (for the Voter's address) on the third slider page. I'm hoping there is a way to prevent the keyboard from opening until the voter gets to the third slider page.
![img_5520](https://user-images.githubusercontent.com/7756031/44413559-8002be00-a520-11e8-98f9-8e06984161c2.jpeg)

DaleMcGrew:
Resolved. Thank you @edward-ly!

Issue #1618, opened 2018-07-26
Move Election Date to left of Election Name


DaleMcGrew:
### Please describe the issue (What happens? What do you expect?)
We Vote offers voters a way to choose which election they want to look at. In desktop view (as opposed to mobile), show the election date on the left side of the election name when you choose an election (in the Green buttons below), as opposed to below the Election Name. Note: the layout can stay the same in mobile mode.

<img width="772" alt="screen shot 2018-07-25 at 6 53 15 pm" src="https://user-images.githubusercontent.com/7756031/43236890-2204e3c6-903c-11e8-8525-f5494b19d16b.png">

DaleMcGrew:
Verified!

PR #1659, opened 2018-08-29
Moved election date to left of election name, manually focus address box in issues modal

edward-ly:
Moved the election date to the left of the election name in the ballot election list in desktop mode (#1618). Manually focus/blur the address box in the issues modal so that the keyboard for mobile devices opens/closes at the appropriate times (#1644).


DaleMcGrew:
Thank you @edward-ly! 👍

Issue #1665, opened 2018-09-04
Changes to "Opinions about..." label to the left of the organization logos


DaleMcGrew:
### Please describe the issue (What happens? What do you expect?)
![screen shot 2018-09-04 at 3 40 16 pm](https://user-images.githubusercontent.com/7756031/45061339-f216ef00-b058-11e8-8314-381b9be9804e.png)
- [x] Instead of "Opinions about Xavier" (wrapped on two lines), please change to "15 Network Opinions" (wrapped on two lines, with "15 Network" on one line, and "Opinions" on the second line
- [x] Please leave the "i" icon where it is
- [x] Please leave "Opinions about Xavier Becerra" as-is in the popover
- [x] When there aren't any opinions, show "No Network Opinions" (split onto two lines like above), even though there won't be any logos to the right.
- [x] If you have the only opinion about a candidate, add a new modal (copied from the modal when there are opinions) that says: "You have the only opinion about Xavier Becerra so far."
- [x] Add a new modal (copied from the modal when there are opinions) that says: "There are no opinions about Xavier Becerra yet."
![screen shot 2018-09-04 at 3 41 24 pm](https://user-images.githubusercontent.com/7756031/45061520-b0d30f00-b059-11e8-88f4-687b73cfa06a.png)

DaleMcGrew:
Verified!

PR #1666, opened 2018-09-05
Changed network opinions list label and popover text on ballot page

edward-ly:
Changed the label to the left of the network opinions list on the ballot page to indicate the number of opinions in the list, and changed the text of the label's corresponding popover accordingly (#1665). Explicitly declared the use of no-cache in index.html.


DaleMcGrew:
👍

PR #1680, opened 2018-09-17
iPhone X Display Problem with Camera Cutout

SailingSteve:
wevote/WeVoteCordova/issues/76
Css/sass needs to be checked in Cordova when it is changed in the Web
app, to avoid these in the future.
Fixed layout for the X on the modal dialogs, the search box distance
from the top, the base distance from the top, and the "Back to settings"
button distance from the top.
Added new isIPhoneX() function in cordovaUtils.js that (unfortunately)
detects the iPhone X and "X S Max" by screen size.
There is no simulator for the "X S Max" yet, so will need to test again
in a week or two when it arrives.


DaleMcGrew:
👍

Issue #1669, opened 2018-09-06
"Post" button on comment box should be "Save" prior to signing in


DaleMcGrew:
### Please describe the issue (What happens? What do you expect?)
When you haven't signed in yet, the text "Post" should instead by "Save". Once you sign in, it should revert to "Post".
<img width="486" alt="screen shot 2018-09-05 at 5 41 21 pm" src="https://user-images.githubusercontent.com/7756031/45128480-1c3cdf80-b133-11e8-83dd-0c8d788f893e.png">

DaleMcGrew:
Closed.

PR #1683, opened 2018-09-17
change POST button to SAVE when user is not signed in

MilcahHalili:
#1669


DaleMcGrew:
Thank you @MilcahHalili! 👍

Issue #1679, opened 2018-09-17
Move "Related Issues" on candidate page to the bottom of the page


DaleMcGrew:
![screen shot 2018-09-17 at 7 04 34 am](https://user-images.githubusercontent.com/7756031/45628017-02ab6a00-ba48-11e8-855d-fc41f7d21ba9.png)

DaleMcGrew:
There is still a display issue (which I asked @edward-ly to leave so someone else can fix it), but this issue is now complete. Thank you Edward!

PR #1687, opened 2018-09-19
Moved "Related Issues" on candidate page to bottom of page

edward-ly:
Moved the "Related Issues" list out of the candidate entry on the candidate page to the bottom of the page (#1679). Changed the position of the issue list popovers to show up above the list rather than below. Attempted to fix an issue where the popovers are still not positioned correctly.


DaleMcGrew:
👍

Issue #1684, opened 2018-09-18
Add Ballot Filtering: Federal, State, Local, Measures


DaleMcGrew:
### Please describe the issue (What happens? What do you expect?)
On the ballot page, add these four "race level" filters (Federal, State, Local, Measures) underneath the "All Items, etc." tabs.

I have added a new value in the voterBallotItemsRetrieve API response called "race_office_level" that you can use to know whether or not to turn each item on or off based on the filter:
"race_office_level": "Federal",
"race_office_level": "State",
"race_office_level": "Local",
(Measures do not have race_office_level.)

DaleMcGrew:
Hi Edward, if you have any time tomorrow (Friday, Sept 21st), would you be able to turn off the "race_levels" that are empty? For example, in this screenshot, the Federal Offices and Measures button selectors, and from the right-hand "Summary of Ballot Items".

![screen shot 2018-09-20 at 10 50 17 pm](https://user-images.githubusercontent.com/7756031/45862680-b2722780-bd27-11e8-81c1-d8a104ebcc67.png)

DaleMcGrew:
Excellent work @edward-ly!!!!

PR #1688, opened 2018-09-20
Added ballot filtering by race office level on ballot page

edward-ly:
Implemented filtering of ballot items by "Federal", "State", "Local", and "Measure" and added buttons above the list of ballot items to switch between filters (#1684). Added horizontal inset CSS layout classes for utility.


DaleMcGrew:
👍

PR #1689, opened 2018-09-21
Grouped ballot item links in ballot side bar by race_office_level

edward-ly:
Re-organized the BallotSideBar by grouping ballot items of the same type or race_office_level together. Changed the order of the ballot list filter buttons to be the same as in the order in the side bar. Fixed the behavior of the ballot item links to account for the filtered ballot list.


DaleMcGrew:
Thank you @edward-ly! Amazing work. 👍

PR #1690, opened 2018-09-21
Fixed a bug that prevented measure descriptions from displaying on ballot page, hide filter buttons not in use

edward-ly:
Fixed a bug that prevented ballot measure descriptions from displaying on the ballot page after the "Measure" filter is selected more than once and the list of ballot items is reconstructed. Used setState's callback function argument to correctly calculate the values of the measure state variables.

Implemented the hiding of each filter button if a ballot contains zero items of each filter type (#1684). Adjusted the space around the filter buttons to be equal on all sides.


DaleMcGrew:
👍 Thank you @edward-ly!

PR #1692, opened 2018-09-24
Blue bar covers top of search box, after searching for candidate

SailingSteve:
It was the ios7plus-spacer that was appearing in the wrong circumstances


DaleMcGrew:
Thank you @SailingSteve! 👍

PR #1716, opened 2018-10-12
Upgraded to React 16, Bootstrap 4, and Gulp 4

SailingSteve:
Upgraded to React 16, Bootstrap 4, and Gulp 4

React 16 has some breaking changes (they say it doesn't), it is much more picky about passed parameters, or possibly it just passes through some parameters that it didn't in the past.

React 16, requires that the local Node on your Mac be upgraded to 10.10.0

ReactBootstrap 1.00 (we were at 0.32) is Bootstrap 4 compatible, and we now load Bootstrap 4.  Bootstrap 4 no longer honors the (previously deprecated) "hidden-*" or "visible-*" which we had 
used extensively, there are functional equivalents.  See https://getbootstrap.com/docs/4.1/utilities/display/ for the substitutes.  This was also true for
React 16 wich also requires the substitutes, all instances of "hidden-*" and "visible-*" have been converted.

Bootstrap 4, does not contain the Glyphicons from http://www.glyphicons.com/ anymore, so I downloaded the ones we previously got from Bootstrap into src/img/global/svg-icons/glyphicons-pro-halflings and src/img/global/svg-icons/glyphicons-pro-social and now use

ReactCSSTransitionGroup is deprecated and has been replaced withCSSTransitionGroup and CSSTransition (these two classes must be called adjacently -- put any other class inbetween and you will get an error.)
As a result there are significant changes in each file that previously used ReactCSSTransitionGroup.

Add "ENABLE_FACEBOOK: true," to your config.js

In BallotSideBar, the npsinput.js project has been commented out, it looks like we had disabled it, 
and it has not been updated in 2 years, and it caused problems with React 16.

In index.html, I added https prefixes to all the loading of script packages since without "https://" specified, they were all starting off http and then being redirected from http to https every time, 
and they are now loaded from the cloudflare's CDN, which is very very fast.

Seems to reset/crash/reload after a complete load on the candidate page when using the chrome
debugger, I think I saw that on a weeks old code using California election 6000 and the 16 second 
long voterGuidesToFollowRetrieve API call.

The crash seems to be related to the Facebook api code that we download on the fly, that's why I 
put in the "ENABLE_FACEBOOK" config option, but haven't had a chance to test it yet.

10/12/18: After pinterest, the gulpfile now includes three scss files
and processes them.  The index.html, now loads the resultant css files,
and the styles appear in the dom.


DaleMcGrew:
Thank you @SailingSteve I am merging this into the develop branch. 👍

PR #1731, opened 2018-10-18
Fixed Issue on Alphabetize Measures on Ballot

ACODE91:
Fixed issue by sorting filtered_ballot_items array with the "MEASURE" category by using the localeCompare method.

### DIFF
--- a/src/js/components/Ballot/BallotItem.jsx
+++ b/src/js/components/Ballot/BallotItem.jsx
@@ -23,7 +23,6 @@ export default class BallotItem extends Component {
   }
 
   render () {
-    console.log('aarons log', this.props.kind_of_ballot_item)
     renderLog(__filename);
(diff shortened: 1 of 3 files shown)


DaleMcGrew:
Hello @ACODE91, it was great to meet you last night. Thank you for this pull request! Comments on this pull request:
- [ ] Please remove the console log in src/js/components/Ballot/BallotItem.jsx
- [ ] Normally we don't check in config.js, but in this case those are default changes everyone will need. In future checkins please make sure to not check in config.js.
- [ ] It looks like your solution in src/js/stores/BallotStore.js only includes the Measures? I would expect to see code that removes the measures from filtered_ballot_items and then adds them back in from alphanumeric_filtered_items.

DaleMcGrew:
👍 Thank you @ACODE91!

PR #1736, opened 2018-10-19
Thursday's progress on React 16 for iOS

SailingSteve:
Intro network is an example on how React 16 wants refs handled, the ones
that I didn't change seem to still work, so the others are low priority.

The margin at the top of the ballot, to stop it from going under the
secondaryHeader is different for different iPhone sizes.  Haven't tested
Android yet.

The iPhone XR needs a different secondaryHeader height, and voter photo/
default avatar on the HamburgerMenu.

Bootstrap 3 style center-block went missing in 4, and has been added to
_bootstrap3LeftBehind.scss

Known issues:
The issues modal, "what do you care about covers" the next button and the
slick dots.
On the about page, the icons at the top have inconsistent size.
The react player does not work in Cordova, it is because the video moved
to youtube from vimeo, needs permissions work in our Cordova repo.


DaleMcGrew:
👍

Issue #1712, opened 2018-10-07
Office/Race "Level" (Federal, State, Local) Filter buttons: Remove ability to de-select


DaleMcGrew:
### Please describe the issue (What happens? What do you expect?)
Currently if you are looking at Federal races (or State, or Local) on the Ballot, and then you click the Office/Race "Level" filter button again, it "de-selects" the filter, and shows you ALL items. This is confusing, and the option to de-select a filter should be removed.

<img width="632" alt="screen shot 2018-10-07 at 4 49 26 am" src="https://user-images.githubusercontent.com/7756031/46581396-e4a0ac00-c9ec-11e8-90c1-5f15f8f4d9d2.png">

DaleMcGrew:
Verified! Part of this solution is the "componentShouldUpdate" code that Dale did, so we might need to revisit this issue if that method changes. Thank you @ACODE91.

PR #1737, opened 2018-10-19
Resolved Issue: #1712  Office/Race "Level" (Federal, State, Local) Filter buttons: Remove ability to de-select

ACODE91:
Fixed filter issue by removing the ability to display all items on the second click of a filter button.


DaleMcGrew:
👍

PR #1742, opened 2018-10-21
Friday's progress on React 16 for iOS

SailingSteve:
Replaced ReactBootstrapToggle (3k downloads per wk), with ReactToggle
(33k downloads per wk), to resolve a messy styling issue triggered by
React 16.

Switched the loading of font-awsome css from maxcdn to cloudflare, which
was about 5 times faster for this file.


DaleMcGrew:
Hi @SailingSteve I'd like to recommend that you delete your config.js file and then recreate it, so it doesn't get picked up when you do check in's. Thank you!

Issue #1743, opened 2018-10-21
ItemSupportOppose - thumbs up/down showing duplicates


DaleMcGrew:
### Please describe the issue (What happens? What do you expect?)
When in tablet width (wider than mobile, but not as wide as desktop), duplicate thumbs up icons and thumbs down icons are appearing, when we should only see one. Needs bootstrap class visibility fine tuning:
<img width="505" alt="screen shot 2018-10-21 at 1 58 39 pm" src="https://user-images.githubusercontent.com/7756031/47272376-a30a1800-d539-11e8-8c0a-7d7f707dd937.png">

DaleMcGrew:
Fixed! Thank you @ACODE91.

PR #1750, opened 2018-10-23
Fixed issue #1743 ItemSupportOppose

ACODE91:
The issue originally occurred because the original code did not set the proper upper and lower bound of width adjustments for mobile/tablet and desktop devices.

After investigating the ranges of min width of most used mobile/tablet devices, I found the range of width pixels to be anywhere from 320px (min) to 992px (max, specifically for iPad Pro users). Therefore, specifying "d-lg-none d-xl-none" as our div's class names will allow the support/oppose button to be visible only to mobile/tablet users because we are hiding these elements from large and extra large screens. Same with the upper bound width issue, by specifying "d-none d-lg-block" as our div's class names to will allow the support/oppose button to be visible to desktop screens and hide the button from lower screen width devices. 

Sources: 
https://getbootstrap.com/docs/4.0/utilities/display/
https://stackoverflow.com/questions/47876143/how-to-render-different-components-based-off-device-size
https://codepen.io/_yatko/pen/ZJXQxy


DaleMcGrew:
👍 Thank you Aaron!

PR #1753, opened 2018-10-23
React 16 for iOS: Post button works

SailingSteve:
Left space for the slot at the bottom of iphone X* phones.
Added another "class" where is should be "className" in
ItemPositionStatementActionBar to avoid a crash in Cordova when
clicking into the text field.


DaleMcGrew:
👍

Issue #1767, opened 2018-10-25
TransitionGroup animations are not fully implemented


gj262:
A number of list displays use TransitionGroup/CSSTransition in what looks to be an attempt to add some kind of fade animation to items as they are added to/removed from the list.

However the code is missing the appropriate CSS to make the animation happen (i.e. .fade-enter/exit*) and the keys on the CSSTransitions should use a unique id related to the item if one wants to animate on exit. 

I found this confusing so opening this issue in case anyone else runs into it. However as it stands the code will render these lists just fine (though there could be a performance hit as TransitionGroup & CSSTransition do their work).

DaleMcGrew:
Resolved.

PR #1769, opened 2018-10-26
Guides list churn

### LINKED ISSUE #1710: Reduce Search "Churn" on Network Page
On the network page, a voter can search for organizations to follow. When you type in the search box, it starts searching as you are typing. What can happen is that the API call can cause "churn" as the responses come in and try to get loaded on the page. I think there are a few things we can do to reduce this problem:

- [ ] Do not start instant search until there are at least 3 characters
- [ ] Limit the requested search results to 50
- [ ] Increase the delay after a key stroke before requesting from the API
- [ ] Note also that deleting the contents of the search box should make a fresh call to the voterGuidesToFollowRetrieve API in order to reset the search page to the original listing.

This is the API Call used:
https://api.wevoteusa.org/apis/v1/voterGuidesToFollowRetrieve/?google_civic_election_id=0&start_retrieve_at_this_number=0&maximum_number_to_retrieve=25&search_string=pla&add_voter_guides_not_from_election=false

screen shot 2018-10-06 at 4 15 23 am

gj262:
Fixes #1710 

- TransitionGroup & CSSTransition were causing a noticeable delay in processing list updates and it seemed queuing prior updates so these are removed. (See also: #1767)
- The loadEndpoint ajax request is bubbled up to the action caller so that in _can_ be aborted. This ensures render for the most recent search term rather then render for the most recent search response. i.e. avoids overlapping requests/response.
- The delay between last change to the search term and making the request is reduced to 100ms to make it more responsive (the above fixes mean it no longer needs to be delayed to 1s).

It would also be a good thing™️ to give the user feedback that a change to the list is underway (initial load or responding to a search term). I've noticed > 5 second response times. Perhaps adding a "pending-" + endpoint action dispatch to Dispatcher.js could be a future improvement.

### DIFF
--- a/src/js/actions/VoterGuideActions.js
+++ b/src/js/actions/VoterGuideActions.js
@@ -17,7 +17,7 @@ export default {
 
   voterGuidesToFollowRetrieve: function (election_id, search_string, add_voter_guides_not_from_election, start_retrieve_at_this_number = 0) {
     let maximum_number_to_retrieve = 50;
-    Dispatcher.loadEndpoint("voterGuidesToFollowRetrieve", {
+    return Dispatcher.loadEndpoint("voterGuidesToFollowRetrieve", {
       google_civic_election_id: election_id,
       start_retrieve_at_this_number: start_retrieve_at_this_number,
       maximum_number_to_retrieve: maximum_number_to_retrieve,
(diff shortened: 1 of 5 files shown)


DaleMcGrew:
👍 Thank you @gj262!

PR #1776, opened 2018-10-29
update main nav icons

ixley:
This just updates the main nav items to be more visually consistent with the other updated ones.

image


DaleMcGrew:
👍 Beautiful!

ixley:
ahh.. just realized I didn't replace the ones that are in the footer. I'll get to that later.

PR #1778, opened 2018-10-29
More device specific spacing for Cordova

SailingSteve:
Cleaned out a few lint warnings
Cleared Icon property exception in ItemActionBar
Simplified device specific header size code and moved most of it to cordovaUtils from Ballot.


DaleMcGrew:
👍

PR #1779, opened 2018-10-31
Eliminate gray gap above ballot and Network pane for both iOS and Android

SailingSteve:
HeaderSecondaryNavBar PollingLocator toggle not called from
PollingPlaceLocatorModal when using Cordova SafariViewController to load
vip.html.  All the _open functions that were actually toggle functions
were renamed.
In cordovaUtils.js, lots of changes to support all the various device
sizes, for iOS we support and have tested on each device from iPhone 6
on to the latest iPhone XS Max.  On Android, we just check for one of
four screen sizes (by ranges of megapixels of screen real-estate) and
make some assumptions.  Worst case with have some grey useless screen
below the SecondaryNavBar, or on top of the Network page.


DaleMcGrew:
👍

PR #1782, opened 2018-10-31
add issue icons

ixley:
This is the first pass at the issue icons. They are a bit messy as a set but should work for this proof of concept. We can revisit and selectively update them moving forward. They are currently sized at 24px and probably shouldn't be scaled much below 20px.


DaleMcGrew:
Thank you @ixley! 👍

PR #1786, opened 2018-11-01
Simplified, fixed, and decoupled header space for WebApp and Cordova

SailingSteve:
We should now be able to change header spacing for the WebApp and
individual devices in Cordova, without breaking the rest.


DaleMcGrew:
👍

PR #1793, opened 2018-11-03
update styling for issues endorsements section

ixley:
This PR includes a few things but the main changes were to update the styling for the issues endorsements section. 

### Changes included this pull request
- Added remaining issues icons
- Cleaned up some old and conflicting Bootstrap references, including some `.btn` overrides. We should double check to make sure buttons look okay, but the override seemed to be causing problems and didn't appear to be directly solving for anything.
- Updated some of the styling for issues endorsements. 
- *I removed the thumbs-up icon next to endorsements for the time being until we're able to include the desired version.

----------
image


DaleMcGrew:
Thank you @ixley! Can you please revert the deletion of those prior svg's and update this pull request? Prior bundles will still be requesting those assets, so we need to leave them in the repo for at least a couple of more months.

Also, the src/js/components/Issues/IssuesByBallotItemDisplayList.jsx code is built to show the number of support and oppose endorsements, so I think we will need the icons so there aren't just two numbers side by side.

Thank you, -Dale

ixley:
@dalemcgrew Sorry, I shouldn’t have removed those unrelated images without checking. I’m still interested to better understand what continues to reference them. I’m a big stickler for keeping things clean and organized so I was just trying to do some easy cleanup. 

After elections perhaps we can touch base about some front end cleanup initiatives and how they might fit into overall priorities. My familiarity is such that I at least have a general sense where code could be cleaned up but can’t necessarily lead those efforts. I know tech debt is always a tricky thing to balance against feature work but I have the sense that some cleanup could make the app easier to navigate and work on. That’s at least my impression. I know you have better overall visibility!

DaleMcGrew:
Hi Jeff, it is the install iPhone and Android apps that reference those older svg's. Even when we update the website, those non-upgraded apps continue to look for older assets. A few weeks after the election, we can start deleting older assets.

I am looking forward to refactoring and clean up after the election.

PR #1798, opened 2018-11-05
Refactored header spacing down to one section of code in Application, plus better iPad support

SailingSteve:
Refactored header spacing down to one section of code for Application, plus better iPad support.
I moved header size calculation to cordovaUtils.appBaseClass() for web app and the 12
variants of Cordova iOS and Android devices.  

Cleared all the open csslint issues, and most of the js lint issues.
In FacebookActions, in the facebookApi call to ":/me?fields=picture..."
I found that sending in "GET" as a parameter was causing a nasty uncaught
exception in the c++ code that was messing up the stack, which forced the
(illogical) requirement that we send in class= instead of className= in
ItemPositionActionBar and AddressBox to stop a c++ level app crash when
clicking in a text entry field.  Changed the class= to className= in
both files.  Resolves wevote/WeVoteCordova/issues/89

In a couple of files added in getDerrivedStateFromError and
componentDidCatch React 16 state handlers -- they provide an
error-boundary which catches otherwise uncaught errors - solved a few
problems already with these!
end ballot to my friends, needs to send links to the webapp, not to the mobile device. Fixes recent addition to /wevote/WeVoteCordova/issues/11 

Load the vip.html static page from the webapp instead of from AWS, works on my local, but the Nginx rules might need adjustment to make this work in production.  Resolves wevote/WeVoteCordova/issues/83


src/js/Application.jsx
```diff
@@ -411,14 +420,41 @@ export default class Application extends Component {
...
     let appBaseClass = "app-base";
-    if (isCordova()) {
-      if (!hasIPhoneNotch()) {
-        appBaseClass += " cordova-base-iphonex";
+    if (isWebApp()) {
+      appBaseClass += " headroom-webapp";
+    } else {
+      appBaseClass += " cordova-base";
+      if (isIOS()) {
+        appBaseClass += " headroom-ios";
```
DaleMcGrew: Hi @SailingSteve, we often search for styles based on the BEM naming conventions that assume stylenames like "app-base-headroom-webapp" are all one continuous script. http://getbem.com/naming/ 
Lines that start with "__" or "--" indicate a style definition that is on its own line in the scss, and style names with "-" in them should not be split across multiple lines. Could you please refactor this pull request?

SailingSteve: @DaleMcGrew 

Hi Dale,
  I made the change you requested.  Please don't merge it, I haven't tested the new way yet (takes an hour or two to go through all the permutations). 
  We build up class names conditionally all over the place, but the way I did it here takes that further than before.
  To me the continuous way seems like sprawling duplicated code, about 3x as much, and less clear and more mistake prone than before.  I'll probably have to add sizes for iPads at some point, so it will get longer.
  If you really want it this continuous way, I'll test it and remove the "built up" way.

Thanks,
Steve

DaleMcGrew: Hi Steve, code like this where there is a space breaking up the styles is fine:
let appBaseClass = "app-base";
if (isWebApp()) {
  appBaseClass += " headroom-webapp";

What would *break* BEM naming standards (and how we search for the styles across our code base) would be if you did this:
let appBaseClass = "app-base";
if (isWebApp()) {
  appBaseClass += "-headroom-webapp";

Or this, since "getAndroidSize" tacks on text like "-xl". You want to keep all styles like "headroom-android-xl" as one continue string. Now if you made "getAndroidSize" return "--xl" then that would be ok and would follow BEM:
appBaseClass += " headroom-android";
appBaseClass += getAndroidSize();

src/js/Application.jsx
```diff
@@ -411,14 +420,117 @@ export default class Application extends Component {
...
+        appBaseClass = "app-base headroom-webapp-secondary";
+      } else if (isBackto) {
+        appBaseClass = "app-base headroom-webapp-backto";
+      } else {
+        appBaseClass = "app-base headroom-webapp-full";
+      }
+    } else if (isAndroid()) {
+      if (androidSize === "-xl" && isSecondary) {
+        appBaseClass = "app-base cordova-base headroom-android-xl-secondary";
+      } else if (androidSize === "-lg" && isSecondary) {
```
DaleMcGrew: Repeating "app-base cordova-base " isn't necessary. You could do this following code and it would be fine, since there is a space between "cordova-base" and "headroom-android-xl-secondary":
} else if (isAndroid()) {
      appBaseClass = "app-base cordova-base";
      if (androidSize === "-xl" && isSecondary) {
        appBaseClass += " headroom-android-xl-secondary";

src/sass/base/_base.scss
```diff
@@ -38,15 +38,41 @@ body {
...
 }
 
-.cordova-base-iphonex {
-  padding-top: $space-lg;
-}
+// stylelint-disable sh-waqar/declaration-use-variable
+.headroom-webapp-secondary      { margin-top: 9vh; }
+.headroom-webapp-backto         { margin-top: 3vh; }
+.headroom-webapp-full           { margin-top: 3vh; }
+.headroom-ios-notch-secondary   { margin-top: 10vh; }
```
DaleMcGrew: Per this explanation, https://css-tricks.com/snippets/sass/bem-mixins/ you might want to change these to:
.headroom-webapp {
  &--secondary { margin-top: 9vh; }
  &--backto         { margin-top: 3vh; }
  &--full           { margin-top: 3vh; }
}
.headroom-ios {
  &--notch-secondary   { margin-top: 10vh; }
}

SailingSteve: @DaleMcGrew 
I have studied mixins this week, and I don't think they would help in this situation.  My challenge is that the actual offsets as percentages are as follows from _base.scss ...
```
.headroom-webapp-secondary      { margin-top: 9vh; }
.headroom-webapp-backto         { margin-top: 3vh; }
.headroom-webapp-full           { margin-top: 3vh; }
.headroom-ios-notch-secondary   { margin-top: 10vh; }
.headroom-ios-notch-backto      { margin-top: 8vh; }
.headroom-ios-notch-full        { margin-top: -2vh; }
.headroom-ios-678-secondary     { margin-top: 10vh; }
.headroom-ios-678-backto        { margin-top: -1vh; }
.headroom-ios-678-full          { margin-top: -5vh; }
.headroom-ios-678plus-secondary { margin-top: 10vh; }
.headroom-ios-678plus-backto    { margin-top: 6vh; }
.headroom-ios-678plus-full      { margin-top: -4vh; }
.headroom-ios-ipad-secondary    { margin-top: 7vh; }
.headroom-ios-ipad-backto       { margin-top: 3vh; }
.headroom-ios-ipad-full         { margin-top: 1vh; }
.headroom-android-xl-secondary  { margin-top: 7vh; }
.headroom-android-xl-backto     { margin-top: -3vh; }
.headroom-android-xl-full       { margin-top: -8vh; }
.headroom-android-lg-secondary  { margin-top: 8vh; }
.headroom-android-lg-backto     { margin-top: -3vh; }
.headroom-android-lg-full       { margin-top: -8vh; }
.headroom-android-md-secondary  { margin-top: 8vh; }
.headroom-android-md-backto     { margin-top: -3vh; }
.headroom-android-md-full       { margin-top: -8vh; }
.headroom-android-sm-secondary  { margin-top: 10vh; }
.headroom-android-sm-backto     { margin-top: -5vh; }
.headroom-android-sm-full       { margin-top: -10vh; }
```

All that is left, is how I encode them with sccs.

src/js/Application.jsx
SailingSteve: My original pass met this criteria: "Lines that start with "__" or "--" indicate a style definition that is on its own line in the scss, and style names with "-" in them should not be split across multiple lines. Could you please refactor this pull request?"  See line 48 through 74 of _bass.scss in this PR.

SailingSteve: @DaleMcGrew 
I updated all we talked about this morning, and merged in yesterday's changes that eliminate the FacebookAPI exception that was manifesting in the phone when you clicked into a text field.
I'll go on to the twitter signin issue.

DaleMcGrew:
👍

PR #1803, opened 2018-11-08
Fix twitter signin bug

SailingSteve:
Also fixed a anonymous icon style problem.
Also fixed a scss calc typo in _issues.scss
Addresses the new bug at the bottom of wevote/WeVoteCordova/issues/11


DaleMcGrew:
👍

PR #1807, opened 2018-11-12
Fixed black bar above keyboard on iPhone 7

SailingSteve:
On certain occasions this was happening on all iOS phones, but it no
longer shows up in the simulators, or on my iPhone 7plus.

Re-enabled email ballot in Cordova


DaleMcGrew:
👍

PR #1811, opened 2018-11-26
Fixed Facebook login on Cordova to work with the latest Facebook SDK v4.38.1

SailingSteve:
The order of the login function parameters had changed.  Now requires
an array of strings for permissions instead of JSON.


DaleMcGrew:
👍 Thank you @SailingSteve!

PR #1813, opened 2018-11-27
Also update package-lock.json

josephevans:
Bump gulp-livereload version so we don't include malicious flatmap-stream package marked as dependency by event-stream@3.3.6


DaleMcGrew:
👍

PR #1814, opened 2018-11-27
Refactored top header into a separate component

SailingSteve:
This will (eventually)allow us to avoid all those device specific header offsets


DaleMcGrew:
👍

PR #1816, opened 2018-12-04
Adopted Airbnb lint rules as the default for the WebApp

SailingSteve:
Airbnb rules have been added to enforce code consistency, and reduce
bugs.

After adding Airbnb rules with the "--fix" option, it made changes to
almost every file... In made lots of changes to code style, it got rid of hundreds
of bits of code that did nothing, it found a number of real bugs (which have
now have been resolved), simplified hundreds of import lines, and flagged
22 thousand small rule violations most of which have now been resolved.
What remains is about 6700 rule warnings, of which 3500 are camelcase warnings.
The "--fix" option does not handle the spacing of plain html text very well
for a line containing only "Your network is undecided", but almost all
lines like that are now properly indented.

Let's fix the camel case warnings as we do regular maintenance, they can
not be automatically fixed, since sometimes they are state variables that
are sent with {...state} spread.  Some are variable names in the Python
format that come in from the API.  All can be addressed over time.

The "React property destructuring" warning is currently turned off, it
would add about 4000 warnings, but it is a good feature that we should
enable over time.  It encourages intermediate variables for elements of
structures like using
```const { weVoteId }  = this.props;``` to refer to weVoteId instead of
```this.props.weVoteId``` which avoids many reference problems and allows the
"compiler" to check structure references.

see https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/destructuring-assignment.md

New mandatory rules (all violations have been fixed)
* correct indentation (to the best of eslint's current abilities)
* use of const instead of let wherever possible
* useless constructors are not allowed
* `tabIndex` should only be declared on interactive elements
* Curly braces for multi-line if statements
* Unnecessarily quoted properties not allowed
* Hard tab characters are flagged as errors
* HTML entities must be escaped (only the most potentially confusing ones) like '>' to &gt;
* Parenthesis around multi-line jsx
* Nested ternary operators ```? :``` are not allowed.  Simple ones have been fixed, errors for the complex ones suppressed to warnings until the file receives other maintenance.
See https://blog.logrocket.com/5-common-practices-that-you-can-stop-doing-in-react-9e866df5d269 for a good way to deal with them.

New warnings
* no more camelcase (we should make mandatory over the next 6 months, and mandatory incrementally for files that become compliant)
* no more findDOMNode, since it is scheduled for deprecation from React

Someday we should enable
* react/destructuring-assignment -- this would find sloppy mistakes and avoid dereference errors, but there are 3990 of them as of now.

Latest eslint found a important security vulnerability
* Using ```target="_blank"``` without ```rel="noopener noreferrer"``` is a security risk: see https://mathiasbynens.github.io/rel-noopener and will throw a test error going forward (all existing errors of this type have been fixed)


DaleMcGrew:
I have done an extensive code review. We will need to run the entire system through regression testing to verify no issues were introduced. 👍

PR #1818, opened 2018-12-09
Add marginTop ToastContainer if iOS to contain Toast to application area

SailingSteve:
Nothing needed for Android
Fix for wevote/WeVoteCordova/issues/94
Also fixed missing ballot-election-list__upcoming, which makes a difference for Cordova

PR #1823, opened 2018-12-11
Add marginTop ToastContainer if iOS to contain Toast to application area (Ver2)

SailingSteve:
Replaces PR #1818
Manually rebased
Nothing needed for Android
Fix for wevote/WeVoteCordova/issues/94
Also fixed missing ballot-election-list__upcoming, which makes a difference for Cordova


DaleMcGrew:
👍

PR #1827, opened 2018-12-12
Updated babelify, to stop lint-css error

SailingSteve:
Had to fall back to earlier version of my repository for react-svg-icons
so this brings back the vulnerabilities (dev only) in issues/1693


DaleMcGrew:
👍

Issue #1829, opened 2018-12-18
Remove "Upcoming Elections in STATE" when there are no upcoming elections


DaleMcGrew:
### Please describe the issue (What happens? What do you expect?)
In this screenshot, there is the text "Upcoming Elections in California". That text and all text under it can be removed. Please work with Dale on this Issue.
![noupcomingelections](https://user-images.githubusercontent.com/7756031/50127040-3ef78d80-0224-11e9-85a7-f79911e301db.png)

DaleMcGrew:
This has been fixed and merged, and now needs to be tested.

DaleMcGrew:
Verified!

PR #1830, opened 2018-12-18
[Issue #1829] Remove "Upcoming Elections" when there are no upcoming …

lazorfuzz:
…elections

Issue #1829 
If `props.showRelevantElections` is true but there are no upcoming state elections and no upcoming elections, return empty div.


DaleMcGrew:
Thank you @lazorfuzz! I really appreciate your contribution. 👍

Issue #1755, opened 2018-10-23
Update Credits page to support Company Logos


DaleMcGrew:
### Please describe the issue (What happens? What do you expect?)
Right now we have this page which displays organizations:
WebApp/src/js/routes/More/Credits.jsx

...from "organizationalDonors" (from this file: WebApp/src/js/routes/More/people.js)
Please add "logo" field in the "organizationalDonors" dict, where we can link to logos to be stored in this folder:
WebApp/src/img/global/logos

Please add the "Fastly.com" logo (we already have them in our credits list). Please make sure display works in both mobile and desktop mode.

I am envisioning Name on the left with title underneath that, and then the logo (top aligned with the name) to the right of the Name & Title.

DaleMcGrew:
- [ ] Please update logo display code to prevents the display of a logo image if one doesn't exist
- [ ] Please add a resized version of the We Vote logo, either https://github.com/wevote/DesignInventory/blob/master/WeVoteLogo/PNG/we-vote-logo-wordmark-horizontal-color-on-white-1500x495.png or https://github.com/wevote/DesignInventory/blob/master/WeVoteLogo/PNG/we-vote-logo-wordmark-horizontal-inverted-color-1500x495.png

DaleMcGrew:
Verified!

PR #1834, opened 2018-12-21
[Issue #1755] Add company logos to credits page

lazorfuzz:
Issue #1755 
Added logos for every organization except We Vote Education. Changed styling and formatting.


DaleMcGrew:
Thank you @lazorfuzz! 👍

DaleMcGrew:
Hi @lazorfuzz, I'm wondering if we should add yarn.lock to our gitignore file? It looks like there is a complaint about Mac/Linux incompatibility: https://travis-ci.org/wevote/WebApp/builds/470964781

lazorfuzz:
@DaleMcGrew thanks for pointing that out, that yarn error seems to have revealed a misplaced module. fsevents is a Node module for MacOS and is incompatible for the browser, so it shouldn't be in dependencies in the first place. I'll remove it from package.json in my next PR. I'd recommend keeping the yarn.lock file since it helps to [reduce undeterministic behavior](https://github.com/yarnpkg/yarn/issues/1583#issuecomment-257544217), that is unless it causes further build issues.

PR #1848, opened 2019-01-08
Got Facebook login going again, updated to latest FB api

SailingSteve:
Fixed a typo in the SECURE_CERTIFICATE.MD that showed wrong working dir.

Fixed some potential problems that I had introduced in the stores -- the
dispatcher seems to call all of the stores until it finds a match, so
you can't make assumptions about what is in the action until there is
a type match.

Added a React 16 error boundary to SignIn.js


DaleMcGrew:
👍 Thank you @SailingSteve!

PR #1853, opened 2019-01-11
Change from react-svg-icons to html img and react-svg to remove last npm vulnerabilities

SailingSteve:
(Finally) resolves wevote/WebApp/issues/1693
Most of the <Icon /> components (react-svg-icons) were replaced with simple
html  components which can load SVGs.  In a few places we need to
modify the SVG fill parameter which is inside the SVG file, and for that
we need <ReactSVG /> (react-svg), which does the job.
At this point, just using standard npm packages (no patched packages anymore),
'npm audit' returns zero vulnerabilities.


DaleMcGrew:
Thank you @SailingSteve, great work! 👍

PR #1857, opened 2019-01-15
Fix for issue icons within circles, and color of Facebook icon

SailingSteve:
Also eliminated two npm packages that we don't use (and wouldn't want to)
Added an attributions file to comply with MIT license conditions.


DaleMcGrew:
👍

Issue #1863, opened 2019-01-17
Fix eslint warnings for js/routes/TwitterHandleLanding.jsx


DaleMcGrew:
Camel case updates:
WebApp/src/js/routes/TwitterHandleLanding.jsx

Fix the warnings that you get when you run `npm test`.

/Users/dalemcgrew/PythonProjects/PersonalGitForks/WebApp/src/js/routes/TwitterHandleLanding.jsx
   56:3   warning  _onTwitterStoreChange should be placed after getIncomingActiveRoute         react/sort-comp
   58:11  warning  Identifier 'twitter_followers_count' is not in camel case                   camelcase
   60:7   warning  Identifier 'kind_of_owner' is not in camel case                             camelcase
   60:22  warning  Identifier 'owner_we_vote_id' is not in camel case                          camelcase
   60:40  warning  Identifier 'twitter_handle' is not in camel case                            camelcase
   60:56  warning  Identifier 'twitter_description' is not in camel case                       camelcase
   60:78  warning  Identifier 'twitter_name' is not in camel case                              camelcase
   61:7   warning  Identifier 'twitter_photo_url' is not in camel case                         camelcase
   61:26  warning  Identifier 'twitter_user_website' is not in camel case                      camelcase
   64:16  warning  Identifier 'twitter_followers_count' is not in camel case                   camelcase
   65:7   warning  Identifier 'twitter_followers_count' is not in camel case                   camelcase
   86:33  warning  Expected 'this' to be used by class method 'organizationCreateFromTwitter'  class-methods-use-this
   86:34  warning  Identifier 'new_twitter_handle' is not in camel case                        camelcase
   92:11  warning  Identifier 'incoming_active_route' is not in camel case                     camelcase
   94:12  warning  Identifier 'incoming_active_route' is not in camel case                     camelcase
  106:14  warning  Identifier 'kind_of_owner' is not in camel case                             camelcase
  106:29  warning  Identifier 'owner_we_vote_id' is not in camel case                          camelcase
  106:47  warning  Identifier 'twitter_handle' is not in camel case                            camelcase
  108:11  warning  Identifier 'signed_in_twitter' is not in camel case                         camelcase
  109:9   warning  Identifier 'signed_in_with_this_twitter_account' is not in camel case       camelcase
  110:9   warning  Identifier 'looking_at_positions_for_friends_only' is not in camel case     camelcase
  111:9   warning  Identifier 'signed_in_twitter' is not in camel case                         camelcase
[...]

PR #1865, opened 2019-01-17
Eslint Warnings TwitterHandleLanding #1863

fulin426:
Camel case updates:
WebApp/src/js/routes/TwitterHandleLanding.jsx

convert TwitterHandleLanding to camel case


src/js/routes/TwitterHandleLanding.jsx
```diff
@@ -24,15 +24,15 @@ export default class TwitterHandleLanding extends Component {
...
     };
     this.getIncomingActiveRoute = this.getIncomingActiveRoute.bind(this);
   }
 
   componentDidMount () {
-    // console.log("TwitterHandleLanding componentDidMount, this.props.params.twitter_handle: " + this.props.params.twitter_handle);
-    this.setState({ twitter_handle: this.props.params.twitter_handle });
-    TwitterActions.twitterIdentityRetrieve(this.props.params.twitter_handle);
+    // console.log("TwitterHandleLanding componentDidMount, this.props.params.twitterHandle: " + this.props.params.twitterHandle);
+    this.setState({ twitterHandle: this.props.params.twitterHandle });
```
DaleMcGrew: Hi Fu-Lin, we cannot change the case of incoming this.props.params variables, so this should be reverted to `this.props.params.twitter_handle`. It can be passed into a state variable `this.state.twitterHandle` though.

src/js/routes/TwitterHandleLanding.jsx
```diff
@@ -24,15 +24,15 @@ export default class TwitterHandleLanding extends Component {
...
+    TwitterActions.twitterIdentityRetrieve(this.props.params.twitterHandle);
```
DaleMcGrew: This should be reverted to `this.props.params.twitter_handle` per my earlier comment.

src/js/routes/TwitterHandleLanding.jsx
```diff
@@ -41,10 +41,10 @@ export default class TwitterHandleLanding extends Component {
 
   componentWillReceiveProps (nextProps) {
     // console.log("TwitterHandleLanding componentWillReceiveProps");
-    if (nextProps.params.twitter_handle && this.state.twitter_handle.toLowerCase() !== nextProps.params.twitter_handle.toLowerCase()) {
+    if (nextProps.params.twitterHandle && this.state.twitterHandle.toLowerCase() !== nextProps.params.twitterHandle.toLowerCase()) {
```
DaleMcGrew: This should be reverted to `nextProps.params.twitter_handle` per my earlier comment.

src/js/routes/TwitterHandleLanding.jsx
```diff
@@ -41,10 +41,10 @@ export default class TwitterHandleLanding extends Component {
...
       // We need this test to prevent an infinite loop
-      // console.log("TwitterHandleLanding componentWillReceiveProps, different twitter_handle: ", nextProps.params.twitter_handle);
-      TwitterActions.twitterIdentityRetrieve(nextProps.params.twitter_handle);
+      // console.log("TwitterHandleLanding componentWillReceiveProps, different twitterHandle: ", nextProps.params.twitterHandle);
+      TwitterActions.twitterIdentityRetrieve(nextProps.params.twitterHandle);
```
DaleMcGrew: This should be reverted to `nextProps.params.twitter_handle` per my earlier comment.

src/js/routes/TwitterHandleLanding.jsx
```diff
@@ -55,25 +55,25 @@ export default class TwitterHandleLanding extends Component {
 
   _onTwitterStoreChange () {
     // console.log("TwitterHandleLanding _onTwitterStoreChange");
-    let { twitter_followers_count } = TwitterStore.get();
+    let { twitterFollowersCount } = TwitterStore.get();
     const {
-      kind_of_owner, owner_we_vote_id, twitter_handle, twitter_description,  twitter_name,
-      twitter_photo_url, twitter_user_website, status,
+      kindOfOwner, ownerWeVoteId, twitterHandle, twitterDescription,  twitterName,
+      twitterPhotoUrl, twitterUserWebsite, status,
```
DaleMcGrew: Since these variables come from the store (which is still using lower case variables with underscores), the way the Twitter store holds the variable names needs to be mapped to our local variable. The format seems odd at first because it seems backwards, but it is like this (I just used the first two variables as an example):
```
const {
      kind_of_owner: kindOfOwner, owner_we_vote_id: ownerWeVoteId
    } = TwitterStore.get();
```

src/js/routes/TwitterHandleLanding.jsx
```diff
@@ -103,43 +103,43 @@ export default class TwitterHandleLanding extends Component {
...
     const {
-      voter, kind_of_owner, owner_we_vote_id, twitter_handle,
+      voter, kindOfOwner, ownerWeVoteId, twitterHandle,
     } = this.state;
-    const signed_in_twitter = voter === undefined ? false : voter.signed_in_twitter;
-    let signed_in_with_this_twitter_account = false;
-    let looking_at_positions_for_friends_only = false;
-    if (signed_in_twitter) {
-      const twitter_handle_being_viewed = twitter_handle; // Variable copied for code clarity
+    const signedInTwitter = voter === undefined ? false : voter.signedInTwitter;
```
DaleMcGrew: The voter object is coming from the API server, and the API server uses variables with lower case with underscores. So this line should be:
`const signedInTwitter = voter === undefined ? false : voter.signed_in_twitter;`

DaleMcGrew:
Hi @fulin426, I added several comments, but aside from those this looks good. Can you please submit an update to this pull request?

fulin426:
Thanks for the comments. I'll incoporate them and update this pull request.

I mainly tried to focus on the camel case errors without realizing props
and API objects may still be in snake case. I can see why this refractoring
needs to be done manually rather than automatically.

src/js/routes/TwitterHandleLanding.jsx
```diff
@@ -151,23 +152,23 @@ export default class TwitterHandleLanding extends Component {
           />
         );
       }
-    } else if (this.state.kind_of_owner === "TWITTER_HANDLE_NOT_FOUND_IN_WE_VOTE") {
+    } else if (this.state.kindOfOwner === "TWITTER_HANDLE_NOT_FOUND_IN_WE_VOTE") {
       // console.log("TwitterHandleLanding TWITTER_HANDLE_NOT_FOUND_IN_WE_VOTE calling UnknownTwitterAccount");
       return <UnknownTwitterAccount {...this.state} />;
```
DaleMcGrew: Hello @fulin426, since we are passing this.state into the component UnknownTwitterAccount, you will need to make sure that the incoming props for WebApp/src/js/routes/VoterGuide/UnknownTwitterAccount.jsx get updated to camel case.

src/js/routes/TwitterHandleLanding.jsx
```diff
@@ -55,25 +55,26 @@ export default class TwitterHandleLanding extends Component {
 
   _onTwitterStoreChange () {
     // console.log("TwitterHandleLanding _onTwitterStoreChange");
-    let { twitter_followers_count } = TwitterStore.get();
+    let { twitter_followers_count: twitterFollowersCount } = TwitterStore.get();
```
DaleMcGrew: I had to map `twitter_followers_count` from TwitterStore to the local camel case twitterFollowersCount.

src/js/routes/TwitterHandleLanding.jsx
DaleMcGrew: That is correct @fulin426.

src/js/routes/VoterGuide/UnknownTwitterAccount.jsx
```diff
@@ -17,7 +17,7 @@ export default class UnknownTwitterAccount extends Component {
 
   render () {
     renderLog(__filename);
-    const { twitter_handle: twitterHandle, twitter_name: twitterName } = this.props;
+    const { twitterHandle: twitterHandle, twitterName: twitterName } = this.props;
```
DaleMcGrew: Thank you @fulin426, when deconstructing props or the state, if the variable is already in the format you want it in, no reaassignment is needed. That is, you don't need to write `twitterHandle: twitterHandle`. I will update this now.

DaleMcGrew:
Looks good, thank you @fulin426! 👍

fulin426:
@DaleMcGrew Thanks for helping me with this first component. I'll use what I learned here in the next assignment.

PR #1867, opened 2019-01-18
removed unused functions and updated from snake_case to camelCase

CoffeeWarrior:
removed unused functions and updated from snake_case to camelCase


DaleMcGrew:
👍 Thank you @CoffeeWarrior!

PR #1872, opened 2019-01-20
updated setState functions to use arrow functions and updated snake_case to camelCase

CoffeeWarrior:
updated setState functions to use arrow functions and updated snake_case to camelCase


DaleMcGrew:
👍

PR #1881, opened 2019-01-22
removed warnings and updated state fields on OrganizationVoterTabs.jsx, learnMore.jsx, and readMore.jsx

CoffeeWarrior:
removed warnings and updated state fields on OrganizationVoterTabs.jsx, learnMore.jsx, and readMore.jsx


DaleMcGrew:
Looks good @CoffeeWarrior, thank you.

Issue #1714, opened 2018-10-07
Make sure ballot data shows, and add "Completed" Text: Items Decided + Office/Race Level


DaleMcGrew:
### Please describe the issue (What happens? What do you expect?)
In 2019, we are starting to host special elections that don't have any Federal races. What happens is that when there aren't any Federal races, the page shows up as blank, until the button is clicked.

![screen shot 2019-01-23 at 2 35 29 pm](https://user-images.githubusercontent.com/7756031/51641856-362d6200-1f1c-11e9-8d7a-585d681c97a8.png)

- [ ] If there is a single option, please remove all "Federal", "State" etc. options, as in this case: https://wevote.us/ballot/brooklyn_november_2017
- [ ] If there are 2+ options, but one of them is not "Federal", please make the furthest left option selected, so that the ballot items show on the page: https://wevote.us/ballot/election/4471

We have "Remaining Decisions", "All Items" and "Items Decided".
Then we filter by the Office/Race Level ("Federal", "State", "Local" and "Measure")
![screen shot 2018-10-07 at 12 08 01 pm](https://user-images.githubusercontent.com/7756031/46585757-af667f00-ca29-11e8-9158-b27e79b93cbe.png)

In some pairings, there is no data to show:
![screen shot 2018-10-07 at 12 08 32 pm](https://user-images.githubusercontent.com/7756031/46585761-c2794f00-ca29-11e8-8320-fc5803e2352f.png)

For each of these cases where there isn't an office or measure shown, add text that a) describes why they aren't seeing an office or measure, and b) what they can do. For example, in the case where "Items Decided" has been chosen, with the "Measure" filter:
"You have not decided on any measures yet. Next steps:

- Click on "Remaining Decisions" to see the Measures you need to decide on.
- Click on "Federal" or "State" to see candidates you have chosen."

DaleMcGrew:
Hi @lazorfuzz the pull request you created did some excellent and critical work (in making sure that the measure data shows up on this page: http://localhost:3000/ballot/brooklyn_november_2017). Thank you! 

The next step is to make sure the single "Measure" button goes away when there is only one option. I found this on line 955 of WebApp/src/js/routes/Ballot/Ballot.jsx
...with the div that includes this classname: `ballot__item-filter-tabs`

DaleMcGrew:
Great work @lazorfuzz!

PR #1887, opened 2019-01-25
[Issue #1714]

lazorfuzz:
#1714 
If there are no Federal ballot items, show the next available items (State, Measure, Local).

### DIFF
--- a/src/js/routes/Ballot/Ballot.jsx
+++ b/src/js/routes/Ballot/Ballot.jsx
@@ -53,6 +53,8 @@ import webAppConfig from "../../config";
 /* eslint no-param-reassign: 0 */
 
 // Related to WebApp/src/js/components/VoterGuide/VoterGuideBallot.jsx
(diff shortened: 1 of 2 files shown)


lazorfuzz:
@DaleMcGrew, there were some parts of the issue that I wasn't able to get to. In the case of the brooklyn_november_2017 election, there are only 3 measure items shown but there are a total of 20 items. I'm not sure why 17 of the ballot items are not shown, but I think it's because they don't have a race_office_level property:

image

In the above image we can see that the statewide proposals (Measures) show properly but the mayoral election and everything after that do not have race_office_levels. Is that an issue on the server side or is there a reason why their race_office_levels are null?

DaleMcGrew:
Thank you @lazorfuzz! Good catch. The lack of data in the race_office_levels variable is a data problem on the API server, and I believe we have fixed that with more recent elections. We can test that against more recent elections.

PR #1894, opened 2019-01-26
[Issue #1714] Show 1st available ballot items when user changes address

lazorfuzz:
#1714 
* Reset completionLevelFilter back to 'filterAllBallots' when the user changes their address (doing this in case the user had the filter on "Decided," seemed to make sense from a UX standpoint)
* Moved the default race filter check from onBallotStoreChange to componentDidUpdate


DaleMcGrew:
👍

PR #1899, opened 2019-01-26
Translate SettingsMenuMobile to use camel case.

Tezaen:
#1850

Convert routes/Settings/SettingsMenuMobile.jsx to use camel case.

### DIFF
--- a/src/js/routes/Settings/SettingsMenuMobile.jsx
+++ b/src/js/routes/Settings/SettingsMenuMobile.jsx
@@ -31,7 +31,6 @@ export default class SettingsMenuMobile extends Component {
   }
 
   componentDidMount () {
-    console.log("In settings menu mobile. Component did mount!!!");
     if (this.props.params.edit_mode) {
       this.setState({ editMode: this.props.params.edit_mode });
     } else {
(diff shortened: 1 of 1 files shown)


utsab:
@DaleMcGrew, we were not sure about the meaning of voter.linked_organization_we_vote_id.    Does this mean that a user's account is somehow linked to a specific organization?  What part of the UI would allow us to set voter.linked_organization_we_vote_id?

DaleMcGrew:
👍 Thank you @Tezaen for your first pull request to We Vote! These changes look great. 

@utsab, to answer your question: voter.linked_organization_we_vote_id is how we link a voter to an organization. Every voter has an organization object in case they want to push any of their opinions public. All publicly visible opinions must be attached to an organization object.

utsab:
@DaleMcGrew, thanks for explaining that.  Where would I go in the website to link myself to an organization? I went to the Settings page, but I wasn't sure where to go from there.

DaleMcGrew:
Hi @utsab, this is unfortunately a file we can only test on our Cordova app, so there is nothing more to test here.

PR #1924, opened 2019-02-03
Did ESLint cleanup on 121 files (including deleting unused files). We are now down to 253 warnings.

DaleMcGrew:
Did ESLint cleanup on 121 files (including deleting unused files). We are now down to 253 warnings.


DaleMcGrew:
Hi @SailingSteve, I have fixed the Twitter Sign in problem and will be checking in the change before Noon.

PR #1925, opened 2019-02-04
AddressBox.jsx removed some warnings, fixed redirect bug

CoffeeWarrior:
AddressBox.jsx removed some warnings, fixed redirect bug


CoffeeWarrior:
reverting these changes and updating in a different manner

DaleMcGrew:
👍 Thank you @CoffeeWarrior!

PR #1930, opened 2019-02-06
fixed the warnings

uroshima:
This fix all the warnings named "Unexpected unnamed function". There were 10 total


uroshima:
Created new pull request

Issue #1949, opened 2019-02-18
Address & Elections Popover opened when you scroll down page


DaleMcGrew:
### Please describe the issue (What happens? What do you expect?)
1. On the ballot page click the "Address & Elections" button, so that the popover opens
2. Close this popover
3. Scroll down the page and the popover opens again when it shouldn't (without clicking on the "Address & Elections" button

DaleMcGrew:
Fixed.

PR #1950, opened 2019-02-18
[Issue #1949] Fix address modal showing when scrolled

lazorfuzz:
#1949 
* Fixed the address modal showing after clicking on it and then scrolling
* Changed some alerts/notifications to use Snackbars instead

Example:
image

It disappears on its own after 5 seconds.


DaleMcGrew:
👍

Issue #1962, opened 2019-02-22
Convert candidates display to horizontal cards in Desktop


DaleMcGrew:
### Please describe the issue (What happens? What do you expect?)
In this file, WebApp/src/js/components/Ballot/OfficeItemCompressed.jsx

...update the display of the candidate cards in Desktop (NOT mobile -- the designs should remain as they currently are in mobile) to be in a left-to-right grid like this:
<img width="700" alt="screen shot 2019-02-22 at 11 00 30 am" src="https://user-images.githubusercontent.com/7756031/53265183-6c344200-3692-11e9-9b41-d83bc57ed0d3.png">

DaleMcGrew:
Good morning @lazorfuzz, I really appreciate your implementation of the horizontal cards. You have achieved the designers' vision! Thank you.

PR #1990, opened 2019-03-07
[Issue #1962] First pass at office items

lazorfuzz:
#1962 
* Converted office items to new designs

@DaleMcGrew I made each candidate take up a row on desktop, because having two candidates on a row could cause problems when they have a long name. If you'd like me to keep at two candidates per row, let me know. Thanks!

### DIFF
--- a/src/js/components/Ballot/OfficeItemCompressed.jsx
+++ b/src/js/components/Ballot/OfficeItemCompressed.jsx
@@ -1,10 +1,11 @@
 import React, { Component } from 'react';
 import PropTypes from 'prop-types';
 import {  Modal } from 'react-bootstrap'; // , OverlayTrigger, Popover
-import { Link } from 'react-router';
 // import TextTruncate from 'react-text-truncate';
 import Slider from 'react-slick';
 import styled from 'styled-components';
+import { withTheme, withStyles } from '@material-ui/core/styles';
+import ArrowForwardIcon from '@material-ui/icons/ArrowForward';
 import BallotItemSupportOpposeCountDisplay from '../Widgets/BallotItemSupportOpposeCountDisplay';
 import { cordovaDot, historyPush, hasIPhoneNotch } from '../../utils/cordovaUtils';
 import { toTitleCase } from '../../utils/textFormat';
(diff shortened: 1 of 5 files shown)


src/js/components/Ballot/OfficeItemCompressed.jsx
```diff
@@ -388,56 +391,56 @@ class OfficeItemCompressed extends Component {
...
                         {/* Candidate Name */}
                         <div>
-                          <Link to={this.getCandidateLink(oneCandidate.we_vote_id)}>
-                            <h4 className="card-main__candidate-name card-main__candidate-name-link u-f5">
-                              {oneCandidate.ballot_item_display_name}
-                              <br />
-                              <span className="card-main__candidate-party-description">{candidatePartyText}</span>
-                            </h4>
-                          </Link>
+                          <h4 className="card-main__candidate-name card-main__candidate-name-link u-f5">
```
DaleMcGrew: Hi Leon, I haven't tested this code yet, but I believe we still need the Links in place here.

src/js/components/Widgets/BallotItemSupportOpposeCountDisplay.jsx
```diff
@@ -619,7 +619,7 @@ class BallotItemSupportOpposeCountDisplay extends Component {
             placement={this.props.popoverBottom ? 'bottom' : 'top'}
             overlay={scoreInYourNetworkPopover}
           >
-            <span className="network-positions-stacked__support-score u-cursor--pointer u-no-break">
+            <NetworkScore>
```
DaleMcGrew: Hi @lazorfuzz, I'd love to discuss the changes like this that are moving away from using WebApp/src/sass to using styles embedded in the components. The biggest problem I see is that with styled divs we don't currently have a way to use standardized variables, which we can (and use heavily) within our WebApp/src/sass files. An example is defining our breakpoints in components, like this: @media (max-width: 768px) {

DaleMcGrew:
👍 Thank you @lazorfuzz! I look forward to discussing when we get a chance.

Issue #1960, opened 2019-02-22
Add right arrow to office name in OfficeItemCompressed


DaleMcGrew:
### Please describe the issue (What happens? What do you expect?)
In this file: WebApp/src/js/components/Ballot/OfficeItemCompressed.jsx

...add an arrow to the right of the display of ballotItemDisplayName, and make the whole like show up as black, per this design:
<img width="280" alt="screen shot 2019-02-22 at 10 42 43 am" src="https://user-images.githubusercontent.com/7756031/53263678-fe3a4b80-368e-11e9-8b0b-c523e3595b17.png">

Please leave the link underline on rollover, as well as changing the mouse to a pointer finger.

DaleMcGrew:
Hello @uroshima I just wanted to flag that this Issue is still open.

DaleMcGrew:
Resolved. Thank you @uroshima!

Issue #1961, opened 2019-02-22
Style the "view all x candidates" and "show more" links


DaleMcGrew:
### Please describe the issue (What happens? What do you expect?)
- [x] Please style the "view all x candidates" and "show more" links in WebApp/src/js/components/Ballot/OfficeItemCompressed.jsx:
![screen shot 2019-02-22 at 10 47 38 am](https://user-images.githubusercontent.com/7756031/53263887-83bdfb80-368f-11e9-8f64-76d03c3f9e75.png)

...including adding the right arrow, per this design:
<img width="296" alt="screen shot 2019-02-22 at 10 50 42 am" src="https://user-images.githubusercontent.com/7756031/53264085-ba941180-368f-11e9-9d33-8aa2b3c9c5e9.png">

- [x] Please also add "show more ->" to the bottom of each candidate on the Office Detail page.

DaleMcGrew:
Hello @uroshima, the next step for this Issue is to migrate the "ShowMoreFooter" component to all places where "Show more" text is shown (including for the Measures). Also, did you check in your "hover over" code that you showed to Ahmed?

DaleMcGrew:
Resolved!

PR #1991, opened 2019-03-07
Styled the "view all x candidates" and "show more" links and also added right arrow to office name in OfficeItemCompressed

uroshima:
This fixes the issues #1960 and #1961


DaleMcGrew:
👍

Issue #1963, opened 2019-02-22
Add office name to the sticky header


DaleMcGrew:
### Please describe the issue (What happens? What do you expect?)
On the office page please add the office name to the sticky header like this in mobile:
<img width="295" alt="screen shot 2019-02-22 at 11 23 36 am" src="https://user-images.githubusercontent.com/7756031/53266017-72c3b900-3694-11e9-9f0d-3d60b1e2236a.png">

...and the same in desktop:
<img width="926" alt="screen shot 2019-02-22 at 11 25 42 am" src="https://user-images.githubusercontent.com/7756031/53266081-9dae0d00-3694-11e9-81ae-db26381914b0.png">

DaleMcGrew:
Hi @flurmbo, this is a great first pass! (I will run this by our design team.) I notice when I start on this candidate page now (https://localhost:3000/candidate/wv02cand55907/b/btdo/), I get this error:
![Screen Shot 2019-03-10 at 11 45 18 AM](https://user-images.githubusercontent.com/7756031/54089821-65f8c380-432a-11e9-8f2e-5b083777ccda.png)

...and if I start on this office page (https://localhost:3000/office/wv02off28318/b/btdb/), I get several console errors:
![Screen Shot 2019-03-10 at 11 50 18 AM](https://user-images.githubusercontent.com/7756031/54089850-b4a65d80-432a-11e9-8f2d-020c7ee93f93.png)

flurmbo:
@DaleMcGrew good catch, I'll take a look at those

DaleMcGrew:
Thank you @flurmbo!

PR #1998, opened 2019-03-11
Fix console warnings introduced in previous PR

flurmbo:
Fixes the warnings @DaleMcGrew noticed from my previous PR for #1963.

1) Renders the OfficeItem component in the HeaderBackToBallot more safely by only rendering it when the path contains '/office' and the state has an officeName.

2) Retreives the officeName more safely from the OfficeStore. It looks like it's possible for the OfficeStore to not be able to retrieve an office object for an officeWeVoteId even when the HeaderBackToBar already has that id in its props. That makes me think there could be a deeper issue going on here, though I'm not sure. In the meantime, I just verified that the OfficeStore returns an object before getting the officeName property off it.


DaleMcGrew:
👍

Issue #2000, opened 2019-03-13
Ballot List: When one candidate, take up full width


DaleMcGrew:
### Please describe the issue (What happens? What do you expect?)
- [x] When on the ballot list in Desktop or Tablet, if there is only one candidate, have the card take up the full width.
- [x] When on the ballot list in Desktop or Tablet, if there are more than two candidates, have the third and fourth candidates wrap to the next line. If there are only three candidates, the third candidate card should *not* widen to fill the full width, but stay in the first column with the first candidate.

<img width="879" alt="Screen Shot 2019-03-13 at 7 15 55 AM" src="https://user-images.githubusercontent.com/7756031/54286290-c709dc80-4560-11e9-9b52-b85a63d7c084.png">

DaleMcGrew:
Thank you @lazorfuzz!

PR #2004, opened 2019-03-14
[Issue #2000] Fix candidate width

lazorfuzz:
#2000 
When one candidate, take up full width

### DIFF
--- a/src/js/components/Ballot/BallotSearch.jsx
+++ b/src/js/components/Ballot/BallotSearch.jsx
@@ -132,7 +132,7 @@ const SearchWrapper = styled.div`
   border: 1px solid ${props => (!props.searching ? 'rgba(0, 0, 0, 0.23)' : props.brandBlue)};
   padding: 0 3px 0 3px;
   margin-right: 16px;
-  @media (max-width: 960px) {
+  @media (max-width: ${({ theme }) => theme.breakpoints.lg}) {
     height: 22.5px;
     margin-right: 4px;
   }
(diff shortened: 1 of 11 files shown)


src/js/components/Ballot/BallotSearch.jsx
```diff
@@ -132,7 +132,7 @@ const SearchWrapper = styled.div`
   border: 1px solid ${props => (!props.searching ? 'rgba(0, 0, 0, 0.23)' : props.brandBlue)};
   padding: 0 3px 0 3px;
   margin-right: 16px;
-  @media (max-width: 960px) {
+  @media (max-width: ${({ theme }) => theme.breakpoints.lg}) {
```
DaleMcGrew: Hi @lazorfuzz, I am going to approve this pull request as-is, but over time could you please give these breakpoints full names that match the ones we use for our sass variables? I have learned that short variables over time can cause confusion for engineers doing maintenance on a system, and full variable names that provide more context help a lot. (ex/ instead of "lg", something like "tablet")

src/js/components/Ballot/OfficeItemCompressed.jsx
```diff
@@ -403,8 +403,8 @@ class OfficeItemCompressed extends Component {
           {/* *************************
             Display either a) the candidates the voter supports, or b) the first several candidates running for this office
             ************************* */}
-          <Container>
-            { this.state.candidateList.map((oneCandidate) => {
+          <Container candLength={candidateList.length}>
```
DaleMcGrew: Please use full-word variables. "candidateLength" instead of "candLength". I go into more depth regarding why here: https://github.com/wevote/WebApp/blob/develop/docs/contributing/CONTRIBUTING_STANDARDS.md

DaleMcGrew:
That is awesome @lazorfuzz  that you are implementing max width variables! Very exciting. Thank you. 👍

Issue #1995, opened 2019-03-09
When signed in, Settings and Address icon disappear from header


DaleMcGrew:
In the header, when you sign in, the Settings and Address icons disappear, and they should remain.

Not signed in on ballot page:
<img width="233" alt="Screen Shot 2019-03-09 at 10 53 14 AM" src="https://user-images.githubusercontent.com/7756031/54075913-958bcb00-4259-11e9-8bdd-189401c28fdf.png">

Signed in on ballot page:
<img width="129" alt="Screen Shot 2019-03-09 at 10 55 46 AM" src="https://user-images.githubusercontent.com/7756031/54075943-edc2cd00-4259-11e9-9861-0a0b2ccb6412.png">

flurmbo:
@DaleMcGrew I've got something that works on the ballot page, should I try factoring it out into a separate component and add it to the "Back to X" components?

DaleMcGrew:
Hi @flurmbo, I don't think we need the Settings and Address icons on the "Back to" pages.

flurmbo:
@DaleMcGrew I PRed my changes, let me know what you think

DaleMcGrew:
Thank you @flurmbo!

PR #2010, opened 2019-03-18
[#1995] show Settings and Address icons on header when logged in

flurmbo:
#1995
When `showFullNavigation && voterIsSignedIn` is true, show the settings and address icons.
`header-nav__avatar`, `header-nav__avatar-cordova` and `header-nav__avatar-container` now have `display: inline` CSS rule. Hardcoded width and height values have been removed because inline elements ignore these style rules.


DaleMcGrew:
👍 Thank you Phil!

PR #2012, opened 2019-03-19
Updated prop-types library to eliminate runtime exception

SailingSteve:
Removed FooterBarCordova, it has been replaced with the newer FooterBar.jsx
Removed outdated lint overrides in Ballot.jsx


DaleMcGrew:
👍

PR #2014, opened 2019-03-19
Basic BrowserStack testing folder

flurmbo:
N/A

@DaleMcGrew, let me know how this looks and if there are any tweaks to the file structure, etc. should make.

One change that might be useful to do might be to add a testing framework like MochaJS to run the tests. Also, right now the test always runs on Chrome on Windows 7, so we might want to pull that out into the config file.


DaleMcGrew:
Hi @flurmbo, this is great! I'll reach out separately about your question about the kind of device configuration. 👍

PR #2018, opened 2019-03-20
Fixed "I represent ...." warning issue

ltsinh:
In the I represent section for the candidates, there was an error:
Encountered two children with the same key. `[object Object]`. Keys should be unique so that components maintain their identity across updates. Non-unique keys may cause children to be duplicated and/or omitted — the behavior is unsupported and could change in a future version.

I gave the keys unique values.

The file parsedTwitterDescription.jsx was revised.


DaleMcGrew:
👍

PR #2056, opened 2019-04-12
Fix error in console from withStyles() export on location guess

jarodpeachey:
Fixed the error in the console from incorrect export of component with styles.

### DIFF
--- a/src/js/routes/Ballot/LocationGuess.jsx
+++ b/src/js/routes/Ballot/LocationGuess.jsx
@@ -79,4 +79,4 @@ const CloseComponent = styled.div`
   opacity: 0.5;
 `;
 
-export default withTheme()(withStyles()(LocationGuess));
+export default (withTheme)(withStyles)(LocationGuess);


DaleMcGrew:
👍

lazorfuzz:
@DaleMcGrew @jarodpeachey @uroshima this is still incorrect. Because in the Location component we are not using any Material theme properties or style overrides, neither withTheme nor withStyles are needed.

`export default Location;` will work just fine.

In the future, if we are using Material theme variables or style overrides, then the proper way to export is as follows: `export default withTheme()(withStyles(styles)(ComponentName));`.

Please take a look at `components/Ballot/BallotSearch.jsx` for a proper example on using Material theme and style overrides, and how components using them should be exported.

jarodpeachey:
Thanks for the tip, @lazorfuzz.  I just checked out the BallotSearch.jsx file to get the proper way to do it.  I'll make sure to use that method next time.

Just to double-check, should I edit the location guess file to use `export default LocationGuess;`?

lazorfuzz:
@jarodpeachey No worries, I'm refactoring the component now, and I'll include that change. It turns out withStyles is actually needed because there was some styling done for the Material Place icon. It was done with inline styles, but I'm changing it to style overrides because we want to avoid inline styles like the plague.

uroshima:
Thanks @lazorfuzz, I'll make sure to use that method next time.

Issue #2005, opened 2019-03-14
Change Style of "Back to" buttons


DaleMcGrew:
### Please describe the issue (What happens? What do you expect?)
This is the current style of the "Back to" button:
<img width="224" alt="Screen Shot 2019-03-14 at 3 27 44 PM" src="https://user-images.githubusercontent.com/7756031/54395844-328da000-466e-11e9-8e72-9e9f1cd2adda.png">

We want to upgrade the style to look like this:
<img width="200" alt="Screen Shot 2019-03-14 at 3 32 33 PM" src="https://user-images.githubusercontent.com/7756031/54395927-784a6880-466e-11e9-89b3-3325128338a2.png">

These buttons live in the "HeaderBackTo..." files like this one:
WebApp/src/js/components/Navigation/HeaderBackToBallot.jsx

flurmbo:
Quick question about this @DaleMcGrew, do you think I should use the [arrow_back_ios](https://material.io/tools/icons/?search=arrow_back_ios&icon=arrow_back_ios&style=baseline) material icon for this? According to the [material design website](https://material.io/design/navigation/understanding-navigation.html) it seems like for android apps and web apps [this style of arrow](https://material.io/tools/icons/?search=arrow_back&icon=arrow_back&style=baseline) is preferred. Not sure how important this is but I figured it's worth mentioning

DaleMcGrew:
Thank you for digging into this @flurmbo. Why don't we use the suggested "arrow_back" icon version as the default, but could you include the "arrow_back_ios" icon in that component with an if-statement for "if iOS", so we can easily use the iOS arrow on the Mac and iOS? (We can figure out how to detect that the voter is on Mac or iOS in the near future.)

flurmbo:
Hey @DaleMcGrew, it looks like I'm going to need to override some Material UI styles. Do you know what is our preferred method to do that? I have been able to do it with the class property as described [here](https://material-ui.com/customization/overrides/#overriding-with-classes), but I've notcied that sometimes overrides are done in WebApp with "!important."

DaleMcGrew:
Hi @flurmbo this new "back to" look is awesome! Thank you for implementing. Great work. Verified.

PR #2059, opened 2019-04-14
Put HeaderBackToButton in more files

flurmbo:

#2005 

I've added the HeaderBackToButton component to the HeaderBackToValues, HeaderBackToFriends, and HeaderBackToVoterGuides. I did not add it to HeaderBackToSettings because when poking around my build I was unable to get that component to render visibly. Any instructions on how to view that component would be much appreciated

### DIFF
--- a/src/js/components/Navigation/HeaderBackToFriends.jsx
+++ b/src/js/components/Navigation/HeaderBackToFriends.jsx
@@ -5,12 +5,13 @@ import Toolbar from '@material-ui/core/Toolbar';
 import Button from '@material-ui/core/Button';
 import { withStyles } from '@material-ui/core/styles';
 import cookies from '../../utils/cookies';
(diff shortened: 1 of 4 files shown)


DaleMcGrew:
@flurmbo I think you can get to the back to settings header if you sign in with Twitter.

flurmbo:
@DaleMcGrew, after logging in with Twitter and going to the settings page, what should I click on next so the backto header appears? Still having difficulty with this

DaleMcGrew:
Hi @flurmbo, navigate to this page: http://localhost:3000/settings/account and then put your browser into mobile mode. It looks like we only have the "back to" for mobile devices.

flurmbo:
Ah that makes sense now

flurmbo:
@DaleMcGrew just pushed the back to settings change, this should be done now

DaleMcGrew:
Thank you @flurmbo! 👍

PR #2065, opened 2019-04-15
Create "Add Endorsements" Promo Box

jarodpeachey:
#2063 
Added the endorsements component to ballot and values pages.

### DIFF
--- a/src/js/components/Widgets/EndorsementCard.jsx
+++ b/src/js/components/Widgets/EndorsementCard.jsx
@@ -1,0 +1,30 @@
+import React, { Component } from 'react';
+import { Button } from '@material-ui/core';
+import OpenExternalWebSite from '../../utils/OpenExternalWebSite';
+
+class EndorsementCard extends Component {
+  render () {
+    return (
+      <div>
+        <div className="card">
+          <div className="card-main">
+            <div className="endorsement">
+              <OpenExternalWebSite
+                url="https://api.wevoteusa.org/vg/create/"
+                className="opinions-followed__missing-org-link"
+                target="_blank"
+                title={this.props.title}
+                body={<Button className="btn endorsement__btn btn-sm" bsPrefix={this.props.bsPrefix} variant="primary">{this.props.buttonText}</Button>}
+              />
+              <div className="endorsement__text">
+                {this.props.text}
+              </div>
+            </div>
+          </div>
+        </div>
+      </div>
+    )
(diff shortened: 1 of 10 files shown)


DaleMcGrew:
Thank you @jarodpeachey. I am going to approve this pull request now, but please refer to the svg icon file img/global/svg-icons/positions-icon-24-x-24.svg with an img tag instead of including the svg text directly in the files. 👍

src/sass/components/_network.scss
```diff
@@ -51,6 +60,49 @@ $mobile-btn-padding: 52px;
   }
 }
 
+.endorsement {
```
DaleMcGrew: When naming styles, think in terms of "searchability". With a system as big as We Vote, I try to use names that are unique and can be found easily. So instead of using a style named "endorsement", please use something like "endorsement-card".

src/js/components/Widgets/EndorsementCard.jsx
```diff
@@ -0,0 +1,51 @@
+import React, { Component } from 'react';
+import PropTypes from 'prop-types';
+import { Button } from '@material-ui/core';
+import OpenExternalWebSite from '../../utils/OpenExternalWebSite';
+
+class EndorsementCard extends Component {
```
DaleMcGrew: I've been doing some thinking about this, and I think it would be better to refactor this EndorsementCard component to use `import { Button } from 'react-bootstrap';` and use the same approach that you do in Widgets/AddEndorsements.jsx

This will allow us to maintain the same look and feel for both components.

src/js/components/Widgets/EndorsementCard.jsx
```diff
@@ -0,0 +1,51 @@
+import React, { Component } from 'react';
```
DaleMcGrew: For a simple component like this, you can use `PureComponent` instead of `Component` so that it doesn't re-render each time new (but identical) props get passed in.

src/js/components/Widgets/EndorsementCard.jsx
DaleMcGrew: Also, we have standardized on this approach to taking in proptypes, inside the component:
```
  static propTypes = {
    buttonText: PropTypes.string,
    title: PropTypes.string,
    bsPrefix: PropTypes.string,
    text: PropTypes.string,
  };
```

Issue #2034, opened 2019-03-29
Measure Detail Page - BallotItemSupportOpposeComment component doesn't show up on scroll up


lazorfuzz:
### Please describe the issue (What happens? What do you expect?)

The support/oppose/comment component should reappear on the Measure page when the measure sticky header disappears.  In React dev tools, the component appears to be rendered, but the component doesn't show up in the DOM. The logic to appear/disappear the component is in MeasureItem.jsx.

### Steps to reproduce the problem (1, 2, 3...), including links
Test link:
https://localhost:3000/measure/wv02meas604/b/btdb/

1. Scroll down, until measure sticky header appears.
2. Scroll back up until the sticky header disappears.

DaleMcGrew:
Verified.

PR #2070, opened 2019-04-17
Fix measureHeader on scroll up

jarodpeachey:
#2034 
I removed the switch statement at line 145 in MeasureItem.jsx.  It checked if the page was scrolled down or not, and displayed the comment box depending on that switch.  However, I realized that the comment box can just be a static element, as it will scroll off the page and back on when the popover header appears and disappears. 

Don't know if this is what you want, but that's my best shot.

### DIFF
--- a/src/js/components/Ballot/MeasureItem.jsx
+++ b/src/js/components/Ballot/MeasureItem.jsx
@@ -26,7 +26,7 @@ class MeasureItem extends Component {
       stateCode: '',
       stateDisplayName: '',
       showPositionStatementActionBar: true,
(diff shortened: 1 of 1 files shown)


DaleMcGrew:
Hi @jarodpeachey, I think that makes sense as a solution, thank you. Please run `npm test` to check for eslint errors with each pull request. For example this pull request throws this error:
```
/home/travis/build/wevote/WebApp/src/js/components/Ballot/MeasureItem.jsx
  98:7  error  'scrolledDown' is assigned a value but never used  no-unused-vars
```

DaleMcGrew:
👍

PR #2073, opened 2019-04-19
Check if header should retreat

lazorfuzz:
Makes sure the pathname starts with `/ballot`, `/candidate`, or `/measure` before retreating.


DaleMcGrew:
👍

PR #2079, opened 2019-04-22
Update friends navigation

jarodpeachey:

Previously, the friends tab did not click to the page if you didn't click on the middle of the tab. This was because the onClick() function was nested inside the badge element.  I moved the onClick() outside of the badge element, as with the other tabs, and now the full tab is clickable.


DaleMcGrew:
👍

Issue #2046, opened 2019-04-08
Text showing twice in header button in when width is 319px or less


flurmbo:
### Please describe the issue (What happens? What do you expect?)

This is an issue I've encountered while working on #2005. Using Firefox, when I resize my window on an office page, the word "ballot" will appear twice if I resize the window small enough as in the following gif:

![wevoteballotresize](https://user-images.githubusercontent.com/35127144/55756992-7929ac80-5a18-11e9-9ef0-22bbb5b1cd64.gif)

The issue seems to stem from the two spans [here](https://github.com/wevote/WebApp/blob/a190238fc47a498918602763b03d5e90119adf1d/src/js/components/Navigation/HeaderBackToBallot.jsx#L332) both being visible. The classes that should be handling this, `u-show-desktop-tablet` and `u-show-mobile`, are defined using the breakpoints sass mixin defined [here](https://github.com/wevote/WebApp/blob/a190238fc47a498918602763b03d5e90119adf1d/src/sass/mixins/_breakpoints.scss#L22). Couldn't warp my head around what's going wrong, so I figured I'd post an issue here!

DaleMcGrew:
Thank you @flurmbo for your awesome clarity on this issue, including the animated gif. It looks like with a screen width smaller than iPhone5 (our smallest supported device), the visibility limitations are lifted. I think you are right that we need to look into tweaking the style `u-show-desktop-tablet` to not allow visibility for very small screens.

DaleMcGrew:
Thank you Jarod for tackling this issue. Please do not mark issues as closed, but instead label them "Verify Fix" so QA can verify them before closing them.

DaleMcGrew:
I am able to reproduce this bug in Chrome if you use the "Responsive" display:
![Screen Shot 2019-04-22 at 11 27 38 AM](https://user-images.githubusercontent.com/7756031/56517392-b3ec1400-64f1-11e9-9d1b-7e59aa8da618.png)

jarodpeachey:
For a solution, should I just change the 'small' variable breakpoint in _variables.scss to 100 pixels or so?  That should fix it.

DaleMcGrew:
Hi @jarodpeachey I think the solution would be to adjust `u-show-desktop-tablet` and `u-show-desktop` in:
/WebApp/src/sass/utilities/_u-layout.scss

...to "display: none" when smaller than "small". You might need to add a lower end breakpoint called `width-zero` with a setting of "0".

[...]

PR #2081, opened 2019-04-22
Fix ballot showing twice on mobile xsmall

jarodpeachey:
#2046


DaleMcGrew:
👍

Issue #2038, opened 2019-04-01
Turn off "All Items" tab until choices have been made


DaleMcGrew:
### Please describe the issue (What happens? What do you expect?)
Prior to making any choices, we currently see this, and we want to NOT display "All Items" when it is the only option:
![Screen Shot 2019-04-01 at 2 10 12 PM](https://user-images.githubusercontent.com/7756031/55359926-e45b0800-5487-11e9-8df9-3913ffafcd13.png)

Once you make choices, we want the options tabs to appear.
![Screen Shot 2019-04-01 at 2 08 31 PM](https://user-images.githubusercontent.com/7756031/55359878-c68da300-5487-11e9-8438-9117daa8ec2d.png)

DaleMcGrew:
Verified!

PR #2082, opened 2019-04-23
conditionally render BallotTabsRaccoon

flurmbo:
#2038 
Very simple solution, just check if there are a positive number of remaining choices and that this is different than the ballot length. If that condition holds, render the BallotTabsRaccoon (and also adjust the padding on the ballot body)


DaleMcGrew:
👍

flurmbo:
@DaleMcGrew out of curiosity what did this component have to do with Raccoons ever? I googled it but couldn't find any connection between raccoons and web design

Issue #2020, opened 2019-03-22
OfficeDetail: Hover behavior (1. Values, 2. Endorsements count and 3. rest of card clickable)


DaleMcGrew:
### Please describe the issue (What happens? What do you expect?)
On the office detail, we want *most* of the card to be clickable (to take voter to the Candidate Detail page), and turn link-color blue when you hover over it. Hover over most of the candidate card should also cause the "Choose" and "Oppose" buttons to appear, and be clickable.

There would be two exceptions:
1. Values Icons + text: When hovering over the values, do *not* cause the whole candidate card to turn link-color blue. We will be adding popovers to each value.
2. Endorsements: When hovering over the endorsement counts, do *not* cause the whole candidate card to turn link-color blue. InsteWe will be adding a popover explaining the endorsement counts.

This design image is slightly incorrect -- the values icons + text would not be link-blue, based on where the pointer is over the card.
<img width="491" alt="Screen Shot 2019-03-22 at 1 36 35 PM" src="https://user-images.githubusercontent.com/7756031/54852461-2f1b9980-4caa-11e9-9cc8-aad1ad21c2fa.png">

DaleMcGrew:
Verified, thank you @jarodpeachey!

PR #2089, opened 2019-04-25
Add hover state to office detail candidate card

jarodpeachey:
#2020 
I added a hover to the state, and toggled between true and false on the hover of the card.  I use a prop to the BallotItemSupportOpposeCountDisplay and  IssuesByBallotItemDisplayList components in order to turn off the hover state when they were hovered.  Normal onMouseOver and onMouseEnter did not work without passing as a prop.

Let me know if this is a good solution.

### DIFF
--- a/src/js/components/Ballot/CandidateItem.jsx
+++ b/src/js/components/Ballot/CandidateItem.jsx
@@ -126,13 +126,13 @@ class CandidateItem extends Component {
   }
 
   render () {
-    const handleHover = (e) => {
-      console.log('Handle hover', e.target);
+    const handleHover = () => {
+      // console.log('Handle hover', e.target);
       this.setState({ hover: true });
     };
 
-    const handleLeave = (e) => {
-      console.log('Handle leave', e.target);
+    const handleLeave = () => {
+      // console.log('Handle leave', e.target);
       this.setState({ hover: false });
     };
 
@@ -177,9 +177,9 @@ class CandidateItem extends Component {
     return (
       <div
         className={this.state.hover ? (
-          "card-main candidate-card card-main--outline"
+          'card-main candidate-card card-main--outline'
         ) : (
-          "card-main candidate-card"
+          'card-main candidate-card'
         )}
         onMouseEnter={handleHover}
         onMouseLeave={handleLeave}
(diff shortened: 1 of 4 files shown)


DaleMcGrew:
Hi @jarodpeachey I think changing a state when the subcomponents are hovered over is a great way to do it. Could you please run `npm test` locally and fix the errors? The continuous integration Travis CI tests failed: https://travis-ci.org/wevote/WebApp/builds/524259886?utm_source=github_status&utm_medium=notification

Thank you!

DaleMcGrew:
👍 Thank you @jarodpeachey!

src/js/components/Ballot/CandidateItem.jsx
```diff
@@ -125,6 +126,16 @@ class CandidateItem extends Component {
   }
 
   render () {
+    const handleHover = () => {
```
DaleMcGrew: I would move these functions handleHover and handleLeave out of the render. In the ReactJS world, calling setState directly within the render function is frowned upon.

src/js/components/Values/IssuesByBallotItemDisplayList.jsx
```diff
@@ -155,7 +157,10 @@ class IssuesByBallotItemDisplayList extends Component {
     );
 
     return (
-      <Wrapper>
+      <Wrapper
+      onMouseEnter={this.props.handleLeave}
```
DaleMcGrew: Since handleLeave and handleHover are not required functions (which is correct), I would create a local function that wraps the incoming functions, and then call those onMouseEnter and onMouseLeave. For an example of this, search for goToCandidate and goToCandidateLinkLocal in WebApp/src/js/components/Widgets/BallotItemSupportOpposeCountDisplay.jsx This will fail much more gracefully if the handleLeave and handleHover functions aren't passed in.

jarodpeachey: Just to clarify: 

In the IssuesByBallotItemDisplayList.jsx file, you want me to create a function inside render() that is called by the `onMouseEnter` on line 161, and through that function, access `this.props.handleLeave`?

DaleMcGrew: I would recommend naming the local function `handleLeaveHoverAreaLocal` and call that local function from line 161. I would recommend changing the incoming prop to be named `handleLeaveHoverArea` for clarity.

PR #2096, opened 2019-04-26
Rewrite hover state for Candidate.jsx for better code

jarodpeachey:
An update for #2020 
I added the changes as you requested in the earlier.  I'm working on disabling the hover on the candidate profile page and will submit a pull request for that when ready.

### DIFF
--- a/src/js/components/Values/IssuesByBallotItemDisplayList.jsx
+++ b/src/js/components/Values/IssuesByBallotItemDisplayList.jsx
@@ -13,8 +13,8 @@ class IssuesByBallotItemDisplayList extends Component {
   static propTypes = {
     ballotItemWeVoteId: PropTypes.string.isRequired,
     children: PropTypes.object,
-    handleLeave: PropTypes.func,
-    handleHover: PropTypes.func,
+    handleLeaveCandidateCard: PropTypes.func,
+    handleEnterCandidateCard: PropTypes.func,
   };
 
   constructor (props) {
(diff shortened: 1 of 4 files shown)


src/js/components/Widgets/BallotItemSupportOpposeCountDisplay.jsx
```diff
@@ -267,6 +267,14 @@ class BallotItemSupportOpposeCountDisplay extends Component {
   }
 
   render () {
+    const handleEnterHoverLocalArea = () => {
+      this.props.handleLeaveCandidateCard();
+    };
+
+    const handleLeaveHoverLocalArea = () => {
+      this.props.handleEnterCandidateCard();
```
DaleMcGrew: Since this function is coming in as a prop, and may or may not exist, you want to wrap it with:
```
if (this.props.handleEnterCandidateCard) {
  this.props.handleEnterCandidateCard();
}
```

src/js/components/Values/IssuesByBallotItemDisplayList.jsx
```diff
@@ -72,6 +72,14 @@ class IssuesByBallotItemDisplayList extends Component {
   }
 
   render () {
+    const handleEnterHoverLocalArea = () => {
+      this.props.handleLeaveCandidateCard();
```
DaleMcGrew: Since this function is coming in as a prop, and may or may not exist, you want to wrap it with:
```
if (this.props.handleLeaveCandidateCard) {
  this.props.handleLeaveCandidateCard();
}
```

src/sass/components/_card.scss
```diff
@@ -48,8 +49,8 @@
     }
 
     &--outline {
-      border: 1px solid $selected-color;
-      border-radius: $radius-sm3;
+      border: 1px solid $link-color;
+      border-radius: 3px;
```
DaleMcGrew: Please test with `npm test` -- I don't think specifying 3px here is allowed by eslint.

PR #2097, opened 2019-04-26
Disable hover state on candidate page

jarodpeachey:
I added a `showHover` prop to the `CandidateItem.jsx` component.  I only passed in that prop on the pages we wanted the hover state to show on.  On the `Candidate.jsx` page, the hover state is disabled.


DaleMcGrew:
👍 Thank you @jarodpeachey!

PR #2101, opened 2019-04-26
Update to remove underline on hover

jarodpeachey:
I noticed an underline on the content when the candidate card was hovered, so I disabled that.  I don't know if that's what you want or not, but here it is. :)


DaleMcGrew:
👍

PR #2115, opened 2019-05-01
Remove ghost hover of menu and make sure it still displays again

jarodpeachey:
There was a glitch in the IssueFollowToggleButton.  Because I hid the toggle button when you aren't following the issue, when you hit unfollow, the menu stayed visible.  I fixed this by grabbing the menu and removing the show class from it (Bootstrap)


DaleMcGrew:
👍

Issue #2094, opened 2019-04-25
Measure and Candidate Sticky header: Remove shadow under Back to Ballot bar


DaleMcGrew:
Measure pages and Candidate pages have a "Sticky header" that appears when you scroll down the page that shows either the measure name, or the candidate name. Please remove the shadow under the header bar <- Ballot when the Stick header is displayed:

1. Go to https://quality.wevote.us/ballot/id/wv02ballot16447
2. Click on "Measure"
3. Click on the first measure and scroll down on the page after positions appear on the screen

![Screen Shot 2019-04-25 at 11 28 42 AM](https://user-images.githubusercontent.com/7756031/56759765-95368900-674e-11e9-83df-963192f44cc7.png)

PR #2118, opened 2019-05-03
Removed shadow under Back to Ballot bar when the measure sticky header appears

uroshima:
#2094


DaleMcGrew:
👍  Thank you @uroshima! It has been great working with you.

PR #2124, opened 2019-05-06
iOS testing script and some housekeeping

flurmbo:

These changes resulted in 3/3 passing tests before making a PR.

* Added iOS capabilities to config file and handled switching to webview context for iOS in test script.
* Updates selectors to reflect changes to cordova and live QA site
* removed 'mocha' environment from the global .eslintrc file in favor of using it in a separete .eslintrc for the tests/browserstack folder. Added a new dependency, eslint-plugin-wdio, and included it in that file as well. Basically, this just prevents the Webdriverio globals ($, $$, browser, and driver) from raising linting warnings.
* updated .gitignore to reflect new name for user's local config file  (which was changed after we switched to using Webdriverio from Selenium Webdriver)

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -7,8 +7,7 @@
   "env": {
     "browser": true,
     "node": true,
-    "es6": true,
-    "mocha": true
+    "es6": true
   },
 
   "extends": "airbnb",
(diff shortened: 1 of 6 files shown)


DaleMcGrew:
👍

Issue #2102, opened 2019-04-26
Add rollover to "Our Team" photos


DaleMcGrew:
### Please describe the issue (What happens? What do you expect?)
https://quality.wevote.us/more/about
The left 1/2 of the image below shows our current display for one team member. We want to upgrade it to change when you rollover with your mouse (for Desktop) or click (for Mobile), and show the image on the right 1/2 of the image below (shrunken team member photo and title, with the bio shown below):
<img width="464" alt="Screen Shot 2019-04-26 at 1 21 03 PM" src="https://user-images.githubusercontent.com/7756031/56834541-84594680-6826-11e9-8f2a-b1f3e5a546fd.png">

The rollover/click version should fit in the same space as the default view.

The code that controls this display is in this component:
WebApp/src/js/components/More/TeamMemberDisplayForList.jsx

DaleMcGrew:
Thank you @jarodpeachey, in mobile mode, could you please make the Team names and job definitions  centered horizontally?

DaleMcGrew:
Fixed.

PR #2133, opened 2019-05-09
About Page Team Member Rollovers

jarodpeachey:
#2102 
I updated the component to use rollovers, and also disable the rollover if there was no description text coming from the props.teamMember.title array.  If you'd like, I can remove the functionality.  I didn't know if you were planning on adding descriptions to all the team members or not.


DaleMcGrew:
👍 I think disabling the rollover when there is no bio text is a great idea.

Issue #2132, opened 2019-05-09
Measure Page Sticky Header


DaleMcGrew:
- [x] Please center the sticky header (similar to how the Candidate Page Sticky Header looks). Currently it stick to the left side of the browser:
![Screen Shot 2019-05-09 at 9 57 54 AM](https://user-images.githubusercontent.com/7756031/57472039-5c63dd00-7241-11e9-9d4c-635c462b86f9.png)

- [x] Please add the WebApp/src/js/components/Widgets/BallotItemSupportOpposeCountDisplay.jsx

DaleMcGrew:
Great work @jarodpeachey! Verified.

PR #2138, opened 2019-05-11
Update measure sticky header

jarodpeachey:
#2132 
Updated the sticky header to use the same method as the candidate page, with the avatar removed.  Everything's good! 👍


DaleMcGrew:
Thank you @jarodpeachey! 👍

PR #2143, opened 2019-05-11
About page fixes

jarodpeachey:
This should fix those issues you emailed me about.  Most stemmed from the DescriptionContainer being a flex object. Let me know if this is good or I should look into another solution.


DaleMcGrew:
👍

PR #2163, opened 2019-05-15
updated gif uris in intro components

flurmbo:
Update the uris of the gifs in the Cordova intro dialog boxes. They were recently moved to a different folder. I also noticed that the gifs do not loop. I'm not sure if that is intended behavior.


DaleMcGrew:
👍 Thank you @flurmbo, we set up all of the animated gifs to only animate once (on purpose).

Issue #2157, opened 2019-05-14
Pricing Page: Show 3 options


DaleMcGrew:
An empty page already exists here:
https://quality.wevote.us/more/pricing
WebApp/src/js/routes/More/Pricing.jsx

Please add these columns in desktop:
<img width="756" alt="Screen Shot 2019-05-14 at 5 40 10 AM" src="https://user-images.githubusercontent.com/7756031/57698915-9b13e180-760b-11e9-883d-f1c3521456c4.png">

In mobile, please only show one column at a time, with a toggle (against the dark blue background, right below "Pricing") between the three options "Free", "Professional", "Enterprise", based on the current "HeaderSwitch.jsx" component (please make copy -- see below):
![Screen Shot 2019-05-14 at 5 47 04 AM](https://user-images.githubusercontent.com/7756031/57698977-c0085480-760b-11e9-84e9-3b705c8add6e.png)

Please copy this component (and rename "PricingSwitch.jsx") and modify it to work for a narrow mobile screen:
WebApp/src/js/components/Widgets/HeaderSwitch.jsx

DaleMcGrew:
Great work @jarodpeachey!

PR #2167, opened 2019-05-16
Pricing cards

jarodpeachey:
#2157 
Finished the pricing cards, including the mobile PricingSwitch component.


src/js/components/More/PricingCard.jsx
```diff
@@ -0,0 +1,698 @@
...
+              <CardWrapper>
+                {premium ? (
+                  <PremiumName>{planName}</PremiumName>
+                ) : (
+                  <DefaultName>{planName}</DefaultName>
+                )}
+                {price === 0 || price ? (
+                  <React.Fragment>
+                    <span style={{
+                      fontSize: '18px',
```
lazorfuzz: @DaleMcGrew is it a good idea to allow inline styling?

DaleMcGrew: Hello @jarodpeachey, I think @lazorfuzz has a good point. Could you please implement these as a named styled component (like you did with "PriceDescribe" for example) instead of using a span with inline styles?

jarodpeachey: Sure thing, I'm going out  right now, but should have that fixed by tonight!

DaleMcGrew:
👍 Aside from the inline styles comment, I am approving this pull request. I might make some text modifications (and submit another pull request) before you work on this again tonight, fyi. I'm excited to make public the Pricing page! Thank you @jarodpeachey.

PR #2169, opened 2019-05-16
Remove inline styles from PricingCards

jarodpeachey:
Updated the Pricing Cards to remove the inline styles


DaleMcGrew:
👍

Issue #2148, opened 2019-05-12
Footer: "Contact Sales" button has outline cover button text


DaleMcGrew:
When you resize the browser width when looking at a Welcome page, there are some widths when the "Get Started" and "Contact Sales" buttons in the footer have the outer border overlapping the button text:
<img width="129" alt="Screen Shot 2019-05-11 at 6 04 45 PM" src="https://user-images.githubusercontent.com/7756031/57576576-866ef800-7417-11e9-9125-af1939bc9835.png">

jarodpeachey:
I'll tackle this.  There's also an issue that the contact sales button does not go to the page unless you click the text.  Something with the "Get Started" button being a link, but the "Contact Sales" button only containing a link.  I'll take a look at this tomorrow.

DaleMcGrew:
Verified!

PR #2177, opened 2019-05-17
Fix button errors on mobile and update styling

jarodpeachey:
#2148 
Fixed the outline error and the font size being larger on mobile

### DIFF
--- a/src/js/components/Welcome/Footer.jsx
+++ b/src/js/components/Welcome/Footer.jsx
@@ -125,16 +125,9 @@ class Footer extends Component {
               color="default"
               variant="outlined"
               classes={{ root: classes.buttonOutlined }}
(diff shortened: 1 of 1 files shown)


src/js/components/Welcome/Footer.jsx
```diff
@@ -125,16 +125,9 @@ class Footer extends Component {
               color="default"
               variant="outlined"
               classes={{ root: classes.buttonOutlined }}
+              onClick={() => window.open('https://help.wevote.us/hc/en-us/requests/new', '_blank')}
             >
```
DaleMcGrew: Hi @jarodpeachey, in general we don't want to directly open external links like this because it makes the page not compatible with Cordova, the system we use to make our mobile apps. We can get away with this here though since we don't offer the Welcome pages within our iOS and Android apps.

DaleMcGrew:
👍

Issue #2175, opened 2019-05-17
Refactor Pricing to use Data array for options


DaleMcGrew:
https://quality.wevote.us/more/pricing

Without changing the look or functionality of the current pricing page, refactor it so that all of the options and text are in a new data structure as proposed here: 
https://github.com/wevote/WebApp/pull/2174/files

WebApp/src/js/routes/More/Pricing.jsx

This file has examples of how to use this type of data structure:
WebApp/src/js/routes/HowItWorks.jsx

Please note that we will also have two sets of pricing page language (the text will be similar, but different). One written for nonpartisan organizations, and another written for campaigns that advocate for candidates.

jarodpeachey:
Hey @DaleMcGrew.  I just finished refactoring the ForCampaigns data, and I just copied that to use as the forOrganizations data because I don't have the correct data for that.  

Anyways, should I submit this pull request, or update the pricing page to use the same slider method as the How It Works page to toggle between campaign and organization pricing cards?

DaleMcGrew:
Good morning @jarodpeachey, using a copy of the same data forCampaigns and forOrganizations is perfect. I will adjust it per Hillary Lehr, our marketing writer. It sounds like you are ready to submit. I think I will set up the choice between forCampaigns and forOrganizations based on the URL the voter is visiting.

DaleMcGrew:
Minor bug:
- [ ] Between browser width 960px and 991px the "Start With..." buttons don't line up horizontally.
<img width="721" alt="Screen Shot 2019-05-19 at 10 40 01 AM" src="https://user-images.githubusercontent.com/7756031/57985823-888e1380-7a22-11e9-9fa2-5c4bab23f799.png">

DaleMcGrew:
Well done! Verified.

PR #2180, opened 2019-05-18
Pricing page data refactor

jarodpeachey:
#2175 
Updated the pricing cards to use the new data structure.


src/js/components/More/PricingCard.jsx
```diff
@@ -108,202 +111,456 @@ class PricingCard extends Component {
                 <br />
                 <Collection>
                   <CollectionItem>
-                    <Check
-                      style={{
-                        color: '#2E3C5D',
-                      }}
-                    />
-                    <ItemText>Enter Your Own Positions</ItemText>
+                    {pricingCardFeatures[0].iconType === 'checkMark' ? (
```
DaleMcGrew: Hi @jarodpeachey, in order to reduce duplicate code, I was hoping you could create a for/loop for the CollectionItem, so that you could pass in "pricingCardFeatures" and it could loop through all of the data in that array. This would avoid hard-coding "ItemText" values in this file too.

DaleMcGrew:
Awesome, thank you @jarodpeachey! 👍

Issue #2153, opened 2019-05-13
Mobile SignInModal: Make wider


DaleMcGrew:
When you are in mobile mode, and not signed in, we open the component WebApp/src/js/components/Widgets/SignInModal.jsx when you click "Get Started" from any welcome page. Currently this SignInModal is too narrow in mobile mode and should fill the entire width of the mobile phone.

DaleMcGrew:
Verified!

PR #2183, opened 2019-05-19
Make modal wider on mobile

jarodpeachey:
#2153 
Updated the width of the sign in modal only on mobile.  The whole component (eg. the buttons) will need to be made responsive at some point and updated to use the current style of the site, but that can be a later issue.


DaleMcGrew:
👍 We have some new visual designs for the sign in page that I will get posted as a github Issue.

Issue #2092, opened 2019-04-25
Office Detail: Add Choose/Oppose buttons


DaleMcGrew:
### Please describe the issue (What happens? What do you expect?)
On the office detail, hovering over most of the candidate card should also cause the "Choose" and "Oppose" buttons to appear, and be clickable.

This design image is slightly incorrect -- the values icons + text would not be link-blue, based on where the pointer is over the card.
<img width="491" alt="Screen Shot 2019-03-22 at 1 36 35 PM" src="https://user-images.githubusercontent.com/7756031/54852461-2f1b9980-4caa-11e9-9cc8-aad1ad21c2fa.png">

jarodpeachey:
What component should I use? I was looking and there doesn't seem to be a component for just the choose/oppose buttons.  How should I approach making sure the buttons fire the function they are supposed to?

DaleMcGrew:
@jarodpeachey I believe you could use this component for the Choose / Oppose buttons:
WebApp/src/js/components/Widgets/ItemActionBar.jsx

You would need to pass in the prop: commentButtonHide

jarodpeachey:
This would be on the CandidateItemCompressed, correct?

DaleMcGrew:
No, it looks like it is in:
WebApp/src/js/components/Ballot/CandidateList.jsx
WebApp/src/js/components/Ballot/CandidateItem.jsx

jarodpeachey:
OK

DaleMcGrew:
Great work @jarodpeachey, verified.

PR #2192, opened 2019-05-20
Update to include choose/oppose buttons on candidate card hover

jarodpeachey:
#2092 
Included the choose/oppose buttons. However, I can't fix the issue that when we click on one of the buttons to change our position on that candidate, the button changes and the popover appears, but after 1 second the position changes back to what it was before.  This could be a separate issue.

WeVoteChooseOppose


DaleMcGrew:
👍

PR #2193, opened 2019-05-20
Added new buttons with shortened text to the Address & Elections Moda…

ffrancetic:
Added new buttons with shortened text to the Address & Elections Modal and for the Supported Elections page


DaleMcGrew:
👍

PR #2216, opened 2019-05-24
Fixed the width of the buttons and the spacing in between on the dona…

ffrancetic:
Fixed the width of the buttons and the spacing in between on the donations page. Also made an attempt to fix the problem of the buttons not working on refresh.


DaleMcGrew:
👍

PR #2221, opened 2019-05-25
Position item updates

jarodpeachey:
Updated the position items as per your email.  To me, the follow toggle button looks kinda in an odd position on mobile, but let me know what you think.


DaleMcGrew:
👍

PR #2223, opened 2019-05-25
Adjusted the top margin of the scrollable area for Android and Apple

SailingSteve:
New function ifCordovaGetOffset() works with the new styled.div
The profile-menu now displays in Cordova -- it had been styled off the screen.
Should fix wevote/WebApp/issues/2172


DaleMcGrew:
👍

PR #2229, opened 2019-05-29
Stable top margins, header sizes, offset to scrolling pane for Android and iOS in Cordova.

SailingSteve:
Also tested in WebApp.
Big refactor and simplification of the code that adjusts the styles for different device sizes
and mobile OSes.  
Renamed components/Welcome/Header to components/Welcome/HeaderWelcome to avoid duplicate
class names that made debugging more difficult.


DaleMcGrew:
👍 Thank you @SailingSteve!

Issue #2185, opened 2019-05-19
Problem with rendering organization placeholder image


DaleMcGrew:
When looking at the positions for a candidate:
https://localhost:3000/candidate/wv02cand40208/b/btdb/

...if a logo is missing for the person/organization, we see a big "C":
![Screen Shot 2019-05-19 at 7 53 46 AM](https://user-images.githubusercontent.com/7756031/57985529-0ac80900-7a1e-11e9-9f43-f3de9948e177.png)

We should be showing our default icon, and this used to work.

I am also seeing this error in the console.log with I think is related:
<img width="544" alt="Screen Shot 2019-05-19 at 10 05 48 AM" src="https://user-images.githubusercontent.com/7756031/57985521-f1bf5800-7a1d-11e9-900a-df6d11457662.png">

I believe this is a problem in: 
WebApp/src/js/components/Ballot/PositionItem.jsx

...with either "imagePlaceholder" or <ImageHandler

jarodpeachey:
Sure thing, I'll take this next

DaleMcGrew:
Thank you @jarodpeachey, this looks great. Could you please figure out a way to show the placeholder icon smaller when in mobile?

![Screen Shot 2019-05-29 at 9 25 21 AM](https://user-images.githubusercontent.com/7756031/58573993-da3d5780-81f3-11e9-9910-852a3b55e5dc.png)

DaleMcGrew:
Solved!

PR #2232, opened 2019-05-29
Fix placeholder not showing and update position item so placeholder i…

jarodpeachey:
…mages were formatted properly

#2185 
Updated the import path to pull in the font files and formatted the position item placeholder image to be the same width and height as the profile images


DaleMcGrew:
Thank you @jarodpeachey! 👍

PR #2236, opened 2019-05-29
Updated the ballotMainTest.js file

ffrancetic:
Updated the ballotMainTest.js file


DaleMcGrew:
Filip and myself are working on this so it is a bit rough. 👍

PR #2239, opened 2019-05-30
Update the mobile version of the placeholder images

jarodpeachey:
Mobile image placeholders are now the correct size


DaleMcGrew:
👍

PR #2242, opened 2019-05-30
More header size adjustments, and fine tuning for Android.

SailingSteve:
Added buttons to the bottom of the hamburger menu that only appear for developers in
the emulators, that allow cookies to be cleared, and forces a navigation
to the /wevoteintro/network page for testing.


DaleMcGrew:
👍

PR #2246, opened 2019-05-31
Removed retreating header for Cordova, and fixed headers for Vote.jsx

SailingSteve:
retreating header (similar to headroom.jsx) shouldHeaderRetreat(), etc.
Added dummy divs to identify in debugger which styled.div is which, since
you can't add an id to a styled.div
Retested on 6s, 8, 8+, X, Xr, XsMax, iPadPro, Nexus one, Nexus 5x,
Nexus 6p, & Nexus 10.


DaleMcGrew:
👍

PR #2252, opened 2019-05-31
Update the comment box to show more lines on focus

jarodpeachey:
Here are the changes we discussed in the standup!


DaleMcGrew:
👍

PR #2255, opened 2019-06-05
Move selectBallotModal out of Ballot.jsx into HeaderBar.jsx

jarodpeachey:
#2225 

I moved the component, but the modal still closes if you press it before the page is completely loaded.  This might be a deeper issue than we originally thought.


DaleMcGrew:
Thank you @jarodpeachey, I will play with this to see what is going on. 👍

PR #2261, opened 2019-06-09
Adjusted headers throughout, now sensitive to long/short header on ballot.

SailingSteve:
Tested on the 11 browsers.
Improvement on the Vote tab header, and scrollable top padding.


DaleMcGrew:
👍

Issue #2262, opened 2019-06-10
ItemPositionStatementActionBar: Stack "Friends only" and "Public" options in mobile


DaleMcGrew:
Stack the "Friends only" and "Public" options in mobile:
<img width="652" alt="screen shot 2019-02-22 at 12 13 47 pm" src="https://user-images.githubusercontent.com/7756031/53268586-6000b280-369b-11e9-931f-35663264f8ff.png">

Currently they make the interface too wide when shown left-to-right:
![Screen Shot 2019-06-09 at 5 32 18 PM](https://user-images.githubusercontent.com/7756031/59166262-98a68980-8adc-11e9-9c82-a598a47b68ba.png)

jarodpeachey:
My bad, I missed this. I'll grab this issue.

jarodpeachey:
Hey @DaleMcGrew ! I'm working on this now and found another way we can fit this on the mobile screen. I've made the buttons smaller and took away the button text on mobile (The toggle still isn't stacked, but I'm still working on this.) Does this method look okay?

DaleMcGrew:
Verified. Thank you @jarodpeachey!

PR #2269, opened 2019-06-10
Friends/Public toggle fit on screen on iPhone 5

jarodpeachey:
#2262 
I didn't stack the toggle component because it was not needed after I changed the support oppose buttons and added flex: 1 1 0; to the toggle component so it does not get cut off.

**iPhone 5 display**
Capture4

If you don't like this, I can stack the toggle options and revert the buttons.


DaleMcGrew:
Hi @jarodpeachey, thank you for trying that solution. I would prefer to have the Friends/Public radio buttons stacked vertically so we can continue to include the text for the Choose and Oppose buttons. You might need to tighten up the margins for the Choose and Oppose buttons in order for them to fit, even after you stack the Friends/Public radio buttons.

jarodpeachey:
Updates are completed and the options are now stacked 👍

DaleMcGrew:
👍

PR #2284, opened 2019-06-11
Updated new devices for testing. Added clicks on badges on the ballot…

ffrancetic:
Updated new devices for testing. Added clicks on badges on the ballot page for testing.


DaleMcGrew:
👍

Issue #2259, opened 2019-06-07
Organization Follow Toggle Not Full Width


jarodpeachey:
### Please describe the issue (What happens? What do you expect?)
On the organization page, when you're following that organization, the button is full width. However, if you're not following, the button is not full width.
![Capture](https://user-images.githubusercontent.com/46693952/59115632-af4fb380-8917-11e9-8f39-871965e02738.PNG)
![Capture2](https://user-images.githubusercontent.com/46693952/59115636-b080e080-8917-11e9-98ac-a068d549a1fd.PNG)

jarodpeachey:
I'll take this issue, because the entire component has issues. For example, the drop down menu still is hidden behind the edge of the card in most places, and the horizontal separator line is still out of whack.  I'll see what I can do to fix this.

DaleMcGrew:
Works great!

PR #2287, opened 2019-06-11
Fix FollowToggle component

jarodpeachey:
#2259 
I added a position of relative to the parent elements that were overlapping the follow toggle drop-down menu so Popper.js would recognize them and re position the menu. As far as I can see nothing else has been affected, let me know if you find anything!

** I also added an anchorLeft prop so we can display that better for different situations, as well as get rid of unneeded divs by changing them to React.Fragments

### DIFF
--- a/src/js/components/Ballot/PositionItem.jsx
+++ b/src/js/components/Ballot/PositionItem.jsx
@@ -106,7 +106,7 @@ class PositionItem extends Component {
                     </Link>
                   </StickyPopover>
                 </DesktopItemImage>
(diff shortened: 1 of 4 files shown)


DaleMcGrew:
Thank you @jarodpeachey. Could you please fix this eslint warnings and errors and update your pull request?
Screen Shot 2019-06-11 at 8 19 14 PM

jarodpeachey:
All good!

DaleMcGrew:
👍

PR #2289, opened 2019-06-12
Update so form focuses in and out again

jarodpeachey:
There was a problem on the comment box on the candidate page where it would no longer change height based on focus or blur. I fixed this by adding the onFocus and onBlur events to the form rather than the input, so that the button was included as well.


DaleMcGrew:
👍

PR #2291, opened 2019-06-12
Hide "show more" footer in print mode. Make checkmarks and the oppose circle visible during print mode.

LohitG:
Hide "show more" footer in print mode. Make checkmarks and the oppose circle visible during print mode.


DaleMcGrew:
Congratulations with your first We Vote pull request @LohitG! 👍

PR #2292, opened 2019-06-12
Changed all off the showToastErrors and showToastSuccess' to SnackBar.

ffrancetic:
Changed all off the showToastErrors and showToastSuccess' to SnackBar.


DaleMcGrew:
👍

PR #2297, opened 2019-06-13
Update popover styling

jarodpeachey:
#2190 
I updated the styling of the popover component. I did not include the followers count and issues list, didn't know if we wanted that yet or not, I just changed the styling of what was currently there.

I also updated the follow toggle button so the blue button had no border, and changed the menu to have a box-shadow


DaleMcGrew:
Looks great @jarodpeachey! 👍

PR #2303, opened 2019-06-13
Disabled StickyHeaders in Cordova

SailingSteve:
On iOS, the StickyHeader only appears after you stop scrolling, this results in the header often appearing over what you are currently reading.  Apple does this on purpose, queues updates until scrolling stops, to avoid repainting the DOM while scrolling.

See:  https://github.com/sroze/ngInfiniteScroll/issues/104#issuecomment-68000337

Later comments on the thread incorrectly indicate that the issue is fixed.  When I switched from 'scroll' to 'touchmove' you could see the touchmove data, but the actual on screen behavior confirmed the (purposeful by apple) issue.  Also when I put a console.log in the event handler, the entries made it to the log, only after scrolling had stopped.

### DIFF
--- a/src/js/Application.jsx
+++ b/src/js/Application.jsx
@@ -58,7 +58,7 @@ class Application extends Component {
     this.appStoreListener = AppStore.addListener(this.onAppStoreChange.bind(this));
     this.voterStoreListener = VoterStore.addListener(this.onVoterStoreChange.bind(this));
     if (isWebApp()) {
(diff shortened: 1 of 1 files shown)


DaleMcGrew:
👍

SailingSteve:
On May 22nd, 2020 the underlying iOS issue seems to have been resolved in newer versions of Cordova (or maybe in iOS updates?).  So backing out the change in Application.js that disabled handleWindowScroll() if isCordova.

PR #2304, opened 2019-06-14
Popover fixes

jarodpeachey:
I fixed the popover as per your email. However, I worked at trying to display it underneath the sticky header for over an hour, and cannot figure out what to do. I've tried z-index, opacity, and just about everything else on Google.  Let me know if you can come up with anything!


DaleMcGrew:
👍 Thank you for the banner fix @jarodpeachey, and thank you for taking a look at the popover placement relative to the header. I'm sure we will get it figured out!

PR #2313, opened 2019-06-15
Show AppBar logo in print

lazorfuzz:
Shows the AppBar with logo, but not the sign in button when in print mode. Also set the ballot filter row display to none when in print mode.


DaleMcGrew:
👍

Issue #2315, opened 2019-06-15
Welcome Pages: Our Network Section too large on desktop


DaleMcGrew:
The "Our Network" section of all three welcome pages has changed, and the logos are no longer in two columns like they were:
<img width="707" alt="Screen Shot 2019-06-15 at 6 21 27 AM" src="https://user-images.githubusercontent.com/7756031/59551940-dcaee980-8f35-11e9-856b-b3bc3fcf60c3.png">

Also in mobile the logos need some space between them:
<img width="300" alt="Screen Shot 2019-06-15 at 6 21 11 AM" src="https://user-images.githubusercontent.com/7756031/59551944-e20c3400-8f35-11e9-8e94-5315b7d0381b.png">

I'm not sure what changed because my recollection was that they were working.

DaleMcGrew:
Resolved!

PR #2319, opened 2019-06-17
[Issue #2315] Welcome Pages: Our Network Section too large on desktop

flurmbo:
#2315

Looks like I forgot to add a couple semicolons in the CSS and the browser interpreted three lines as a single CSS rule. Oddly, this just started happening. Perhaps something in the build process was adding semicolons in and is no longer doing so?

Also, it looks like this fixes the spacing issue.


DaleMcGrew:
Thank you @flurmbo!

Issue #2318, opened 2019-06-17
OrganizationPopoverCard: If no banner photo, spacing is off


DaleMcGrew:
For organizations without any banner photo, the OrganizationPopoverCard spacing is off at the top. You can reproduce this on this page:
http://localhost:3000/candidate/wv02cand40208/b/btdb/

<img width="416" alt="Screen Shot 2019-06-16 at 8 16 21 PM" src="https://user-images.githubusercontent.com/7756031/59576239-eb031f80-9073-11e9-96e6-0660ad911534.png">

Make sure to grab the latest code, because this problem appeared when I turned off the image when it doesn't exist.

DaleMcGrew:
This looks great! Resolved.

Issue #2294, opened 2019-06-13
Issue popover: X positioning wrong, and popover centering instead of being at bottom


DaleMcGrew:
When you click an Issue under a candidate:
<img width="303" alt="Screen Shot 2019-06-12 at 7 25 42 PM" src="https://user-images.githubusercontent.com/7756031/59399274-4b2e5480-8d48-11e9-85e9-51432cdec538.png">

The popover opens with two problems:
<img width="309" alt="Screen Shot 2019-06-12 at 7 25 33 PM" src="https://user-images.githubusercontent.com/7756031/59399293-55505300-8d48-11e9-9891-6852cc22b136.png">

- [ ] 1: The x to close the popover should be right aligned
- [ ] 2: The popover should open under the Issue, instead of opening centered over the issue

jarodpeachey:
@DaleMcGrew Should this be a hover popover as well, sort of like the OrganizationPopoverCard? I can use the sticky popover component, that way all the popovers have a consistent look and feel.

DaleMcGrew:
Yes @jarodpeachey, that sounds great to use the same kind of popover as we are for OrganizationPopoverCard. In this case we want to open it with a click, and include an "x" to close it with another click. I am also fine with the popover closing when you move your mouse away from it. Please make sure to test extensively in mobile, since the OrganizationPopoverCard isn't used in mobile, so this code may have some different reactions when viewed in a mobile context.

DaleMcGrew:
@jarodpeachey your solution works beautifully. Great work!

PR #2321, opened 2019-06-17
Issues popover

jarodpeachey:
#2294 #2318
I ended up having to set placement to auto to make sure that it wouldn't overflow off the screen. I also updated the spacing of the issues list so it didn't take up as much vertical space. I also updated the popover so that if there was no banner image, it pushes the content down


DaleMcGrew:
👍

Issue #2305, opened 2019-06-14
Update Bootstrap Button with Twitter icon


DaleMcGrew:
I couldn't figure out how to add a white Twitter bird icon to this Bootstrap social button. It should be easy, but I ran out of time. 
Code is here: WebApp/src/js/components/Widgets/ThisIsMeAction.jsx
![Screen Shot 2019-06-14 at 8 00 38 AM](https://user-images.githubusercontent.com/7756031/59518632-db6cb680-8e7a-11e9-8fdb-d774ccdecc2a.png)

jarodpeachey:
@DaleMcGrew I can take this. I do have one comment though. Currently, the Add Endorsements button, as well as this Claim button, are using the bootstrap social buttons. This leads to the buttons having the blue outline when clicked, and they don't have the ripple. Should I fix all of these and use the technique I used for the Follow/Toggle  component, which is to use a Material UI button with the bootstrap-social buttons. This way, the buttons can fit in more with the other buttons in the site?

DaleMcGrew:
Yes please @jarodpeachey, that would be awesome!

DaleMcGrew:
Great work!

PR #2327, opened 2019-06-18
Update bootstrap social buttons to material-ui and custom split button

jarodpeachey:
#2305 
I updated this icon, and I also updated the endorsement buttons and twitter sign in button to use a custom split button I designed so the style is consistent with the rest of the site. If you would like, I could get rid of the CSS I used and create an entirely new component (SplitButton.jsx) and use styled-component and pass in props to adjust button text and colors, etc;  Let me know what you think!

### DIFF
--- a/src/js/components/Twitter/TwitterSignIn.jsx
+++ b/src/js/components/Twitter/TwitterSignIn.jsx
@@ -1,5 +1,7 @@
 import React, { Component } from 'react';
 import PropTypes from 'prop-types';
+import Button from '@material-ui/core/Button';
+import { shortenText } from '../../utils/textFormat';
 import { oAuthLog, renderLog } from '../../utils/logging';
 import $ajax from '../../utils/service';
 import cookies from '../../utils/cookies';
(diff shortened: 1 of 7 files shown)


leonflix:
@jarodpeachey I believe such a component is the ContainedIconButton that was already being used in TwitterSignIn.jsx.  It might need some additional tweaks but it should serve the purpose.

lazorfuzz:
@jarodpeachey I believe such a component "ContainedIconButton" was already being used in TwitterSignIn.jsx. It probably needs additional tweaks but should serve the purpose.

DaleMcGrew:
Thank you @jarodpeachey, and thank you @lazorfuzz for pointing out WebApp/src/js/components/Twitter/ContainedIconButton.jsx so we can leverage the work you have done.
I like the idea of trying to fit all of our buttons that include an icon and text into a component so we have a standardized look. I'm going to approve this pull request so I can check out the look of your changes. I will Slack you @jarodpeachey to see if we can review all of the button locations.

PR #2332, opened 2019-06-18
Make text bigger when printing. Remove underlines from linked text when printing. Hiding a few more popups that are not needed in print mode. Bring back the header on the Ballot page that was previously removed.

LohitG:
Make text bigger when printing. Remove underlines from linked text when printing. Hiding a few more popups that are not needed in print mode. Bring back the header on the Ballot page that was previously removed.


DaleMcGrew:
👍

PR #2339, opened 2019-06-19
Split button

jarodpeachey:
I updated the remainder of the bootstrap social buttons to use a new component, SplitIconButton. I also made some minor updates to the IssuesByBallotItemDisplayList while I was at it. Let me know if everything works as expected 👍


DaleMcGrew:
👍 Thank you for refactoring all of these buttons!

PR #2342, opened 2019-06-19
Layout fix for the WelcomeAppbar hamburger (IconButton)

SailingSteve:
Fixed top header margin for HowItWorks, WelcomeForCampaigns,
WelcomeForOrganizations, WelcomeForVoters.
Fixed apostrophe that did not render correctly in WelcomeForOrganizations.
Removed donate from Cordova builds.

### DIFF
--- a/src/js/Application.jsx
+++ b/src/js/Application.jsx
@@ -277,7 +277,7 @@ class Application extends Component {
       return (
         <div className="app-base" id="app-base-id">
           <div id="the styled div that follows is the wrapper for theatre mode">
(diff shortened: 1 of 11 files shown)


DaleMcGrew:
👍 Hi @SailingSteve, I am approving this pull request now. Do you think we should:
A) comment out the "About We Vote" link from the Hamburger menu, or
B) leave the "About We Vote" link and instead comment out the Footer links to these pages which are broken: 1) Register to Vote, 2) Get Your Absentee Ballot?

Either option would be viable, but option B would be more work...

PR #2343, opened 2019-06-19
Updated the testing files and added a browser test for the OnePlus 6T

ffrancetic:
Updated the testing files and added a browser test for the OnePlus 6T


DaleMcGrew:
👍

PR #2348, opened 2019-06-19
Layout fix for the Attributions, Pricing, Terms, Tools

SailingSteve:
Also fixed Attributions bottom padding
resolves wevote/WebApp/issues/2345
Added OpenExternalWebSite conditionals for new links in Footer.jsx to allow sites to load in a new tab, with a close button to return to the app.


DaleMcGrew:
👍

PR #2358, opened 2019-06-20
Vote page spacing fixes

jarodpeachey:
I updated the vote page to use conditional padding based on if there are candidates selected or not.


DaleMcGrew:
👍

PR #2360, opened 2019-06-20
Fixed bugs on the BallotItemSupportOpposeCountDisplay and changed some color schemes when printing

LohitG:
Following a person who opposes a candidate will cause a "-1" to replace the thumbs down. This matches what happens when a "+1" appears when you follow a person who supports a candidate. Negative endorsement totals are now red and a 0 net endorsements is grey. Some minor changes in colors when printing were made.


DaleMcGrew:
👍  Great work!

PR #2361, opened 2019-06-21
Fix two bugs that show in Cordova

SailingSteve:
Resolves wevote/WebApp/issues/2353
return to Vote page from 'https://wevote.us/vip.html'

Resolves wevote/WebApp/issues/2352
If your ballot does not appear momentarily, please click to change your address. vertical spacing


DaleMcGrew:
👍

PR #2363, opened 2019-06-21
Header/Scrollable area vertical spacing code for 4" iPhones (5s & SE)

SailingSteve:
Can't debug iPhone 5 due to simulator crash, but it is the same size as
the 5s so it should be ok.

Also resolves wevote/WebApp/issues/2350
Couldn't scroll to top of candidate page after scrolling to bottom


DaleMcGrew:
👍

PR #2371, opened 2019-06-21
Fixed the height of the Next button on the intro pages for Cordova

SailingSteve:
Also fixed the layout of the intro pages for the Galaxy Tab A 10.1 like in
in wevote/WebApp/issues/2357,
but I don't have access to a simulator for the Galaxy Tab S4 so I can't
test for that device.


DaleMcGrew:
👍

PR #2375, opened 2019-06-22
Update PositionItem for iPhone 5/SE Layout

jarodpeachey:
This should be ready to go, found a fix to only break the word if it's impossible to wrap it at the space. I also updated the layout for iPhone 5 to move the issues down to their own row.


DaleMcGrew:
Thank you @jarodpeachey! 👍

PR #2378, opened 2019-06-23
Minor fixes to sign-in modal and position item

jarodpeachey:
I updated the width of the sign in modal, and changed the colors of the position item to use the brands green and red colors. I also updated the candidate page to use the card for the position list


DaleMcGrew:
Thank you @jarodpeachey for making these improvements and refactoring changes. It is refactoring investments like this that will keep our code-base strong! 👍

PR #2386, opened 2019-06-24
Added Xs for the Dialog Modals and updated the Address Modal.

ffrancetic:
Added Xs for the Dialog Modals and updated the Address Modal.


DaleMcGrew:
👍

PR #2389, opened 2019-06-24
Fixed some npm errors and removed footer when keyboard appears

LohitG:
Fixed some npm errors and removed footer when keyboard appears. Someone needs to test the footer disappearing because I cannot.


DaleMcGrew:
👍

PR #2391, opened 2019-06-25
Cordova header fixes for Vote page based on logged in status

SailingSteve:
Other misc header offset fixes.
Added a new dialog, launches from the bottom of the hamburger menu if
isCordova, shows the device name which is needed to resolve problems on Android devices that we haven't tested on.
Now attempt to determine device type via the window.device.model, which is more reliable that detecting by screen size, if we don't have a device model match we drop back to screen size.
Added a new logger "cordovaOffsetLog" which requires another line in config.js


DaleMcGrew:
👍

PR #2393, opened 2019-06-26
Added animations on How It Works Page

LohitG:
Added animations on How It Works Page

### DIFF
--- a/src/js/components/Widgets/AnnotatedSlideshow.jsx
+++ b/src/js/components/Widgets/AnnotatedSlideshow.jsx
@@ -15,6 +15,10 @@ class AnnotatedSlideshow extends PureComponent {
     classes: PropTypes.object,
   };
 
+  componentDidMount() {
+    this.autoAdvanceSlide();
+  }
+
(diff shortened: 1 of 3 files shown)


src/js/components/Widgets/AnnotatedSlideshow.jsx
```diff
@@ -23,14 +27,27 @@ class AnnotatedSlideshow extends PureComponent {
...
+
   }
 
   render () {
     const { slides, selectedStepIndex, classes } = this.props;
     const data = Object.values(slides);
     const { length } = data;
     const { title, description, imgSrc } = data.find(slide => slide.index === selectedStepIndex);
-
+    this.componentDidMount();
```
DaleMcGrew: Hi @LohitG componentDidMount is a built in lifecycle function that shouldn’t be called with every render.

LohitG:
Oh sorry I’m out for the evening. I can change it tomorrow if you would
like.

DaleMcGrew:
👍

Issue #2347, opened 2019-06-19
Need to change color of text when printing


LohitG:
In the ballot badge in Ballot.jsx, there is a styled theme. The theme turns the background blue and the text grey when the badge is selected. However when printing, the background turns white, leaving the grey text very hard to read. I can't seem to figure out how to turn the grey text a darker color when printing. Dale told me that you were very well versed in this so I was hoping you could help out. Is there a way to set a "theme" like we do with "theme.breakpoints" but that is applied to print mode?

<img width="921" alt="Screen Shot 2019-06-19 at 15 06 31" src="https://user-images.githubusercontent.com/23069689/59804661-dad18700-92a3-11e9-88f2-41c22ca34b60.png">
<img width="1094" alt="Screen Shot 2019-06-19 at 15 07 03" src="https://user-images.githubusercontent.com/23069689/59804682-e9b83980-92a3-11e9-8b14-ac0274780090.png">
<img width="931" alt="Screen Shot 2019-06-19 at 15 07 25" src="https://user-images.githubusercontent.com/23069689/59804707-fa68af80-92a3-11e9-9de5-040abbc1391c.png">

ffrancetic:
[Values, Public Figures & Organizations - We Vote.pdf](https://github.com/wevote/WebApp/files/3352121/Values.Public.Figures.Organizations.-.We.Vote.pdf)
The text colors are still off. This is only Firefox related.

ffrancetic:
@lazorfuzz Setting the MuiButton print color to white globally doesn't seem to change the text color for the FireFox browser.
<img width="964" alt="Screenshot 2019-07-03 at 12 36 53" src="https://user-images.githubusercontent.com/36188818/60620151-4e3cc380-9d8f-11e9-8e40-e914ba411e75.png">

LohitG:
Yeah when I was trying to fix this I found that Firefox seemed to resist print related CSS while other browsers changed accordingly.

jarodpeachey:
@DaleMcGrew The badges are completely disappearing from the print mode. Was this from debugging this issue previously, or is that a new bug I should work on?

DaleMcGrew:
@jarodpeachey that sounds like a new bug. Thank you!

DaleMcGrew:
No longer a problem with new design.

PR #2398, opened 2019-06-28
[Issue #2347] Change badge color in print mode

lazorfuzz:
#2347 

@LohitG this is a Material UI component. In the style overrides, you can add something like this: 
```javascript
'@media print': {
      color: theme.palette.primary.main,
    },
```


DaleMcGrew:
👍

PR #2399, opened 2019-06-28
Update cancel button to close modal

jarodpeachey:
I updated the ballot modal to close if the location cancel button is hit


DaleMcGrew:
👍

PR #2401, opened 2019-06-29
Removed a number of media queries, if Cordoba - they cause lots of problems

SailingSteve:
Split the CordovaUtilites into two files -- new CordovaOffsets.jsx
Pulled the inline BallotTitleHeader code out of Ballot and Vote, it was
the same code, and would have needed three level ternary in each instance.
Adjusted the height of the Values page for all Android.

This code and all other code merged to WebApp before 2:40pm on 8/28/19, are in version 1.2.1 (build 1.2.1.2).

[Test matrix for headers and footers in Cordova](https://docs.google.com/spreadsheets/d/1-_0jthH0coeWqs3KZh4TxSToSgscL-UP64wV7nwkIsI/edit?usp=sharing)

Commit 2:
Fixed problem in Cordova where the guess dialog was off screen
Fixed the on screen name of the  We Vote icon to be "We Vote" instead of "WeVoteCordova"

Improved the top margin on all the ballot pages in iOS, the other pages are still "as is" with top margins that conserve vertical space, but are not in keeping with the UI design.
Android has similar changes.  Discovered a problem, that probably was pre-existing, where when the guess dialog appears, a horizontal band of white obscures the middle, maybe 10% of the page, at around the 1/3 point.  It appears to be a styling issue, it is too complex to fix today.  Does not reproduce on my physical Motorola G5 phone.

These Commit2: changes made it into iOS version 1.2.1 (build 1.2.1.4)


DaleMcGrew:
👍

PR #2405, opened 2019-07-01
Fixed the "Unexpected unnamed function func-names" warning

ffrancetic:
Fixed the "Unexpected unnamed function func-names" warning


DaleMcGrew:
👍

PR #2397, opened 2019-06-27
Update summary of ballot items 

jarodpeachey:
### What github.com/wevote/WebApp/issues does this fix?
#2377 
### Changes included this pull request?
I created an accordion component to use for the ballot summary component. Everything seems to work for every election, if you see anything that I missed, let me know.

PR #2411, opened 2019-07-03
Everything seems to work, crashing bug was fixed

jarodpeachey:
#2397 
There was already an activeRaceLevel set in Ballot.jsx, so I just passed in that same prop down, and that opens the accordion depending on which race level item is active. I also fixed the crashing values-ballot tab bug by changing line 200 in BallotSideBar.jsx.


DaleMcGrew:
👍 Great work!

Issue #2288, opened 2019-06-12
When at 575px, the screen shows two different breakpoints


jarodpeachey:
### Please describe the issue (What happens? What do you expect?)
When the screen size is 575px, both the mobile and desktop versions of everything show up. This is something to do with the media querys.

**Below, the ballot text shows up twice, the PositionItem shows up twice and other errors**
![Capture5](https://user-images.githubusercontent.com/46693952/59315022-e3afd080-8c85-11e9-9fd0-89edaa6a83ec.PNG)

ffrancetic:
Cannot be reproduced on a Mac for some reason. Made multiple attempts.

DaleMcGrew:
Resolved.

PR #2414, opened 2019-07-03
Fix double media query display issue

jarodpeachey:
#2288 
I updated the media queries to fix the double display issue, everything else seems to be unaffected by my change.


DaleMcGrew:
👍

PR #2426, opened 2019-07-08
Removed links for the IssueCards on the Values Listing page. Added the sum of public figures and organizations on the IssueCards.

ffrancetic:
Removed links for the IssueCards on the Values Listing page. Added the sum of public figures and organizations on the IssueCards.


DaleMcGrew:
👍

Issue #2419, opened 2019-07-05
Domain Settings: Create First Draft Interface


DaleMcGrew:
http://localhost:3000/settings/domain
Please update src/js/components/Settings/SettingsDomain.jsx to include these interface elements. Please note that the API calls haven't been created yet, so all data can be saved locally in a state in the SettingsDomain component:
<img width="557" alt="Screen Shot 2019-07-05 at 10 34 30 AM" src="https://user-images.githubusercontent.com/7756031/60737796-d0082a80-9f10-11e9-902e-2a0f617fe82b.png">

Please note that the blue-ish button should actually be the current We Vote blue button. 

When text has been entered in either box, please show the Cancel and Save buttons like this:
<img width="555" alt="Screen Shot 2019-07-05 at 10 38 32 AM" src="https://user-images.githubusercontent.com/7756031/60737865-1067a880-9f11-11e9-9f8b-1d7706749589.png">

...and this (which will show prior to a person paying for the "Paid Professional" account):
<img width="552" alt="Screen Shot 2019-07-05 at 10 38 59 AM" src="https://user-images.githubusercontent.com/7756031/60737886-1f4e5b00-9f11-11e9-885a-4d577648ddd1.png">

We will be setting up a "Paid Professional" setting, so that when someone has paid for a Professional account, the second setting will show like this:
<img width="547" alt="Screen Shot 2019-07-05 at 10 40 17 AM" src="https://user-images.githubusercontent.com/7756031/60737932-4dcc3600-9f11-11e9-979d-b7ca029d7228.png">

Please note that I have created another issue that goes into detail regarding the "Pro" icon: https://github.com/wevote/WebApp/issues/2420

For errors:
<img width="580" alt="Screen Shot 2019-07-05 at 10 54 59 AM" src="https://user-images.githubusercontent.com/7756031/60738514-6e958b00-9f13-11e9-9d3b-3ee4bae2b262.png">

DaleMcGrew:
Verified!

PR #2429, opened 2019-07-09
Domain Settings

jarodpeachey:
This is the UI for the domain settings, I've inserted code comments where the API calls and such should go, for now everything is stored in the state.

#2419


DaleMcGrew:
This looks awesome Jarod! I hope to have the API updated so we can make this component live later today. I will create another issue for connecting this to the API server. 👍

Issue #2422, opened 2019-07-05
Analytics Settings: Create first draft with interface elements


DaleMcGrew:
http://localhost:3000/settings/analytics
Please update src/js/components/Settings/SettingsAnalytics.jsx to include these interface elements. Please note that the API calls haven't been created yet, so all data can be saved locally in a state in the SettingsAnalytics component:
<img width="682" alt="Screen Shot 2019-07-05 at 11 39 22 AM" src="https://user-images.githubusercontent.com/7756031/60740194-94259300-9f19-11e9-8406-2ab5dc837ea3.png">

When text has been entered into lower box, please show the Cancel and Save buttons like this. Please note that the blue-ish Save button should actually be the current We Vote blue button:
![Screen Shot 2019-07-05 at 11 40 06 AM](https://user-images.githubusercontent.com/7756031/60740221-b5867f00-9f19-11e9-8e14-7229f3881b0e.png)

This is how the buttons would look for the top box, when viewed by a voter using a "Free" account:
![Screen Shot 2019-07-05 at 11 41 33 AM](https://user-images.githubusercontent.com/7756031/60740274-eebeef00-9f19-11e9-937a-a0a8dc6fd6af.png)

...and a person who is already paying for the "Paid Professional" account would see the "Upgrade to Professional" as the We Vote blue button, with the text "Save".

Please note that I have created another issue that goes into detail regarding the "Pro" icon: https://github.com/wevote/WebApp/issues/2420

DaleMcGrew:
Verified. Thank you @jarodpeachey!

PR #2431, opened 2019-07-15
Update styling of analytics settings

jarodpeachey:
#2422 
Created the form components for the analytics setting without API calls as they are not constructed yet

### DIFF
--- a/src/js/components/Settings/SettingsDomain.jsx
+++ b/src/js/components/Settings/SettingsDomain.jsx
@@ -175,10 +175,10 @@ class SettingsDomain extends Component {
     const { classes } = this.props;
 
     return (
-      <Wrapper>
+      <div>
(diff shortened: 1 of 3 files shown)


src/js/components/Settings/SettingsDomain.jsx
```diff
@@ -175,10 +175,10 @@ class SettingsDomain extends Component {
     const { classes } = this.props;
 
     return (
-      <Wrapper>
+      <div>
```
DaleMcGrew: Hi @jarodpeachey, in general, I'm moving towards giving divs names (even if there aren't style associated with them yet) because I think it make the code easier to maintain, and then also easier to add styles if we need to make changes directly within the component in the future.

DaleMcGrew:
👍 Thank you @jarodpeachey!

PR #2432, opened 2019-07-15
Fixed back to header on /value/* path

SailingSteve:
Added warning in config-template, that if you do a hardware reset on the
iOS simulator, the uuid changes.


DaleMcGrew:
👍

PR #2434, opened 2019-07-16
Added the OnePlus and the GalaxyTabS4 to the marketing tests.

ffrancetic:
Added the OnePlus and the GalaxyTabS4 to the marketing tests.


DaleMcGrew:
👍

PR #2441, opened 2019-07-18
Donations are working again via Stripe

SailingSteve:
Fixed prior refactors in DonationCancelOrRefund.jsx, DonateList.jsx and Donate.jsx.
Donate.jsx had morphed into a page within About (with the dark blue header
and footer), but DonateThankYou.jsx had not -- that is now fixed.
In Application.jsx, clear out lint warnings, and now if the if the initial url is `/more/donate` wait for Stripe's js api to load from Stripe, before progressing past the initial loading screen spinner  -- this ensures that the Stripe API is available when you get to the page so that "the buttons work".  Also made a fix to DonateStore.jsx, so error conditions get through to the UI so that they can be displayed.

Resolves wevote/WebApp/issues/2217


DaleMcGrew:
Looks excellent. Thank you @SailingSteve. 👍

PR #2443, opened 2019-07-20
Updated the FacebookSDK in WebApp from 3.2 to 3.3 (the latest one)

SailingSteve:
Made a couple of function names more self documenting
Cleared out the last few lint warnings in Application.jsx
(Facebook SDK updates for the native apps will be in WeVoteCordova)
Updated the candidate page top padding in iOS, to match Android.


DaleMcGrew:
👍

PR #2448, opened 2019-07-23
Fixed yarn/npm conflict -- we need to pick one of these, and stick with it.

SailingSteve:
My vote is to stick with NPM, but it can not be up to individual developers.

Resolves the bogus Travis CI error: "TypeError: Cannot read property 'range' of null"


lazorfuzz:
@SailingSteve @DaleMcGrew I was the one who introduced Yarn into this project and I agree, we should pick one and stick with it.  The purpose of lock files is defeated if developers are using different package managers.  That said, my reasons for favoring Yarn are that [it's faster, although nowadays only marginally,](https://www.dnasir.com/2018/11/26/is-yarn-still-faster-than-npm-in-late-2018/) and that the teams working on [React,](https://github.com/facebook/react) [Vue](https://github.com/vuejs/vue), [Webpack,](https://github.com/webpack/webpack) [Parcel,](https://github.com/parcel-bundler/parcel) [VSCode](https://github.com/microsoft/vscode), [GitHub Desktop, ](https://github.com/desktop/desktop)[Electron,](https://github.com/electron/electron) [Material-UI,](https://github.com/mui-org/material-ui)  [Babel,](https://github.com/babel/babel) etc. all have a yarn.lock file in their repositories.

jarodpeachey:
I would probably stick with NPM unless there is a direct performance benefit, although this is just personal preference.

DaleMcGrew:
👍

PR #2449, opened 2019-07-23
Material UI Version 4.0.0

jarodpeachey:
I updated to Material UI v4.0.0. I had to update the method of exporting withTheme() and withStyles(), as well as updating some components. Everything seems to be working, I checked all the pages. If you find anything, let me know and I can fix it. 👍


DaleMcGrew:
Thank you @jarodpeachey! 👍

PR #2452, opened 2019-07-25
Fix Follow/Toggle component menu

jarodpeachey:
I finally (I think) fixed the Follow/Toggle component menu so it goes away on click.


DaleMcGrew:
👍

PR #2453, opened 2019-07-25
Refactored and simplified the data flow, and the flux reduce store for donations.

SailingSteve:
Use the new API server end points introduced with today's wevote/WeVoteServer/pull/1051 Python PR.
Avoids the need to manually refresh the page to see the latest donation
on the thank you page, and the data that comes into the API server slowly
via Webhooks.
Fixed some styling issues that were introduced with the partial change
over from React-bootstrap, to Material-UI.

### DIFF
--- a/src/js/routes/More/Donate.jsx
+++ b/src/js/routes/More/Donate.jsx
@@ -18,6 +18,7 @@ import VoterStore from '../../stores/VoterStore';
 import WelcomeAppbar from '../../components/Navigation/WelcomeAppbar';
 import Section from '../../components/Welcome/Section';
 import Footer from '../../components/Welcome/Footer';
+import DonateActions from '../../actions/DonateActions';
 
 class Donate extends Component {
   static getProps () {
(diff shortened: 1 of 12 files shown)


src/js/routes/More/Donate.jsx
```diff
@@ -192,8 +194,8 @@ class Donate extends Component {
               <span className="u-no-break">Other Amount</span>
             </Button>
             {this.state.showCustomInput ? (
-              <span>
-                <InputGroup className="mb-3" style={{ width: '50%' }}>
+              <span style={{ display: 'flex', marginTop: 10, marginLeft: 20 }}>
```
DaleMcGrew: Hi @SailingSteve in general I think we should move away from inline styles like this, in favor of Styled components like this, where we replace the `<span` with `<DescriptionText`:
```
const DescriptionText = styled.p`
  font-size: 12px;
  margin: 0;
  @media print {
    font-size: 1.5rem;
  } 
`;
```

DaleMcGrew:
👍 @lazorfuzz we are removing the yarn lock in preparation for transitioning to the new webpack build system.

PR #2463, opened 2019-07-30
Adding sticky header. You still can't change between the sections.

LohitG:
Adding sticky header. You still can't change between the sections.


DaleMcGrew:
👍 Thank you @LohitG!

PR #2468, opened 2019-08-06
Add source map to fix broken line numbers in console

utsab:
https://github.com/wevote/WebApp/issues/2465

Add a source map that is compatible for production build.  This should fix the broken line numbers you see in the console.  Note that the choice of source map is inefficient for the development build (but still much faster than the previous build without webpack). Ideally we can configure the development build to use a special source map which is even more efficient.  We can work on this later.

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -62,4 +62,5 @@ module.exports = {
     open: true,
     writeToDisk: true,
   },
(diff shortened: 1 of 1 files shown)


DaleMcGrew:
👍

PR #2469, opened 2019-08-07
fixed voter welcome image

JamielCapatayan:
https://github.com/wevote/WebApp/issues/2465

image

### DIFF
--- a/src/js/routes/WelcomeForVoters.jsx
+++ b/src/js/routes/WelcomeForVoters.jsx
@@ -44,6 +44,7 @@ import Section, {
   SignUpMessage,
 } from '../components/Welcome/Section';
 import WelcomeAppbar from '../components/Navigation/WelcomeAppbar';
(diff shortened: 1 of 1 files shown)


DaleMcGrew:
Thank you @JamielCapatayan. @SailingSteve will this be an approach that works for Cordova as well?

SailingSteve:
@DaleMcGrew  
That would break cordovaDot() as it exists today for Cordova, but the function could be made smarter to examine the path.  I would go ahead with the merge, and I can fix it (them) after some experimentation.
Steve

SailingSteve:
`import logoDark from '../../../img/global/svg-icons/we-vote-logo-horizontal-color-dark-141x46.svg';`
gets compiled to `/img/global/svg-icons/we-vote-logo-horizontal-color-dark-141x46.svg` and cordovaDot turns that into `./img/global/svg-icons/we-vote-logo-horizontal-color-dark-141x46.svg` which loads properly.

So changes with this pattern should work in Cordova

Issue #2451, opened 2019-07-24
Payment Process: Add payment choice and Stripe processing


DaleMcGrew:
We already have this modal (Desktop) WebApp/src/js/components/Settings/PaidAccountUpgradeModal.jsx and we want to add the following logic to this component, using the switch statement variable "paidAccountProcessStep".
![Screen Shot 2019-07-24 at 2 08 31 PM](https://user-images.githubusercontent.com/7756031/61828897-98f4ad00-ae1c-11e9-972c-28fa38212c2c.png)

Add slide 2 (Desktop). The Stripe processing fields can be added by Steve as he wires up the processing, so you can simply provide a space for the Stripe fields:
<img width="636" alt="Screen Shot 2019-07-24 at 2 04 05 PM" src="https://user-images.githubusercontent.com/7756031/61828619-f76d5b80-ae1b-11e9-81e9-ea0e3530a0e3.png">

After coupon code is applied (Desktop):
<img width="629" alt="Screen Shot 2019-07-24 at 2 05 08 PM" src="https://user-images.githubusercontent.com/7756031/61828672-1a980b00-ae1c-11e9-9efb-792a4af557a1.png">

Final step, after Stripe payment processed (Desktop):
<img width="533" alt="Screen Shot 2019-07-24 at 2 05 54 PM" src="https://user-images.githubusercontent.com/7756031/61828722-356a7f80-ae1c-11e9-9621-a0554ad2c07d.png">

In mobile view, their should be one extra step, since we have to put the plan choice on one slide, and then the processing fields on the next step:
![UpgradeToPro_PaymentPlan_Mobile](https://user-images.githubusercontent.com/7756031/61830379-da3a8c00-ae1f-11e9-9b07-0381a3799f25.png)

Adding the coupon (mobile):
![UpgradeToPro_Coupon_Mobile](https://user-images.githubusercontent.com/7756031/61830392-e6264e00-ae1f-11e9-91b7-ccc67a0ac1e8.png)

Payment (mobile):
![UpgradeToPro_Payment_Mobile](https://user-images.githubusercontent.com/7756031/61830408-ec1c2f00-ae1f-11e9-9a76-e005e9e2ce14.png)

Confirmation (mobile):
![UpgradeToPro_Confirm_Mobile2](https://user-images.githubusercontent.com/7756031/61830663-9005da80-ae20-11e9-9a61-da7f5b8dc72b.png)

DaleMcGrew:
[...]

PR #2474, opened 2019-08-09
Payment Process Modal 

jarodpeachey:
#2451 

Sorry this took so long!

This covers the design of the first 3 stages of the payment process, excluding the final one (the payment summary stage). 

I added test coupon codes to the component state in an array to simulate having multiple promos running at once. I also added the validForProfessionalPlan and validForEnterprisePlan booleans to these coupon objects to simulate some coupons only being valid for one plan, and not the other.

I also made sure that the monthly and annual plan prices will update based on if the user goes back and changes their plan on the first screen. I made sure that the data stays on screen resize as well, and switches between desktop and mobile layout without losing the state.

Let me know if this is good to go, or if I should design the last page before you approve this. I'll demo it during the standup tomorrow so everyone can see it, and we can decide then.


src/index.html
```diff
@@ -20,14 +20,15 @@
     <meta property="og:title"         content="We Vote" />
     <meta property="og:description"   content="View your ballot. Learn from friends. Share your vision." />
     <meta property="og:image"         content="https://WeVote.US/img/global/logos/we-vote-logo-wordmark-vertical-inverted-color-256x256.png" />
-    <link href="/css/loading-screen.css" rel="stylesheet" type="text/css" />
+    <base href="/">
+    <link href="/css/loading-screen.css" type="text/css" />
 
-    <link rel="apple-touch-icon" sizes="180x180" href="/img/global/icons/apple-touch-icon.png">
```
DaleMcGrew: Hi @jarodpeachey, thank you for this pull request! I wanted to check in about the removal of the "link rel=" attributes. I believe we still want these "rel" tags. What is the benefit of removing them?

jarodpeachey: I can double-check, but I believe these helped with making sure that Webpack loaded the styles properly. Once I removed these, the MIME type error no longer showed in the console.

Like I said, I'll double-check and get back to you before or during the standup. 👍

jarodpeachey:
@DaleMcGrew Here's the comparison of the console before and after I remove the rel attributes:

### **With rel attributes**
with_rel


### **Without rel attributes**
without_rel

DaleMcGrew:
Thank you @jarodpeachey this analysis is very helpful. I will roll back the index.html changes, but create a new issue for the team working on webpack, since this is an interesting clue. 👍

PR #2486, opened 2019-08-13
Fix styles issue with Webpack

jarodpeachey:
I fixed the issue with the styling from the _variables.scss file not applying to the site, which affected the background color, heading styles and many other aspects. It turns out that in main.scss, the styles are applied based off of when they are imported. Because the `@import '~bootstrap/dist/css/bootstrap';` occurred after all the other files, it reset the styling of the site.

I don't know if this fixes some of the other issues, I just know it's a major fix as far as styling of the site goes (although it was just one line)

### DIFF
--- a/src/sass/main.scss
+++ b/src/sass/main.scss
@@ -4,6 +4,8 @@
 // 1. Setup
 // ===============================================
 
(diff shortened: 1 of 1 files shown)


DaleMcGrew:
👍

Issue #2480, opened 2019-08-12
Webpack: Problem loading the css for the loading screen


DaleMcGrew:
Problem loading the css for the loading screen:
```
Refused to apply style from 'https://a445e2e0.ngrok.io/css/loading-screen.css' because its MIME type ('text/html') is not a supported stylesheet MIME type, and strict MIME checking is enabled.
```
But when you try to access the style sheet, it says:
`Cannot GET /css/loading-screen.css` so I suspect SASS is not building the file. I found mention of `loading-screen.scss` in this file, around line 112: `WebApp/Gulpfile.js`

DaleMcGrew:
It looks like @jarodpeachey found a fix for this issue with his latest pull request (to be verified) https://github.com/wevote/WebApp/pull/2486
Thank you @jarodpeachey!

DaleMcGrew:
My bad -- my comments should have been on https://github.com/wevote/WebApp/issues/2481
This issue still remains.

DaleMcGrew:
Thank you @utsab, this is resolved!

PR #2489, opened 2019-08-15
Fix for issue #2480: Refactor the way webpack loads scss files.  Prev…

utsab:
…iously webpack was configured to import them directly into the javascript files.  Now webpack will transpile the scss into an explicit css file.

#2480 

Refactor weback to convert .scss files into explicit .css files.  Previously webpack was configured to import the scss files directly into javascript.


DaleMcGrew:
Thank you @utsab! 👍

PR #2494, opened 2019-08-17
Test page to prove StripeElements usage for the PaidAccountUpgradeModal

SailingSteve:
The validation of coupons, and the display of StripeElements is fine,
some of the backend work is ongoing to distinguish between donations and
organization paid subscriptions -- both use Stripe and mostly the same
Python code and SQL tables on the server.

I have a support issue in to stripe on how to recover our plan name code
which does not seem to be in their charge webhook return, but is mentioned
in a [html receipt for the charge which they generate](https://pay.stripe.com/receipts/acct_19sJvMD153UBwRss/ch_1F8YV3D153UBwRss0KKRKr7C/rcpt_FdqBflp8V3DA4lTn95dniG5VUCbP1yt) -- should have the
answer in a few days.

Use https://localhost:3000/more/stripe_elements_test to access page
Only the "25OFF" coupon is defined on the server (plus 'DEFAULT-PROFESSIONAL_MONTHLY' and 'DEFAULT-ENTERPRISE_MONTHLY' for
use when there is no coupon entered by the voter.  See https://github.com/wevote/WeVoteServer/blob/develop/donate/models.py create_initial_coupons().  Once we get the Python Admin UI going for editing coupons, then
we can create as many as we want, and maintain them going forward.

The DonateActions.validateCoupon (planType, couponCode) also returns the list
and discounted prices, for those default coupons (whose coupon code should
not need to be exposed to voters), for default, the discounted and list prices are the same
and can be used to lookup the current prices without having to hard code
them in JavaScript.

further work on wevote/WebApp/issues/2488

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -999,6 +999,18 @@
         "@babel/helper-plugin-utils": "^7.0.0"
       }
     },
+    "@babel/plugin-transform-runtime": {
+      "version": "7.5.5",
+      "resolved": "https://registry.npmjs.org/@babel/plugin-transform-runtime/-/plugin-transform-runtime-7.5.5.tgz",
+      "integrity": "sha512-6Xmeidsun5rkwnGfMOp6/z9nSzWpHFNVr2Jx7kwoq4mVatQfQx5S56drBgEHF+XQbKOdIaOiMIINvp/kAwMN+w==",
+      "dev": true,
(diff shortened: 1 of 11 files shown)


DaleMcGrew:
👍 Also, I have updated the API server with the changes related to this WebApp update.

Issue #2493, opened 2019-08-17
Add Endorsements Tool: Desktop-to-mobile style cleanup


DaleMcGrew:
### Please describe the issue (What happens? What do you expect?)
The mobile display of the "Add Endorsements" tool needs to be reworked to fill the full width of the phone. (Eliminate space to the left and right of the Office card)

See current endorsements:
![Screen Shot 2019-08-17 at 8 30 37 AM](https://user-images.githubusercontent.com/7756031/63214324-d0791100-c0cb-11e9-8dd2-216602607c30.png)

Add new endorsements:
![Screen Shot 2019-08-17 at 8 49 16 AM](https://user-images.githubusercontent.com/7756031/63214335-eab2ef00-c0cb-11e9-9ba6-044917154f57.png)

### Steps to reproduce the problem (1, 2, 3...), including links
1. Sign in (Mobile mode)
2. Go to Settings
3. Under "Your Endorsements" choose the "US 2020 Presidential Primary"
4. Go to "Add Endorsements"

Please make the Office cards fill the full width, and make sure the content area isn't wider than the phone's full width.

This is the file that needs work: WebApp/src/js/components/Settings/BallotItemForAddPositions.jsx

This div (around line 119):
`<div className="BallotItem card"`
...should be converted to a styled div `BallotItemCard` (see the start of work on this near the bottom of this file)

Please verify your changes in both desktop and mobile.

DaleMcGrew:
These changes look great @jarodpeachey! Issue resolved.

Issue #2491, opened 2019-08-17
PaidAccountUpgradeModal: "X" Close-modal disappears in mobile


DaleMcGrew:
When you get the "Choose Your Plan" modal from the settings section, the "X" that closes the modal appears in Desktop. That "X" is missing in mobile mode.

Visible in Desktop:
![Screen Shot 2019-08-17 at 7 12 20 AM](https://user-images.githubusercontent.com/7756031/63212797-6ad25800-c0be-11e9-8f95-b2ecffa037bc.png)

NOT Visible in Mobile:
![Screen Shot 2019-08-17 at 7 12 33 AM](https://user-images.githubusercontent.com/7756031/63212801-7291fc80-c0be-11e9-9d03-45ddb74da598.png)

To open this modal, start from this page: http://localhost:3000/settings/sharing
...and click on "Upgrade to Premium" button.

File is here:
WebApp/src/js/components/Settings/PaidAccountUpgradeModal.jsx

Search for the id `profileClosePaidAccountUpgradeModal`

DaleMcGrew:
Verified. Thank you @jarodpeachey!

PR #2496, opened 2019-08-18
Fix close button on PaidAccountUpgradeModal and update mobile view fo…

jarodpeachey:
…r the add endorsements page

#2491 #2493 
I removed the media query hiding the close button on mobile, and I also updated the add endorsements page for a better view on mobile.

### DIFF
--- a/src/js/components/Filter/FilterBase.jsx
+++ b/src/js/components/Filter/FilterBase.jsx
@@ -28,6 +28,7 @@ class FilterBase extends React.Component {
       lastFilterAdded: '',
       showAllFilters: false,
       selectedFilters: [],
+      fullWidth: PropTypes.bool,
     };
   }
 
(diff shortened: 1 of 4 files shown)


DaleMcGrew:
Hi @jarodpeachey, thank you for these changes! 👍  I am approving this pull request, but could you please explore combining `WrapperFullWidth` and `Wrapper` in FilterBase into a single style (where we define the styles at the bottom of the page) so that the `WrapperFullWidth` styles are used on mobile, and `Wrapper` styles are used in tablet and desktop, using width breakpoints? I don't want to have to declare two copies of FilterBase to accomplish Responsiveness.

DaleMcGrew:
Hi @jarodpeachey, I got a chance to look more closely, and since `Wrapper` and `WrapperFullWidth` in FilterBase only have one very simple difference, you would want to use only one styled component, and pass in a variable. (An example is line 52 in this file: WebApp/src/js/components/Welcome/Section.jsx ) Doing major switch statements that duplicate lines of code (like you have with `this.props.fullWidth`) is something we should avoid whenever possible. Would you be willing to refactor your change to use `this.props.fullWidth` as a variable in `Wrapper`, and remove the major switch statement?

jarodpeachey:
Hey @DaleMcGrew ,
Just submitted these fixes. 

I know that in the past I’ve used switch statements incorrectly like that, so would you like me to go through and use this new method to reduce the number of lines of code in other files?

DaleMcGrew:
Hi Jarod, you are always welcome to re-factor as you feel inspired, but since we have so much new work to do, I would hold off on putting too much time into re-factoring at this point.

Issue #2500, opened 2019-08-21
Values popover has display problem


DaleMcGrew:
Another instance of the problem you solved before @jarodpeachey:
![Screen Shot 2019-08-21 at 7 50 01 AM](https://user-images.githubusercontent.com/7756031/63442761-8011f400-c3e8-11e9-8ac4-3202825bd6ee.png)

PR #2501, opened 2019-08-21
Update issues popover card

jarodpeachey:
#2500 
I updated the width of the popover wrapper so it displays like it used to before the Material UI v4 upgrade


DaleMcGrew:
👍

PR #2506, opened 2019-08-21
Upgraded the way PaidAccountUpgradeModal accepts incoming API variables. I have not updated the API calls to return different values, but instead use mocked up data stubs.

DaleMcGrew:
Upgraded the way PaidAccountUpgradeModal accepts incoming API variables. I have not updated the API calls to return different values, but instead use mocked up data stubs.


DaleMcGrew:
I will update this code to the new OrganizationSubscriptionPlans class.

PR #2508, opened 2019-08-24
Refactor styled components

jarodpeachey:
I found around 4 files that needed refactoring to use props in styled-components instead of using multiple ternary operators.


DaleMcGrew:
These are awesome improvements. Great work! 👍

PR #2516, opened 2019-08-29
Subscription to a paid plan is working, but still need edge case handling

SailingSteve:
Still need to prevent two paid plans for one organization
Unsubscribe from a paid plan needs work.

Dale's version of validateCoupon was incompatible with my current work
and it is commented out.  I should be done by mid first week in
September, at which point, this can be reversed.

Test page for React is https://localhost:3000/more/stripe_elements_test
Coupon management page is http://localhost:8000/organization_plans/plan_list/

Work towards the completion of wevote/WebApp/issues/2488


DaleMcGrew:
👍

PR #2522, opened 2019-08-31
Trigger re-render on VoterGuidesUnderOneValue

jarodpeachey:
Updates to #2513 
I added a componentWillRecieveProps to the VoterGuidesUnderOneValue component, as well as the GuideList component. I also made sure the ValuesList.jsx component did not show the current issue being viewed.

### DIFF
--- a/src/js/components/VoterGuide/GuideList.jsx
+++ b/src/js/components/VoterGuide/GuideList.jsx
@@ -1,4 +1,4 @@
-import React, { PureComponent } from 'react';
+import React, { Component } from 'react';
 import PropTypes from 'prop-types';
 import CandidateStore from '../../stores/CandidateStore';
 import FollowToggle from '../Widgets/FollowToggle';
@@ -9,7 +9,7 @@ import { renderLog } from '../../utils/logging';
 import EndorsementCard from '../Widgets/EndorsementCard';
 import { openSnackbar } from '../Widgets/SnackNotifier';
 
-export default class GuideList extends PureComponent {
+export default class GuideList extends Component {
   static propTypes = {
     ballotItemWeVoteId: PropTypes.string,
     incomingVoterGuideList: PropTypes.array,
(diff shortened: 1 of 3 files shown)


src/js/components/VoterGuide/GuideList.jsx
```diff
@@ -49,18 +49,26 @@ export default class GuideList extends PureComponent {
     const voterGuideList = this.sortOrganizations(nextProps.incomingVoterGuideList, ballotItemWeVoteId);
     this.setState({
       voterGuideList,
-      ballotItemWeVoteId: nextProps.ballotItemWeVoteId,
+      ballotItemWeVoteId: this.props.ballotItemWeVoteId,
```
DaleMcGrew: Hi @jarodpeachey, I can't think of any case where you would want to reassign ballotItemWeVoteId from this.props here. I will edit this back to nextProps unless you some reason for redeclaring that I'm not thinking of.

src/js/components/VoterGuide/GuideList.jsx
```diff
@@ -49,18 +49,26 @@ export default class GuideList extends PureComponent {
...
-      }
+      this.setState({
+        // organizationsWithPositions: orgsWithPositions,
+        filteredOrganizationsWithPositions: orgsWithPositions,
+      });
     });
   }
 
+  shouldComponentUpdate (nextState) {
+    if (this.state.voterGuideList !== nextState.voterGuideList) {
```
DaleMcGrew: With lists and objects, React typically does a shallow compare, which I think would not find voterGuideList had changed even if list items had changed. I found this article that suggests using a function that compares the length of the lists, and other quick approaches: https://developmentarc.gitbooks.io/react-indepth/content/life_cycle/update/using_should_component_update.html

src/js/routes/Values/ValuesList.jsx
```diff
@@ -28,8 +30,22 @@ export default class ValuesList extends Component {
...
+
+    this.setState({ currentIssue: this.props.currentIssue });
+  }
+
+  componentWillReceiveProps (nextProps) {
+    this.setState({ currentIssue: nextProps.currentIssue });
   }
 
+  shouldComponentUpdate (nextState) {
+    if (this.state.currentIssue !== nextState.currentIssue) {
```
DaleMcGrew: I would suggest comparing the currentIssue weVoteId here, instead of the whole object.

DaleMcGrew:
👍

PR #2523, opened 2019-09-03
Subscriptions that occurred near midnight GMT, were showing payments in the prior day.

SailingSteve:
Completes all the parts of wevote/WebApp/issues/2488 that can be finished based
on the architecture that I envisioned.


DaleMcGrew:
👍

Issue #2359, opened 2019-06-20
Once the tabs are clicked on the position page the page returns to the top. (iPhone SE and smaller screens)


ffrancetic:
### Please describe the issue (What happens? What do you expect?)
Once the tabs are clicked on the position page the page returns to the top. (iPhone SE and smaller mobile screens). It should just stay fixed and the info should be loaded at the spot.
### Steps to reproduce the problem (1, 2, 3...), including links
Link: https://quality.wevote.us/pdamerica/positions

<img width="249" alt="Screenshot 2019-06-20 at 15 14 20" src="https://user-images.githubusercontent.com/36188818/59885027-cad2a980-936e-11e9-8451-6513777399f3.png">
<img width="247" alt="Screenshot 2019-06-20 at 15 14 32" src="https://user-images.githubusercontent.com/36188818/59885029-d0c88a80-936e-11e9-95fb-48352b9b817c.png">

SailingSteve:
No testing has been done for 4" screens on iOS, since iPhone 6 was the minimum.

DaleMcGrew:
This works very well @jarodpeachey, great work! And thank you.

PR #2528, opened 2019-09-06
Fix organization tabs causing page refresh

jarodpeachey:
#2359
Update the historyPush function around line 172 to use a history.pushState function, which changes the URL but does not rerender.


DaleMcGrew:
👍 Thank you @jarodpeachey!

PR #2529, opened 2019-09-06
Fix values page error; variables were not named the same

jarodpeachey:
There was an error with the values page not displaying the list
There seemed to be an error in the naming of the variables and reassigning them based on the displayOnlyIssuesNotFollowedByVoter prop. It should be working properly now.


DaleMcGrew:
Hi @JarodPeachey did you get a chance to make sure that each individual value page is working as well?

jarodpeachey:
@DaleMcGrew Yeah, it works with each of the individual ones as well. 👍

DaleMcGrew:
👍

Issue #2503, opened 2019-08-21
Values listing page: Empty state display


DaleMcGrew:
On value pages that don't have any endorsements, like this one:
https://quality.wevote.us/value/green_party_politicians
"We don't have any endorsements for Green Party Politicians."

Please implement an "Empty" page like this one:
![Screen Shot 2019-08-21 at 9 19 22 AM](https://user-images.githubusercontent.com/7756031/63449364-d6852f80-c3f4-11e9-81fd-1e647ad0bcdd.png)

DaleMcGrew:
Looks great, thank you @jarodpeachey!

PR #2535, opened 2019-09-09
Update to include empty values component

jarodpeachey:
#2503 
I created a component for if there are no endorsements for a certain value. I did NOT get the button to work, so If someone else could hook that up to go to the endorsements, that would be great.


DaleMcGrew:
👍

PR #2539, opened 2019-09-13
Update width of popover for issues from MaterialUI upgrade

jarodpeachey:
This fixed an issue with the issues popover left over from the Material UI V4 upgrade


DaleMcGrew:
👍

Issue #2545, opened 2019-09-17
Google Address Auto-Complete Stopped working


DaleMcGrew:
Google Address auto complete, which should be suggesting complete addresses when you type into this input box:
![Screen Shot 2019-09-17 at 10 10 05 AM](https://user-images.githubusercontent.com/7756031/65064646-bb470a80-d935-11e9-80a3-db3741295349.png)

...has stopped working for some reason.

I have confirmed that the requests are making it to our Google Developer Console, but I do notice they are for v1beta1, which I wonder if it might have changed something? I don't know where the version we use is set.

This component is where the autocomplete is set up:
WebApp/src/js/components/AddressBox.jsx

We call that AddressBox component from WebApp/src/js/components/Widgets/EditAddressInPlace.jsx

In the past, there was a problem where the `pac-container` z-index had to be increased so it wasn't hidden behind the modal:
WebApp/src/sass/overrides/_overrides.scss

Now it doesn't seem like `_placeChanged` is getting fired, so maybe there is a problem with the inputRef that binds addressAutocomplete to the InputBase (see id "addressBoxText").

Looking through the history of this file might find some clues.

This is the documentation: https://developers.google.com/maps/documentation/javascript/places-autocomplete

DaleMcGrew:
Resolved!

PR #2548, opened 2019-09-20
fix z-index for .pac-container class

flurmbo:
#2545 

Looks like the auto-complete results z-index conflicted with the z-index of the Material UI dialog modal, which has a higher z-index than the previous Bootstrap modal's z-index. As you can see in the accompanying screenshot, it needs a bit of styling to match up with the new component!

Screenshot from 2019-09-20 17-28-33


DaleMcGrew:
Great catch! Thank you @flurmbo. 👍

Issue #2310, opened 2019-06-14
Add "Back" navigation to top of voter guide pages


DaleMcGrew:
### Please describe the issue (What happens? What do you expect?)
Right now in mobile and desktop, if you click on an Organization from the candidate page, there is no easy way (in the mobile apps especially) to get back to the candidate you were just looking at.

We currently have a scheme in place for the Ballot / Office / Candidate section with URL's like this that tell the "Back" button what to do:

Office page, with back to default ballot (btdb) link:
https://quality.wevote.us/office/wv02off28320/b/btdb/

Candidate page, with back to default office (btdo) link:
https://quality.wevote.us/candidate/wv02cand55916/b/btdo/

Candidate page, with back to default ballot link:
https://quality.wevote.us/candidate/wv02cand55916/b/btdb/

Please update the Voter Guide header for these scenarios:
1)
Arrive directly on a voter guide page: https://quality.wevote.us/bradyqg
No change to header behavior.

2)
Arrive on the voter guide page with "back to" variables in route, like:
https://quality.wevote.us/bradyqg/btcand/wv02cand55916/b/btdb/
Change the Voter guide header to have a Back to button + a sign in link. This text on the end "wv02cand55916/b/btdb/" would make the Back to button go to: https://quality.wevote.us/candidate/wv02cand55916/b/btdb/

More examples:
```
<Route path=":twitter_handle/btcand/:back_to_candidate_we_vote_id/b/:back_to_variable/" component={TwitterHandleLanding} />
<Route path=":twitter_handle/btmeas/:back_to_measure_we_vote_id/b/:back_to_variable/" component={TwitterHandleLanding} />
```

3)
Make sure to set up for measures as well.

Also update the links to organizations from the position list under candidates, to include these urls with back to variables.

jarodpeachey:
Hey @DaleMcGrew , I've been trying to figure this out, but I can't find where the params are set, such as the back_to_variable or the office_we_vote_id. Do you know where these are set?

DaleMcGrew:
Implemented!

PR #2551, opened 2019-09-23
Back to candidate buttons

jarodpeachey:
#2310 
The organization pages now display a back to link to the candidate. However, I can't figure out the measures, as it's throwing this error when I go onto a measure page: ```cannot read property of candidate_we_vote_id```. What's weird is this error is coming from the PositionItem.jsx file, NOT the MeasureItem.jsx file.


DaleMcGrew:
Thank you @jarodpeachey, I will check it out! 👍

Issue #2456, opened 2019-07-26
Expanding tags on candidate page has no function when tags are not overflowing


flurmbo:
### Please describe the issue (What happens? What do you expect?)

On a candidate's page, there is a list of tags beneath the candidate's name. There is a three dotted button that expands these tags when they would otherwise overflow their container. However, the button is still present even when the tags do not overflow the container, making the button useless. See the gif below for a clear demonstration of this.

![expandingissuesbug](https://user-images.githubusercontent.com/35127144/61972228-accc1a80-afa6-11e9-8109-5d05613a6bbc.gif)

DaleMcGrew:
Resolved! Thank you @flurmbo and @lazorfuzz.

PR #2559, opened 2019-09-30
[#2456] Hide issues tags when not overflowing

lazorfuzz:
#2456 
* Stores an object with each issue's width in the parent component
* Stores the total width of the issue list in the parent component state
* Passes a subtractTotalWidth method to each child issue component
* When the child issue component renderers, call the subtractTotalWidth method
* Parent component updates totalRemainingWidth state each time the subtract handler is called


DaleMcGrew:
Thank you @lazorfuzz this looks great. I wanted to flag this for @flurmbo as another way of approaching the "three dots expansion" solution you proposed. 👍

Issue #2562, opened 2019-09-30
Send SMS Verification Code: New Designs


DaleMcGrew:
Please add this input box to WebApp/src/js/components/Settings/SettingsAccount.jsx below the Sign in with Email Link box:
![Screen Shot 2019-09-30 at 8 50 40 AM](https://user-images.githubusercontent.com/7756031/65894930-74a6d680-e35f-11e9-8c65-a899a29b97db.png)

For now, when you click "Send Verification Code", open this new file within the modal: WebApp/src/js/components/Settings/SettingsSMSVerify.jsx
![Screen Shot 2019-09-30 at 8 54 26 AM](https://user-images.githubusercontent.com/7756031/65895197-eed75b00-e35f-11e9-9db3-62e5bb984687.png)

Text that is pasted should fill all 6 boxes, and activate the "Verify" button. If voter enters a number, move the focus to the next box. If possible, when all 6 digits have been filled in, auto close the keyboard. (Question: Should this be implemented as a single input box, and just visually appear to be 6 separate boxes?)
<img width="329" alt="Screen Shot 2019-09-30 at 9 01 33 AM" src="https://user-images.githubusercontent.com/7756031/65895882-f1868000-e360-11e9-8ea8-a26d692a5ab7.png">

After clicking the Verify button, if the number is incorrect, show this:
<img width="329" alt="Screen Shot 2019-09-30 at 9 02 53 AM" src="https://user-images.githubusercontent.com/7756031/65895965-2266b500-e361-11e9-8c8a-83261f5ea99e.png">

When a voter clicks on any of the numbers after the error state, either all values should be cleared out, or the number the voter clicked on.

After 5 tries that are incorrect, the boxes should disable, and the following error message shown. Notice the change in the "back to" text ("Try another sign in method?").
<img width="328" alt="Screen Shot 2019-09-30 at 9 07 03 AM" src="https://user-images.githubusercontent.com/7756031/65896284-c6506080-e361-11e9-9a08-17077d24aba5.png">

DaleMcGrew:
Hi @jarodpeachey, after speaking with Ahmed about the impact of the mobile keyboard (and how it takes up the bottom 1/2 of many phone screens), we'd like to make the "Verify" button the top button in all of the above designs. 

Additionally, in mobile mode, when the focus is on the input box (and when the keyboard would pop up), we'd love to have a way to reduce the vertical spacing of all of the elements, so the page "scrunches" up like this:
[...]

PR #2565, opened 2019-10-03
Sms updates

jarodpeachey:
#2562 
This includes error state for the form, as well as a disabled verify button if all the fields are not filled in. I also compressed the component if the keyboard on mobile appears.


DaleMcGrew:
This looks excellent @jarodpeachey! 👍 Did you get a chance to test clicking on the first input digit, and then pasting? The desired result is to have the first six digits fill in, and then any remaining characters be ignored.

Issue #2476, opened 2019-08-09
Payment Process: Confirmation Screen


DaleMcGrew:
We already have this modal (Desktop) WebApp/src/js/components/Settings/PaidAccountUpgradeModal.jsx:
![Screen Shot 2019-07-24 at 2 08 31 PM](https://user-images.githubusercontent.com/7756031/61828897-98f4ad00-ae1c-11e9-972c-28fa38212c2c.png)

We want to format the final step, after Stripe payment processed (Desktop):
<img width="533" alt="Screen Shot 2019-07-24 at 2 05 54 PM" src="https://user-images.githubusercontent.com/7756031/61828722-356a7f80-ae1c-11e9-9621-a0554ad2c07d.png">

This is the mobile design:
![UpgradeToPro_Confirm_Mobile2](https://user-images.githubusercontent.com/7756031/61830663-9005da80-ae20-11e9-9a61-da7f5b8dc72b.png)

jarodpeachey:
@SailingSteve @DaleMcGrew Is this ready for me to work on yet without causing merge conflicts in the paymentProcessModal.jsx file?

DaleMcGrew:
Hi @jarodpeachey, I'm really glad you asked. I'm in the middle of hooking that file up to the coupon API, so it might be better to wait until I'm done to work on this issue.

jarodpeachey:
Ok, no problem! Are there any other issues you would like me to start on?

DaleMcGrew:
Hi Jarod, in order to work on this Issue, go into this file:
WebApp/src/js/components/Settings/PaidAccountUpgradeModal.jsx
...and change line 71 to `paymentProcessed` instead of `choosePlan`:
```paidAccountProcessStep: 'paymentProcessed'```
You will then be able to work on the case block starting on line 936.

DaleMcGrew:
It looks great. Thank you @jarodpeachey!

PR #2572, opened 2019-10-10
Payment confirmation screen

jarodpeachey:
#2476 
I styled the payment confirmation screen. I put placeholder data in some of the fields, such as the payment method and period, because I don't know where those are defined.


DaleMcGrew:
Thank you @jarodpeachey! 👍

Issue #2574, opened 2019-10-11
Search Box Component: Style conflict between two places


DaleMcGrew:
When looking at Measure detail page, or Candidate detail page, the search box has two problems:
1) It is shorter than the other Filter elements. 
2) It opens so wide that it pushes the other filter elements out of the content area when the browser is wide.
<img width="815" alt="Screen Shot 2019-10-11 at 10 24 41 AM" src="https://user-images.githubusercontent.com/7756031/66679332-cd5d5580-ec22-11e9-8cea-6bcb9a626cbd.png">

We use the same component for the top of the ballot, so we might want to use Prop settings to specify the styling behavior we want in each context.
![Screen Shot 2019-10-11 at 12 30 57 PM](https://user-images.githubusercontent.com/7756031/66679433-03023e80-ec23-11e9-9d1d-bf8040a2f873.png)

One comment on the search box width: I think the width can be significantly reduced -- probably down to 50% of its current width.

Please test across a variety of screen sizes, including mobile.

DaleMcGrew:
This is a great solution! Thank you @lazorfuzz.

PR #2575, opened 2019-10-12
[#2574] BallotSearch width changes

lazorfuzz:
#2574

@DaleMcGrew This change still does not look great on iPhone 5. We might have to think about putting the search bar on a new line on iPhone 5.


DaleMcGrew:
👍

Issue #2557, opened 2019-09-28
Endorsements pop-down spacing problem


DaleMcGrew:
On this page: https://quality.wevote.us/measure/wv02meas925/b/btdb/
The Pop-down from Endorsements has a spacing problem, including the ballot measure name being too long. Can we expand the dark blue header with the ballot name to have enough space to show the full name and still look good?
![Screen Shot 2019-09-28 at 12 22 17 PM](https://user-images.githubusercontent.com/7756031/65821500-2375d600-e1eb-11e9-9aab-0fb582d0e2d6.png)

Also on this page, the "x" is too low, and text margins are wrong:
https://quality.wevote.us/candidate/wv02cand56709/b/btdb/
![Screen Shot 2019-09-28 at 12 26 28 PM](https://user-images.githubusercontent.com/7756031/65821509-4c966680-e1eb-11e9-9ffe-c082a001237e.png)

On this page:
https://quality.wevote.us/office/wv02off28916/b/btdb/
![Screen Shot 2019-09-28 at 12 27 36 PM](https://user-images.githubusercontent.com/7756031/65821514-68017180-e1eb-11e9-8b50-299d543bea1c.png)

DaleMcGrew:
Hi @jarodpeachey, this problem is mostly finished, but I notice that the name of the measure is only showing the second line against that blue background. The first and third lines are floating above and below the blue background. Could we have the blue header background expand show show the full measure name?
![Screen Shot 2019-10-13 at 5 16 14 PM](https://user-images.githubusercontent.com/7756031/66724630-31c41480-eddd-11e9-8780-8969fc5c0fd8.png)

DaleMcGrew:
Fixed! Great work @jarodpeachey.

PR #2576, opened 2019-10-15
Adjust popover header to be dynamic height

jarodpeachey:
This is another update to #2557.

I updated the popover header to adjust height based on the length of the title text. I also made sure the title text does not block the close button


DaleMcGrew:
👍

PR #2579, opened 2019-10-16
Added endorsement-extension path for WebPack CopyPlugin

SailingSteve:
Finishes:  Added three icons for the WebApp will serve for the EndorsementExtension


DaleMcGrew:
👍

Issue #2582, opened 2019-10-17
SettingsVerifySecretCode: Error condition display


DaleMcGrew:
Update WebApp/src/js/components/Settings/SettingsVerifySecretCode.jsx to change the display in these error conditions:

<img width="316" alt="Screen Shot 2019-10-17 at 11 10 23 AM" src="https://user-images.githubusercontent.com/7756031/67035584-c628c280-f0ce-11e9-86e7-0f199ab251c8.png">

<img width="314" alt="Screen Shot 2019-10-17 at 11 10 31 AM" src="https://user-images.githubusercontent.com/7756031/67035591-c9bc4980-f0ce-11e9-886b-a8cd72cd8cd7.png">

DaleMcGrew:
Verified. Great work @jarodpeachey!

Issue #2581, opened 2019-10-17
SettingsVerifySecretCode: Accept pasted digits


DaleMcGrew:
Update WebApp/src/js/components/Settings/SettingsVerifySecretCode.jsx to accept digits pasted in, filling the remaining input boxes with each digit until all input boxes are filled up. So if you put the focus on the first box, and paste, the first 6 digits of whatever is in the buffer should fill the first 6 boxes.
![Screen Shot 2019-10-16 at 9 39 45 PM](https://user-images.githubusercontent.com/7756031/66978138-8249a500-f05d-11e9-983b-2817df5449a3.png)

To get to this code, go to the sign in (or account settings "Security & Sign In"). Enter an email address and click "Email Verification Code". This will open "SettingsVerifySecretCode".
![Screen Shot 2019-10-16 at 9 38 17 PM](https://user-images.githubusercontent.com/7756031/66978112-69d98a80-f05d-11e9-827d-c6fe95e1b2a1.png)

DaleMcGrew:
Verified! Thank you @jarodpeachey.

PR #2589, opened 2019-10-21
Sign In Secret Code Updates

jarodpeachey:
#2581 and #2582 

I updated the sign in modal to accept pasted codes. It displays an error if the code is not enough digits, and it also filters out any non-digit characters. I also created the error state and show the error message.


DaleMcGrew:
👍

PR #2593, opened 2019-10-23
Clear console logs, attempt fix for iOS not focusing on next input

jarodpeachey:
This removes some console logs and fixes an error state issue. It's also an attempt at fixing the iOS issue where it does not auto-focus the next input.


DaleMcGrew:
👍

PR #2599, opened 2019-10-23
Cordova removed the UUID feature, since it did not improve simulation

SailingSteve:
Apple has a few simulators with the wrong screen size, it probably does
not matter for the great majority of developers.  They also dropped most of
their legacy simulators, and even the more recent X models, since the 11
models have the same screen size.


DaleMcGrew:
👍

PR #2603, opened 2019-10-25
Webpack (purposefully!) broke __filename, so added static names everywhere it had been used

SailingSteve:
Needed render logging for debugging login with Cordova
WebPack's decision seems dumb, and they did not explain their reasoning (its not like JavaScript is a compiled language, where their decision could make some minor security difference):
https://github.com/webpack/webpack/issues/5

### DIFF
--- a/src/js/Application.jsx
+++ b/src/js/Application.jsx
@@ -265,7 +265,7 @@ class Application extends Component {
   }
 
   render () {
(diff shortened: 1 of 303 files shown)


DaleMcGrew:
Thank you @SailingSteve for prioritizing maintenance of good logging! 👍

Issue #2608, opened 2019-10-31
Logo shows up as "morphed" on Upload Your Logo page


DaleMcGrew:
1. Sign in.
2. Navigate to:
Your Settings > Sharing > Upload Your Logo

3. Upload this image:
https://www.pride.be/wp-content/uploads/2017/10/The-Belgian-Pride-vzw-asbl-Logo_2018_RGB_PNG-5-300x250.png

4. The image displayed is warped, even though the image that we save is correctly formatted.
![Screen Shot 2019-10-31 at 1 59 45 PM](https://user-images.githubusercontent.com/7756031/67985993-47ac4480-fbe7-11e9-9af8-05d02ac612e9.png)

DaleMcGrew:
Verified.

PR #2617, opened 2019-11-04
Add max height to image

jarodpeachey:
#2608 

I added a max-height to the image container, and now the image prioritizes height, and the width will scale with it.


DaleMcGrew:
👍

PR #2625, opened 2019-11-06
Fixed redirect after twitter login issue on Android

SailingSteve:
And if it also happened on Apple, it doesn't anymore.


DaleMcGrew:
Good catch @SailingSteve! 👍

PR #2628, opened 2019-11-07
iOS Simulators are now returning device code

SailingSteve:
Switched over to using the code, instead of guessing by screen size


DaleMcGrew:
👍

PR #2638, opened 2019-11-09
Fix fa icons for facebook and twitter in cordova

SailingSteve:
One more 'file moving' repair due to WebPack changeover, bootstrap-social.css


DaleMcGrew:
👍

PR #2646, opened 2019-11-13
Two null object dereference errors in Facebook login fixed

SailingSteve:
Facebook login failure message now gets displayed in the SignInModal
These changes have been submitted to Google Play as
android:versionCode="100019" android:versionName="1.2.4"
this new version is now live.
This version does not include changes from wevote/WebApp/pull/2641


DaleMcGrew:
👍

PR #2656, opened 2019-11-21
Update headings on values pages

jarodpeachey:

I updated some of the headings across the site (only headings inside of cards) to use the same font-size, weight and color as the headings on the friends page that I just implemented in the past week.

Feel free to disregard this if you don't like how they look; I just thought it would help with the consistency of the site.


DaleMcGrew:
This is awesome -- I really appreciate your initiative @jarodpeachey! 👍

PR #2662, opened 2019-11-23
Facebook login changes for Android

SailingSteve:
Restored webAppConfig.ENABLE_NEXT_RELEASE_FEATURES, but now it defaults
to true if the config element is not defined (for quality server, etc).
Updated attributions.


DaleMcGrew:
👍

PR #2667, opened 2019-12-02
Added back in Zendesk help in Cordova, but without the Widget

SailingSteve:
I could get the widget working, but after a half day of attempts, was
unable to style it sufficiently to be workable, especially on the full
screen iframe where you enter data and especially on iPhones with
notches.  So I added an icon and link to desktop (viewed in a Cordova
"tab" with a done/return button) from the FooterBar.

Resolves wevote/WebApp/issues/2665

Checked in without honoring pre-commit hook errors for unrelated source files.


DaleMcGrew:
Thank you @SailingSteve for exploring different ways of including the Zendesk help system. I think your solution is exactly right. 👍

Issue #2643, opened 2019-11-11
Friends Promo Box


DaleMcGrew:
Please duplicate this promo box from the "Values" section and place the copy in Friends section.
![Screen Shot 2019-11-11 at 11 06 03 AM](https://user-images.githubusercontent.com/7756031/68613401-85289180-0473-11ea-9c44-2861a8942115.png)

It should be at the top of the page in mobile mode, and in the right column in desktop mode.
Please also add an "X" in the upper right corner when in mobile mode (but not in desktop) so it can be closed (since space in mobile mode is at such a premium). When the "X" is clicked, please set a cookie ("friend_section_box_closed") so we know not to show this promo box again in mobile. This friend promotion box should always show in desktop, irregardless of whether this "X" was clicked in mobile mode. Search for "location_guess_closed" in WebApp to find the code that does this for the address reminder box.

DaleMcGrew:
This works great @flurmbo, thank you!

PR #2668, opened 2019-12-02
Phil nov 17

flurmbo:
#2643
Created FriendsPromoBox component by wrapping the Testimonial component. The cookie should behave as expected and the change in styles from mobile to desktop should as well. I added the padding-right CSS rule as discussed last week by adding an optional "textStyle" prop to the Testimonial that styles that element.


DaleMcGrew:
👍 Thank you @flurmbo!

PR #2669, opened 2019-12-02
Friending System Navigation Updates

jarodpeachey:
I updated the friends navigation to have different state for mobile and desktop in case someone switches between landscape and portrait on a tablet or large phone.

I also made some changes so the routes are more consistent with one another in styling.


DaleMcGrew:
👍

PR #2674, opened 2019-12-11
WebApp is now a progressive web app via Google Workbox.js

SailingSteve:
The basics are in place to make some candididate/measure selections, then
if you go offline, you should be able to get to My Values and Vote pages.
Added a PROGRESSIVE_WEB_APP.MD that describes how to maintain the cache.
Configuration for Workbox is in a new ServiceWorker.js file.
Tested on Chrome, Safari, and Firefox -- no browser specific changes were needed.


DaleMcGrew:
👍

Issue #2661, opened 2019-11-22
Friend Requests & Sent Requests: Listing Format


DaleMcGrew:
Please implement the new design (below) on these pages:
http://localhost:3000/friends/requests
http://localhost:3000/friends/invitationsbyme

See "<FriendInvitationsSentToMePreview" and "<FriendInvitationsSentByMePreview" (which both have "<FriendInvitationList") in WebApp/src/js/routes/Friends.jsx

...and WebApp/src/js/routes/Friends/FriendInvitationsSentToMe.jsx and and WebApp/src/js/routes/Friends/FriendInvitationsSentByMe.jsx which also has "<FriendInvitationList"

Both of these use "<FriendInvitationDisplayForList":
WebApp/src/js/components/Friends/FriendInvitationDisplayForList.jsx
...and "<FriendInvitationEmailForList":
WebApp/src/js/components/Friends/FriendInvitationEmailForList.jsx
We need two different components because "FriendInvitationDisplayForList" is an invitation between two signed in voters, and "FriendInvitationEmailForList" is between a signed in voter and an email account that hasn't been verified yet.

Notes about test accounts you can use while you are working on Friending. You will need to sign in with one of these test accounts in order to work on this: https://docs.google.com/document/d/1TcChEI92Vmg-Fex_4X5aBzWClyFCOMAAps8CSBecONk/edit?usp=sharing

INCOMING FRIEND REQUESTS:
![Screen Shot 2019-11-22 at 11 40 55 AM](https://user-images.githubusercontent.com/7756031/69455289-f8a48b80-0d1c-11ea-82d7-6de842a43390.png)

REQUESTS YOU SENT:
![Screen Shot 2019-11-22 at 12 36 40 PM](https://user-images.githubusercontent.com/7756031/69458686-cb5bdb80-0d24-11ea-9e5e-163936a7f192.png)

Feel free to update all of the components in the Friend section to Styled components as you work.

DaleMcGrew:
Verified!

PR #2678, opened 2019-12-12
Listing format for requests

jarodpeachey:
#2661 

I updated the listing styles for the sent requests components, as well as started on the styling for the requests, as the components are somewhat the same.

I used placeholder data for the info, and the cancel button does not actually do anything at the moment.


DaleMcGrew:
👍

PR #2683, opened 2019-12-15
Refactor of BallotItemCompressed to eliminate deprecated componentWillReceiveProps (2)

SailingSteve:
Also removes componentDidMount, shouldComponentUpdate, the constructor, all state variables and code to maintain state.
Removes 100 out of 135 lines, and does the same thing.
Removed passed in props, that we needed in the past, or might need someday -- if we need them in the future, they will fit in
logically without needing to put props in state (really avoid that!) or create state at all.
State will still be needed on some other classes that call APIs.
----- Sunday, Performance Improvements (more to follow)
Clear travis/npm test errors in 3 Friending classes
Added a comment in BallotStore that calling Actions from Stores should be avoided -- it often causes unneeded cascades of zero value API calls, and can be avoided with a little extra work.  That comment applies to all Actions from Stores, but BallotStore is in the center of BallotItemsRetrieve being called 10x on startup.
Removed CSSPlugin, EasePack, TweenLite and pooper, which we no longer use from index.html, and the (sometime duplicated) inclusions from package.json, also removed immutable and react-intl which we also no longer use.
Some changes to serviceWorker to improve performance.
The loading of stripe has been made asynchronous, so the only time anyone should be waiting for it to load, is if they start the app on the donate page.


DaleMcGrew:
Thank you Steve for driving these upgrades! I look forward to discussing these changes in standup. 👍

PR #2685, opened 2019-12-16
Update friends listing format and add "No Friends" messages to all components

jarodpeachey:
This is more work on the friends system. I updated the iPhone 5 layout for all the listings, as well as added the listing styles to the current friends page. I also created a new component, MessageCard.jsx, to take in a message, button text and button URL to give the user somewhere to go if there is no data for the selected friend tab.


DaleMcGrew:
👍 Thank you @jarodpeachey.

PR #2690, opened 2019-12-18
ENABLE_WORKBOX_SERVICE_WORKER will default to false if not defined, removed cssnano which was only used by gulp for non-production builds

SailingSteve:


DaleMcGrew:
👍

PR #2698, opened 2019-12-24
Removed unused library components from lodash,  moment, material-ui, & bootstrap to speed up initial load of bundle

SailingSteve:
This PR touches many files, but almost entirely in the import statements.

Reduced the size of the "npm run prod" production bundle by about 10% to 2.6 MB (which is still very large).

DO NOT import entire libraries into a class and then only use one component, since that adds every component in the library into the bundle.js  --  see ["Cherry Picking" in DEBUGGING_TOOLS.md](https://github.com/wevote/WebApp/blob/develop/docs/working/DEBUGGING_TOOLS.md) for a more detail.

Reduced the size of the "npm run prod" production bundle by about 10% to 2.6 MB (which is still very large).

The next step is breaking the pages and components into modules, and lazy loading them as needed.
See [Speeding up first page load wevote](https://docs.google.com/document/d/1Ni_xC8OX8ZAZX-lFk4sHyje3G32X3ovxYfBLAYT-hb8/edit?usp=sharing)


DaleMcGrew:
Github won’t let me merge this on the phone, but feel free to merge it if you’d like to. Otherwise I will merge it when I get back to my computer.

SailingSteve:
@DaleMcGrew 
Can you verify that the FriendsOnlyIndicator.jsx works before merging?  Travis is choking on the react-bootstrap ToolTip, but it compiles fine, shows no error in "npm test"  I'll test more tomorrow

DaleMcGrew:
@SailingSteve I am testing locally, and FriendsOnlyIndicator seems to be working just fine. I'm even seeing tool tips. I'm not seeing any errors in `npm test` nor in the console log. Your upgrade looks totally successful to me.

PR #2710, opened 2019-12-29
fix Nunito link + add bold weight - fixes #2708

### LINKED ISSUE #2708: New font Nunito not honoring "font-weight"
We have switched to the font "Nunito Sans", but neither it nor "Nunito" are responding to attempts to change the font-weight. From searches on the internet, it seems like this should be working...

This site has been updated with the new fonts: https://quality.WeVote.US/ballot

Please see where we specify our fonts:
src/sass/base/_fonts.scss
src/sass/base/_variables.scss

I did not make an updates to this folder--should we?
src/sass/base/fonts

ixley:
Fixes #2708
Added `700` (bold) font weight to Google Fonts link for Nunito

### DIFF
--- a/src/sass/base/_fonts.scss
+++ b/src/sass/base/_fonts.scss
@@ -3,7 +3,7 @@
 // ----------------------------------------------------------------------------
 
 // @import url(https://fonts.googleapis.com/css?family=Nunito+Sans);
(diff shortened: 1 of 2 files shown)


DaleMcGrew:
👍 Thank you @ixley!

PR #2745, opened 2020-01-14
DO NOT MERGE:  Progress toward a cordova app update

SailingSteve:
Many (of the typical) header adjustments, with more to come.
Media queries often don't work in Cordova, and have to be patched around.
Autosaving is bad news in Cordova, since we have to remove the footer in order to avoid `Unable to simultaneously satisfy constraints` errors when the software keyboard appears, and these errors sometimes crash the app.
Removed the use of `webAppConfig.ENABLE_NEXT_RELEASE_FEATURES` since this release will be using the latest from git upstream/develop.
Added a save button, and removed autosaving when isCordova(), in SettingsWidgetFirstLastName.jsx and removed the deprecated componentWillMount method.
Added an error boundary to Values to catch an exception thrown by code in develop -- the underlying problem was resolved by someone else.


DaleMcGrew:
Hi @SailingSteve we still need ```webAppConfig.ENABLE_NEXT_RELEASE_FEATURES``` because sharing won't be done in time for this release.

SailingSteve:
Ok, ill add it back in for sharing

PR #2747, opened 2020-01-15
Progress toward a Cordova app update (2nd try with git)

SailingSteve:
Many (of the typical) header adjustments, with more to come.
Media queries often don't work in Cordova, and have to be patched around.
Autosaving is bad news in Cordova, since we have to remove the footer in order to avoid `Unable to simultaneously satisfy constraints` errors when the software keyboard appears, and these errors sometimes crash the app.
Added a save button, and removed autosaving when isCordova(), in SettingsWidgetFirstLastName.jsx and removed the deprecated
componentWillMount method.
Added an error boundary to Values to catch an exception thrown by code in develop -- the underlying problem was resolved by someone else.


DaleMcGrew:
👍

PR #2749, opened 2020-01-15
Progress toward a cordova app -- fixed friends__heading

SailingSteve:
Also added header adjustments for HeaderBackTo called from Settings


DaleMcGrew:
👍

PR #2755, opened 2020-01-17
Progress toward a cordova app -- all iPhones first pass

SailingSteve:
Noticed that twitter sign-in is no longer working.


DaleMcGrew:
👍

PR #2759, opened 2020-01-18
Travis/enzyme fix: Remove '/esm/' from material-ui and bootstrap import paths

SailingSteve:
Webpack was smart enough to not need the '/esm/' path component, and Enzyme could not handle them.
Fixed tests/component/tests.js to handle styled-components in HeaderBarLogo


DaleMcGrew:
👍

PR #2765, opened 2020-01-21
Navigate to WebApp from Cordova, when we need to change paid subscription level.

SailingSteve:
Changes on settings Domain, Sharing, and Subscription pages.


DaleMcGrew:
👍

PR #2781, opened 2020-01-28
iOs signin works all 4 ways: FB, Twitter, Text, & Email.

SailingSteve:
There still are some minor visual glitches on the login sequence: the dialogs bounce around a little with data entry, and sometimes the send button has to be pressed twice.  But I did not want to wait, and get too far diverged from develop.
Maybe it's just the material IO dialog, but when the dialog uses a component that has an action that updates the store, it also updates all the stores for the underlying page. This is a problem because if the underlying page does a historyPush then you lose the dialog. Sometimes this works out fine by accident, but if there were some set up or take down that you needed to do in the dialog and it closes abruptly, the cleanup doesn't happen.  In case of Cordova, the bottom menu might not be restored when you come back, and other bad things can happen. There's also a problem where you can have top dialog doing a dispatch and the bottom dialog could be doing a dispatch at the same time during a store update -- then you get a "dispatch within a dispatch" error that often crashes the dialog.
Because you can't have two dispatches going at once, and it's not safe to set state to flag this condition, and you have to use the signInDialogGlobalState object to set a boolean, and then unfortunately go into all of the individual classes that would be responding to store updates and prevent them from udating state by using that boolean.
This problem is made many times worse by having actions that trigger other actions, since you have to block the cascading actions, that were intended for some other situation, and that have nothing to do with what you were doing in the dialog.  We really shouldn't have those cascading actions, they seemed easier at the time ,but they cause a lot of very difficult to debug troubles for others.

### DIFF
--- a/src/js/components/Widgets/SignInModal.jsx
+++ b/src/js/components/Widgets/SignInModal.jsx
@@ -11,13 +11,15 @@ import { withStyles, withTheme } from '@material-ui/core/styles';
 import { renderLog } from '../../utils/logging';
 import {
   isCordova,
-  isAndroidSizeSM, isAndroidSizeMD, isAndroidSizeLG, isAndroidSizeXL,
-  isIPhone3p5in, isIPhone4in, isIPhone4p7in, isIPhone5p5in, isIPhone5p8in, isIPhone6p1in, isIPhone6p5in,
   isWebAppHeight0to568, isWebAppHeight569to667, isWebAppHeight668to736, isWebAppHeight737to896,
-  isWebApp,
+  isWebApp, prepareForCordovaKeyboard, historyPush, restoreStylesAfterCordovaKeyboard,
 } from '../../utils/cordovaUtils';
 import SettingsAccount from '../Settings/SettingsAccount';
 import VoterStore from '../../stores/VoterStore';
+import { stringContains } from '../../utils/textFormat';
+import signInModalGlobalState from './signInModalGlobalState';
+
+/* global $ */
 
 class SignInModal extends Component {
   static propTypes = {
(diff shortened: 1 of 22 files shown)


src/js/components/Widgets/SignInModal.jsx
```diff
@@ -181,15 +215,7 @@ be honored, Cordova tries to do the best it can, but sometimes it crashes and lo
...
-  dialogRoot: isWebApp() ? {
-    height: '100%',
-    // position: 'absolute !important', // Causes problem on Firefox
-    top: '-15%',
-    left: '0% !important',
-    right: 'unset !important',
-    bottom: 'unset !important',
-    width: '100%',
-  } : {
+  dialogRoot: {
```
DaleMcGrew: I think we still need this "isWebApp" conditional to turn off the `position: 'absolute !important'` to get rid of an error I found in Firefox/WebApp Browser.

src/js/components/Widgets/SignInModal.jsx
```diff
@@ -207,26 +250,37 @@ const styles = theme => ({
...
+  //   maxHeight: '90%',
+  //   offsetHeight: 'unset !important',
+  //   top: '50%',
+  //   left: '50%',
+  //   right: 'unset !important',
+  //   bottom: 'unset !important',
+  //   position: 'absolute',
+  //   transform: 'translate(-50%, -58%)',
+  // },
+  // focusedOnSingleInput: isWebApp() ? {
```
DaleMcGrew: I believe we still need this first `focusedOnSingleInput` branch for WebApp. What do you think @SailingSteve?

src/js/components/Widgets/SignInModal.jsx
SailingSteve: I'll back out this change, ill need to make a full testing pass anyways for cordova.

src/js/components/Widgets/SignInModal.jsx
SailingSteve: same, will back out.

DaleMcGrew:
👍

PR #2783, opened 2020-01-29
iOS signin improvement in screen usage

SailingSteve:
On to critical friending input issues in cordova, then back to signon for more refining


DaleMcGrew:
👍

PR #2792, opened 2020-02-01
Android update, headers and hiding of footer menu for keyboard

SailingSteve:
New:  For Android, prepareForCordovaKeyboard, is invoked on input field focus/blur.  For iOS it is hooked in globally in Application.jsx with a 'keyboardWillShow' event listener.
Android APK: app-debug2020-02-01.apk


DaleMcGrew:
👍

Issue #2719, opened 2020-01-02
Your Endorsements: Edit and Preview Buttons in listing


DaleMcGrew:
On this page (in desktop): http://localhost:3000/settings/voterguidelist
![Screen Shot 2020-01-01 at 9 47 45 PM](https://user-images.githubusercontent.com/7756031/71653276-74143800-2ce0-11ea-852e-da2feb9279c2.png)

You will need to be signed in to work on this page.

Update the listing of each election where the text currently says "US 2020 Presidential Primary":
<img width="516" alt="Screen Shot 2020-01-01 at 9 54 38 PM" src="https://user-images.githubusercontent.com/7756031/71653455-54314400-2ce1-11ea-854c-25b1975a302f.png">

jarodpeachey:
@DaleMcGrew Does this need to show that card for each item? (One for the US 2020 Primary, one for the 2018 Midterms, etc;
)

DaleMcGrew:
Hi @jarodpeachey, yes a Card with the Edit and Preview buttons will be displayed for each election in the "Your Endorsements" list.

DaleMcGrew:
Verified!

Issue #2639, opened 2019-11-11
Positions to display: Desktop


DaleMcGrew:
Please upgrade this new auto-scroll system to work on desktop.

![Screen Shot 2019-11-02 at 7 06 43 AM](https://user-images.githubusercontent.com/7756031/68072164-9d940000-fd3f-11e9-83be-875fa8308cc8.png)

1. Sign in
2. Click your profile photo in the upper right corner
3. "Your Endorsements"
4. Choose an election

Please make it so that as you near the bottom of the page, that function is triggered to load more items in Desktop. When this feature is finished, the "Show More" link can be removed, but it would be good to leave the "Displaying 5 of 77" so there is an indication that there is more data in case the reload is delayed for any reason.

This is where the code is:
WebApp/src/js/components/Settings/VoterGuideSettingsAddPositions.jsx

There are two parallel functions to use:
increaseNumberOfBallotItemsToDisplay
increaseNumberOfPositionItemsToDisplay

DaleMcGrew:
This issue is still happening on desktop.

DaleMcGrew:
Great work @gautam-ergo! Verified.

PR #2793, opened 2020-02-03
Fix Voter Guide Auto-Scroll on Desktop and Redesign Voter Guide Links in Sidebar

jarodpeachey:
#2639 and #2719 

I added the functionality back in so the auto-load would work on desktop on the voter guide page. I also updated the design for the voter guide links in the sidebar to include two buttons.


DaleMcGrew:
Thank you @jarodpeachey!

PR #2796, opened 2020-02-12
Open share modal after signing in

jarodpeachey:

The sign-in modal now opens if you are not signed in and you click on the "We Vote Friends" option in the share modal. The share modal is then re-opened upon successfully signing in.


DaleMcGrew:
👍

Issue #2808, opened 2020-02-24
Complete Your Profile: Step 6 (Choose your first candidate)


DaleMcGrew:
NOTE: Please start with js/components/Ballot/FirstPositionIntroModal.jsx in which Dale made several changes needed for this issue (copied from js/components/Ballot/PersonalizedScoreIntroModal.jsx)

Make Step 6 "Choose your first candidate":
![Screen Shot 2020-02-24 at 9 12 50 AM](https://user-images.githubusercontent.com/7756031/75175044-7befc780-56e6-11ea-9685-e24661af8c7d.png)

In the modal, we want the following steps. Note that the goal is to get the voter to play with the Choose/Oppose + Add comment process and meet the following requirements:
1) Choose or Oppose once
2) Select Public once
3) Enter a comment and "Post" once
All we need them to do is do each requirement once. For example, if they choose the "Public" visibility option, and then move it back to "Friends Only" that is ok -- the "Privacy Setting" requirement will have been satisfied. The same is true if they Choose then "un"-Choose. They can also add a comment first, and then choose/oppose.

STEP 1:
Please use the existing "src/js/components/Ballot/CandidateItem.jsx" component with fake candidate data passed in. For now, feel free to hard-code to an existing candidate.
![Screen Shot 2020-02-24 at 9 22 42 AM](https://user-images.githubusercontent.com/7756031/75175466-3bdd1480-56e7-11ea-972e-3077a0ffa598.png)

STEP 2:
When the voter clicks on "Choose" or "Oppose", advance to this step:
![Screen Shot 2020-02-24 at 9 23 53 AM](https://user-images.githubusercontent.com/7756031/75175552-66c76880-56e7-11ea-9d00-5445e966813e.png)

STEP 3:
When the voter clicks "Public", advance to this step. Please note that all we need is for the voter to click to "Public" once. If the voter then goes back to "Friends Only", we can still consider this requirement satisfied:
![Screen Shot 2020-02-24 at 9 24 39 AM](https://user-images.githubusercontent.com/7756031/75175603-82327380-56e7-11ea-8dc6-c5aff2664cce.png)

STEP 4:
Once text has been entered, advance here:
![Screen Shot 2020-02-24 at 9 25 23 AM](https://user-images.githubusercontent.com/7756031/75175712-b86ff300-56e7-11ea-8acb-b56d5da06358.png)

DaleMcGrew:
[...]

PR #2828, opened 2020-03-06
Choose First Candidate Modal + Framework

jarodpeachey:
#2808 

This includes all the styling and props needed for the voter to choose their first candidate. I'll update this with the proper state so it's easier to update the API.


DaleMcGrew:
👍 Thank you @jarodpeachey.

PR #2835, opened 2020-03-10
Simplified ExtensionSignIn

SailingSteve:
If the device ID stored in extension differs from the device ID in the web app, and the web app is signed in, when you press "SIGN IN" in the extension, you navigate to the web app you will get the "you're signed in" dialogue, you press the button, the web app closes and you're back to the extension.
If the device IDs match, and the web app is not signed in you'll get the login dialogue, and after login you will be at the "you're signed in" signed  with the new button.


DaleMcGrew:
👍

PR #2853, opened 2020-03-18
added an extension.html file, so that you don't get a blank initial screen when loading https://wevote.us/extension.html

SailingSteve:


DaleMcGrew:
Hi @SailingSteve it may take me a couple of days to get this on https://WeVote.US (I need to finish some other code before I feel comfortable updating the live site), but I can get it on https://quality.WeVote.US today.

Issue #2841, opened 2020-03-14
Candidate: Right Flex Column not working when many Values/Issues


DaleMcGrew:
On the Candidate page, we recently added a right column (implemented with Flex) for the Sharing and Ballotpedia buttons. It looks like this:
![Screen Shot 2020-03-14 at 3 48 49 AM](https://user-images.githubusercontent.com/7756031/76680562-bf7f7800-65a6-11ea-8b3b-f041351e887f.png)

Unfortunately, if a candidate has a lot of Issues, and if we limit the issues to one vertical line, the Issues force the page to be very wide, causing the Candidate to break out of the top card:
![Screen Shot 2020-03-14 at 3 44 41 AM](https://user-images.githubusercontent.com/7756031/76680552-a971b780-65a6-11ea-843a-2ba061416c97.png)

The new columns code is in src/js/routes/Ballot/Candidate.jsx and looks like this:
```
        <div className="card">
          <TwoColumns>
            <LeftColumnWrapper>
              <CandidateItem
                ...
              />
            </LeftColumnWrapper>
            <RightColumnWrapper className="u-show-desktop-tablet">
              {candidate.ballotpedia_candidate_url && (
                <ViewOnBallotpedia externalLinkUrl={candidate.ballotpedia_candidate_url} />
              )}
            </RightColumnWrapper>
          </TwoColumns>
```

For now, I added the "expandIssuesByDefault" prop to the "CandidateItem" component which gets passed down to "IssuesByBallotItemDisplayList". When "expandIssuesByDefault" is set, and we show all of the Issues on multiple lines, the TwoColumns layout works.

I would like to be able to NOT use "expandIssuesByDefault", and have all of the issues "shrink" up to one line, without breaking the TwoColumns layout.

DaleMcGrew:
Thank you for solving this @lazorfuzz! 👍

PR #2877, opened 2020-04-04
[#2841] Delayed sequential render of issues chips

lazorfuzz:
#2841 
This will render the issue chips sequentially with a delay. This allows us to stop rendering when we know the next chip will overflow, or to backtrack (remove the last rendered chip) if there is an overflow.

This solution seems to work for all screen sizes, but is not responsive.


DaleMcGrew:
Thank you @lazorfuzz! 👍

PR #2881, opened 2020-04-05
Ballot test updates

QueMona98:
**I added a few more "id" to certain components so that the script can identify the corresponding elements.


DaleMcGrew:
👍

Issue #2870, opened 2020-03-31
Overlay panel needs to gray out entire background in desktop


DaleMcGrew:
Gray background should cover entire left side of screen in desktop
![Screen Shot 2020-03-31 at 11 23 37 AM](https://user-images.githubusercontent.com/7756031/78061688-723f2c80-7342-11ea-9ec8-9a9e6967da65.png)

DaleMcGrew:
Thank you Maria! 👍 I am moving the "x" part of this issue into a new Issue, so it can be picked up by another engineer. I look forward to working with you over summer!

PR #2890, opened 2020-04-08
Fixed styling to make gray background cover desktop. Fixed issue with overlay appearing when another button is clicked.

mdterrenal:
Part of #2870 
This fixes the issue where the gray background was not fully covering the desktop. I still need to work on adding the X button in the overlay panel to close the panel.


DaleMcGrew:
Thank you @mdterrenal! Very clean fix. 👍

PR #2891, opened 2020-04-08
Ballot test updates

QueMona98:
Changes include the following:
- Added some conditionals that are necessary to accommodate for a bug that is generated when clicking "Save" after writing a comment with iOS Browsers --> Once the bug is fixed, these conditionals can be removed
- I removed aspects of the test that require scrolling function
- I added a conditional that forces the scripts to skip over a test when testing on Samsung mobile because it requires a scrolling function--the element can't be clicked unless it is in view. This same element, however, doesn't require a scroll to be clicked on other platforms...
- I also noticed the OnePlus device behaved differently and I had to accommodate for this with a conditional as well--I need to click the "Save" because hitting "Enter" in the text box doesn't do that...it does for other platforms, though.

****Keep in mind that this script is just what I have so far and I have not updated it to test the new elements we just added to the WeVote webpage (i.e. the News sections, etc)

****Moreover, it appears that some elements aren't clicking anymore...So the script may not run as expected and I am still working on fixes. 

****This script should run on the iPhone 6 browser, the OnePlus device browser and the Firefox browser.


DaleMcGrew:
👍 Great work @QueMona98!

Issue #2864, opened 2020-03-28
USA Map: Compare two packages


DaleMcGrew:
We would like to improve the way voters can choose upcoming elections, based on state.

1) First package
I really like this package, but want to better understand how it scales responsively:
https://www.npmjs.com/package/react-usa-map

Please install `react-usa-map` on your local machine and get it rendering at the bottom of the list generated by:
src/js/components/Ballot/BallotElectionListWithFilters.jsx

Question: Does it offer the ability to turn on state names?

2) Second package
https://github.com/zcreativelabs/react-simple-maps
Create US only map with state names turned on.
Figure out way to change shading of each state as you hover.

DaleMcGrew:
Done for now. Great work Jarod!

PR #2892, opened 2020-04-08
Get both USA maps working

jarodpeachey:
#2864 
I included the two packages and got them both working. The first one does not display the names but shows the names in a title box. The second one shows the names, but you can't change the colors. Also, the hover state goes away when you hover the state's abbreviation.


DaleMcGrew:
Thank you @jarodpeachey. 👍

PR #2905, opened 2020-04-11
Issue2879

QueMona98:
Goal: Limit the number of issue icons displayed on a popover on a specific candidate/measure page

- I added a conditional that should limit the number of displayed issue icons --> it does so by determining the length of the container (array in this case) that the issues were encapsulated in. Logic: If that length is greater than 4, use only the first 3 entries (issues) of the container
- Moreover, I also used this logic to add in a '...' string whenever that conditions mentioned above are met


DaleMcGrew:
Hello @QueMona98, this approach will work, but the if/else duplicates a lot of code. I would recommend refactoring it -- we can discuss when we get on a meetup together next. 👍

PR #2906, opened 2020-04-13
USA Map in BallotElectionListWithFilters.jsx

jarodpeachey:
#2864 
I modified both packages to see which would work with the text and went with the second one, react-simple-maps. However, I grabbed the code from node_modules so we have the code in our repo.

I tried to get the text color to change on hover, but that wasn't going well, so I just made the background a darker shade normally, and full dark on hover. Let me know if you'd like me to continue trying to get the hover state to work.


DaleMcGrew:
👍 Thank you @jarodpeachey!

PR #2909, opened 2020-04-14
Issue2879

QueMona98:
I took your feedback into account and decided to use a variable to eliminate the code duplication that was shown in my previous conditional. While I still use the same logic to implement the solution to this issue, I implement that logic differently so that I don't duplicate so much code.


DaleMcGrew:
Yes! That is exactly right @QueMona98. Very good refactoring. 👍

PR #2910, opened 2020-04-15
Changes to CandidateItemEndorsement to allow creation of an new possible position, and deletion of an existing one

SailingSteve:
…

Committed without git hooks, multiple map issues.


DaleMcGrew:
👍

Issue #2896, opened 2020-04-08
Create valuesMainTest, based on tests/browserstack/specs/ballotMainTest.js


DaleMcGrew:
We need to test this entire section:
https://quality.wevote.us/values

DaleMcGrew:
Thank you @4D617278!

PR #2939, opened 2020-04-27
Values main test

4D617278:
#2896 Create valuesMainTest.js
created valuesMainTest.js with successful test for browser


4D617278:
Should I leave methods in the file or replace them with regular code?

tests/browserstack/specs/valuesMainTest.js
```diff
@@ -0,0 +1,210 @@
...
+    const followButton = 'button.issues-follow-btn__main.issues-follow-btn--blue';
+    const unfollow = 'button.dropdown-item.issues-follow-btn__menu-item:first-child';
+    const ignore = 'button.dropdown-item.issues-follow-btn__menu-item:nth-of-type(2)';
+    const unignore = 'span.d-print-none button';
+    const link = 'a.u-no-underline';
+    const mainContentColumn = 'div.col-sm-12.col-md-8';
+    const mainContainer = 'div.opinions-followed__container';
+    const childCard = 'div.card-child.card-child--not-followed';
+    const clearSearchIconSelector = 'img[src="/img/glyphicons-halflings-88-remove-circle.svg"]';
+    const sqlInjectionTest = '\' or 1=1 -- -';
```
DaleMcGrew, line 80: Hi @4D617278 I really like that you are setting variables at the start of the testing script. Since we often rename styles, it will be more maintainable if you use id's. Feel free to add ids to the code as needed. For example instead of `const mainContentColumn = 'div.col-sm-12.col-md-8';` you can add `id='valuesMainContentColumn'` to that div and then reference that.

tests/browserstack/specs/valuesMainTest.js
```diff
@@ -0,0 +1,210 @@
...
+      await browser.url('https://quality.wevote.us/values');
+      await simpleClick('undefined-splitIconButton'); // Clicks on "Add Endorsements"
+      await browser.url('https://quality.wevote.us/values');
+    }
+
+    // //////////////////////
+    // Tests organizations to follow
+    await simpleClick('organizationsToFollowPreviewShowMoreId'); // Clicks on "Explore more organizations"
+    await simpleClick('backToLinkTabHeader'); // Clicks on "Back"
+    publicFigureHeader = await $('h2=Public Figures to Follow');
```
DaleMcGrew, line 180: This is another example of where to use an id, since text like "Public Figures to Follow" may get modified without notice, and then break the test.

PR #2948, opened 2020-05-03
Issue2903

QueMona98:
I did the following:
1) Added in TwoColumns component so that the right column could include the "Share" and "google Search" buttons

2)I made adjustments to the styling code in the MeasureItem component because it was interfering with the seamless implementation of the TwoColumns component

Concerns:
1) I couldn't figure out how to get the value of the ballotpedia_measure_url--I tried using the ballotpedia button from the candidate page (specifically the ballotpedia_candidate_url value) to guide my addition of the measure page equivalent, but to no avail.

2) Moreover, looks like a component called "styles" at the end of the MeasureItem component is no longer required. I wasn't sure if I should refactor and essentially omit it to clear up clutter since it appears that it is no longer being used; I'm not sure, however, so I was wondering what your thoughts on this are.


DaleMcGrew:
Thank you @QueMona98! 👍

PR #2974, opened 2020-05-20
 Split index.js into two, so that the second half (startReactApp) can be loaded after the Cordova 'deviceready' event fires

SailingSteve:
The split leaves non-react startup in index.js, and the react part in startReactApp.
Some early work for "Sign In With Apple" is in Application.jsx, but commented out.
Minor lint cleanup.


DaleMcGrew:
👍

Issue #2944, opened 2020-04-30
Why does countdown in desktop sometimes remove "0" for minutes or seconds


DaleMcGrew:
When on the "Ready?" page in desktop mode, I have noticed that sometimes it will show:
"150 : 11 : 6 : 1"
...instead of the intended:
"150 : 11 : 06 : 01"

Please track down why this happens and fix.

DaleMcGrew:
Verified and live. Thank you @csummer! Please let us know if you want to tackle any other issues. ;)

PR #2977, opened 2020-05-22
[Issue #2944] removed else-if logic for zero-padding time

csummer:

Issue #2944 

All three (hours, minutes, seconds) need to test for a zero prefix. Also removed the `array.includes()` test in favor of the numerical test `n < 10`


DaleMcGrew:
Good fix @csummer!  👍 Thank you.

Issue #2942, opened 2020-04-29
Add Preview Button on Your Endorsements page


DaleMcGrew:
On this "Your Endorsements" page, put the "See Preview >" button at the top of the page in the same header row with "Endorsed or Opposed" and "Add Endorsements".
![Screen Shot 2020-04-29 at 8 13 41 AM](https://user-images.githubusercontent.com/7756031/80613214-d0b70380-89f1-11ea-856b-8d1340bba486.png)

The button to copy currently lives at the very bottom of the page:
![Screen Shot 2020-04-29 at 8 17 16 AM](https://user-images.githubusercontent.com/7756031/80613263-e3313d00-89f1-11ea-9ce1-0492f6d58941.png)

1. Sign in
2. Click on your profile icon, and choose "Your Endorsements"
3. Choose any election

In mobile the text will probably need to be limited to "Preview"

DaleMcGrew:
Verified and live. Thank you @QueMona98!

PR #2979, opened 2020-05-26
Issue2942

QueMona98:
This pull request is centered on Issue #2942 

The "See Preview" button was moved from the button to the page to the top, specifically in the header.
*Please review the placement of the button and ensure that it is placed sufficiently--I tried making as few changes to the code as possible*

Moreover, I made some changes to the code of the Header so that the button link still works. I would love feedback if you stumble upon any opportunities for improvement!

Thanks


DaleMcGrew:
👍

PR #2984, opened 2020-06-01
Intermediate checkin for Sign in with Apple, under Cordova on iOS

SailingSteve:
WebApp developers and users should see no difference


DaleMcGrew:
👍

PR #2998, opened 2020-06-07
Second intermediate WebApp update preparing for an iOS Cordova Release

SailingSteve:
Pretty much everything that is needed for an iOS release except Twitter signin is currently broken
and need a new iOS storyboard for the loading screen.
Fixed the fifteen issues in wevote/WebApp/issues/2978 that are specifically in the WebApp.
Fixed the console warnings for componentWillReceiveProps and componentWillMount in files that I
otherwise modified, so that they can be compiled under React 17, hopefully this will also slow the copying
of these functions into new classes.


DaleMcGrew:
👍  Thank you @SailingSteve!

PR #3026, opened 2020-06-19
Cordova release candidate for iOS

SailingSteve:
"Sign In With Apple" and "Sign in With Twitter" are not available in Cordova on the 5% of iOS devices that are running a version lower than 13.0
Offset changes everywhere
Restored WelcomeAppbar on the About page
Renamed some Cordova offset enums, to match URL changes that have been made since they were defined.


DaleMcGrew:
👍

PR #3065, opened 2020-07-05
Issue2935

QueMona98:
I used a preexisting "Show More" component/button to implement the solution to this issue. Let me know if you like the look of it or would prefer I change the "Show More" link. I just tested it out and the link should also be working properly :)


DaleMcGrew:
Thank you @QueMona98! 👍

Issue #3018, opened 2020-06-15
voterPlansForVoterRetrieve loads 16x per second when entering your voter plan address


SailingSteve:
```
2020-06-15 16:30:32.184622-0700 WeVoteCordova[97812:3886048] AJAX URL: https://api.wevoteusa.org/apis/v1/voterPlansForVoterRetrieve/
2020-06-15 16:30:32.185328-0700 WeVoteCordova[97812:3886048] render ==== VoterPlanModal ====
2020-06-15 16:30:32.189048-0700 WeVoteCordova[97812:3886048] render ==== OpenExternalWebSite ====
2020-06-15 16:30:32.243134-0700 WeVoteCordova[97812:3886048] AJAX Response to endpoint: voterPlansForVoterRetrieve
2020-06-15 16:30:32.245523-0700 WeVoteCordova[97812:3886048] AJAX URL: https://api.wevoteusa.org/apis/v1/voterPlansForVoterRetrieve/
2020-06-15 16:30:32.246371-0700 WeVoteCordova[97812:3886048] render ==== VoterPlanModal ====
2020-06-15 16:30:32.250576-0700 WeVoteCordova[97812:3886048] render ==== OpenExternalWebSite ====
2020-06-15 16:30:32.304581-0700 WeVoteCordova[97812:3886048] AJAX Response to endpoint: voterPlansForVoterRetrieve
2020-06-15 16:30:32.306798-0700 WeVoteCordova[97812:3886048] AJAX URL: https://api.wevoteusa.org/apis/v1/voterPlansForVoterRetrieve/
2020-06-15 16:30:32.307477-0700 WeVoteCordova[97812:3886048] render ==== VoterPlanModal ====
2020-06-15 16:30:32.311080-0700 WeVoteCordova[97812:3886048] render ==== OpenExternalWebSite ====
2020-06-15 16:30:32.362175-0700 WeVoteCordova[97812:3886048] AJAX Response to endpoint: voterPlansForVoterRetrieve
2020-06-15 16:30:32.364418-0700 WeVoteCordova[97812:3886048] AJAX URL: https://api.wevoteusa.org/apis/v1/voterPlansForVoterRetrieve/
2020-06-15 16:30:32.365125-0700 WeVoteCordova[97812:3886048] render ==== VoterPlanModal ====
2020-06-15 16:30:32.368995-0700 WeVoteCordova[97812:3886048] render ==== OpenExternalWebSite ====
[...]

PR #3079, opened 2020-07-14
fix for voter plan rerender

mlevoy:
https://github.com/wevote/WebApp/issues/3018

An extra check in the ready store when retrieving voter plans. If the voter plan list is empty, return state and don't update anything to avoid the infinite re-render.
Also, after merging in latest develop to my branch I had to run npm install for things to compile.  This caused the updates in package-lock.json

### DIFF
--- a/src/js/components/Ready/VoterPlanModal.jsx
+++ b/src/js/components/Ready/VoterPlanModal.jsx
@@ -169,7 +169,7 @@ class VoterPlanModal extends Component {
     let savedVoterPlanFound = false;
     if (!voterPlansForVoterRetrieved) {
       ReadyActions.voterPlansForVoterRetrieve();
-    } else {
+    } else if (savedVoterPlan.google_civic_election_id) {
       savedVoterPlanFound = true;
     }
     this.setState({
--- a/src/js/components/Ready/VoterPlanModal.jsx
+++ b/src/js/components/Ready/VoterPlanModal.jsx
@@ -165,8 +165,9 @@ class VoterPlanModal extends Component {
     const googleCivicElectionId = VoterStore.electionId();
     const savedVoterPlan = ReadyStore.getVoterPlanForVoterByElectionId(googleCivicElectionId);
     this.setVoterPlanSavedStates(savedVoterPlan);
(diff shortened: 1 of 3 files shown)


src/js/components/Ready/VoterPlanModal.jsx
```diff
@@ -165,8 +165,9 @@ class VoterPlanModal extends Component {
     const googleCivicElectionId = VoterStore.electionId();
     const savedVoterPlan = ReadyStore.getVoterPlanForVoterByElectionId(googleCivicElectionId);
     this.setVoterPlanSavedStates(savedVoterPlan);
+    const voterPlansForVoterRetrieved = ReadyStore.getVoterPlansForVoterRetrieved();
     let savedVoterPlanFound = false;
-    if (savedVoterPlan.google_civic_election_id === undefined) {
+    if (!voterPlansForVoterRetrieved) {
       ReadyActions.voterPlansForVoterRetrieve();
     } else {
       savedVoterPlanFound = true;
```
DaleMcGrew, line 173: Hi @mlevoy I think we still wan an `else if (savedVoterPlan.google_civic_election_id)` test before marking "savedVoterPlanFound = true"

src/js/stores/ReadyStore.js
```diff
@@ -70,15 +77,25 @@ class ReadyStore extends ReduceStore {
...
+          voterPlanList.forEach((oneVoterPlan) => {
+            googleCivicElectionId = oneVoterPlan.google_civic_election_id || 0;
+            allCachedVoterPlansForVoterByElectionId[googleCivicElectionId] = oneVoterPlan;
+          });
+          return {
+            ...state,
+            allCachedVoterPlansForVoterByElectionId,
+            voterPlansForVoterRetrieved,
+          };
+        } else {
```
DaleMcGrew, line 93: As extra protection against infinite loop, maybe make this `else if (voterPlansForVoterRetrieved)` for setting "voterPlansForVoterRetrieved" to true since we only need to set it true once. Then finally `return state;` for the final `else`?

DaleMcGrew:
Thank you @mlevoy! 👍

Issue #2691, opened 2019-12-18
Sign out error: "We were not able to find an account for this Twitter handle."


DaleMcGrew:
### Please describe the issue (What happens? What do you expect?)
Signing out on page with Twitter handle in the route shows you the error "We were not able to find an account for this Twitter handle." when there is an account.

### Steps to reproduce the problem (1, 2, 3...), including links
1. Make sure you are signed in
2. Start on a Twitter page for an organization, for example:
https://localhost:3000/sonyabea/ballot
3. Sign out
4. You stay on the same URL, but are now shown this error instead of the content that was shown before signing out:
![Screen Shot 2019-12-18 at 12 04 10 PM](https://user-images.githubusercontent.com/7756031/71119212-8776bd80-218e-11ea-8423-cda6bf8e7a21.png)

Second part:
1. Make sure you are signed out
2. Start on a Twitter page for an organization, for example:
https://localhost:3000/sonyabea/ballot
3. Sign in
4. When you click to close the "You are signed in" modal, the browser is redirected to the /ballot page, and it should remain on the organization's page.

DaleMcGrew:
Fixed!

PR #3080, opened 2020-07-15
modified TwitterStore to handle state reset on logout

gautam-ergo:
https://github.com/wevote/WebApp/issues/2691

twitter sign out error, modified TwitterStore to handle state reset on logout


DaleMcGrew:
👍 Great find @gautam-ergo!

PR #3091, opened 2020-07-22
Changes to allow us to build a single bundle webapp that runs in Cordova

SailingSteve:


DaleMcGrew:
I pulled this locally, removed package.json change, rolled back package-lock.json, and then merged with develop. Thank you @SailingSteve!

Issue #3086, opened 2020-07-18
Filter text from Ballotpedia with stripHtmlFromString


DaleMcGrew:
### Please describe the issue (What happens? What do you expect?)
Ballotpedia sometimes provides candidate descriptive text that has `&nbsp;` in it. Please filter any places that displays paragraph text from Ballotpedia (no need for just names) with this function: 
src/js/utils/textFormat.js stripHtmlFromString

By searching for `stripHtmlFromString` in the latest code in the develop branch, you will find examples of this function in use.

### Steps to reproduce the problem (1, 2, 3...), including links
1. Visit https://wevote.us/candidate/wv02cand65836/b/btdb
2. You will see a `&nbsp;` in the description of Lorena Gonzalez
3. Please trace out the code that places this text on the screen, and then find other places throughout the code that might have this problem too.

DaleMcGrew:
Verified, thank you @mlevoy!

PR #3095, opened 2020-07-22
strip candidate detail text of html

mlevoy:
https://github.com/wevote/WebApp/issues/3086
Added stripHtmlFromString to candidate detail text coming back from ballotpedia.

### DIFF
--- a/src/js/components/Ballot/CandidateItem.jsx
+++ b/src/js/components/Ballot/CandidateItem.jsx
@@ -20,7 +20,11 @@ import ShowMoreFooter from '../Navigation/ShowMoreFooter';
 import SupportStore from '../../stores/SupportStore';
 import TopCommentByBallotItem from '../Widgets/TopCommentByBallotItem';
 import VoterGuideStore from '../../stores/VoterGuideStore';
-import { abbreviateNumber, numberWithCommas } from '../../utils/textFormat';
+import {
+  abbreviateNumber,
+  numberWithCommas,
+  stripHtmlFromString,
+} from '../../utils/textFormat';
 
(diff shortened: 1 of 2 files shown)


mlevoy:
The only location I saw that we were using candidate summary text from ballotpedia was in the CandidateItem.

src/js/components/Ballot/CandidateItem.jsx
```diff
@@ -108,6 +112,7 @@ class CandidateItem extends Component {
       const twitterDescriptionText = twitterDescription && twitterDescription.length ? `${twitterDescription} ` : '';
       const ballotpediaCandidateSummary = candidate.ballotpedia_candidate_summary;
       let ballotpediaCandidateSummaryText = ballotpediaCandidateSummary && ballotpediaCandidateSummary.length ? ballotpediaCandidateSummary : '';
+      ballotpediaCandidateSummaryText = stripHtmlFromString(ballotpediaCandidateSummaryText);
       ballotpediaCandidateSummaryText = ballotpediaCandidateSummaryText.split(/<[^<>]*>/).join(''); // Strip away any HTML tags
```
mlevoy, line 116: happy to remove the split or leave it in as an extra precaution

DaleMcGrew: @mlevoy any modification to remove the HTML can be put in the `stripHtmlFromString` function. Could you more your `split` code into that function, and search for any other uses of `split` like this so we can consolidate it all in that function?

DaleMcGrew:
Looks great @mlevoy! 👍

PR #3112, opened 2020-08-03
Refactor to remove RootSingleBundle, which was for Cordova single bundle builds

SailingSteve:
Use Root.jsx for Cordova since Suspense and LazyLoad don't matter if the code is already there in a single bundle.
Tested with Cordova and WebApp.


DaleMcGrew:
👍

PR #3113, opened 2020-08-04
changing max-width style to include firefox

mlevoy:
squished avatar on firefox
avatar width fit-content switched to max content, which is compatible with firefox


DaleMcGrew:
👍

PR #3114, opened 2020-08-04
Avatar fix2

mlevoy:
fix for avatar not showing up on certain pages without caching
gave avatar a min and max width based on the min and max dimensions see in the app


DaleMcGrew:
👍

PR #3115, opened 2020-08-04
Fixed sign in with twitter for Cordova, and css for voter photo on sign in for iPhones with notches

SailingSteve:
Made the loading of `appleid.auth.js` inline in index.html, instead of as a delayed load


DaleMcGrew:
👍

PR #3133, opened 2020-08-12
Header offset changes to match new dev on Notifications page for Cordova Android

SailingSteve:
Cleared some lint warnings


DaleMcGrew:
👍

PR #3136, opened 2020-08-13
Cordova: Replace button text with "Requires iOS 13" for Apple & Twitter if appropriate

SailingSteve:
Header offset changes for Cordova Android with front facing lens notch


DaleMcGrew:
👍

Issue #3142, opened 2020-08-17
Prior % in ReadyTaskBallot still appears after sign out


DaleMcGrew:
1. Go to the Ready page when you are signed in. Make sure you have a "completion %" on the Ballot To-Do list (by choosing a candidate on your ballot for example).
2. Sign out from the Ready page (using the drop down profile menu)
3. The "completion %" remains after you sign out 

src/js/components/Ready/ReadyTaskBallot.jsx
Search for the "calculateScore" function and "getBallotItemsStatusCounts" in the BallotStore.

DaleMcGrew:
Verified!

PR #3146, opened 2020-08-18
adding check for 0 in addition to truthy for updating state of percen…

mlevoy:
…t completed

https://github.com/wevote/WebApp/issues/3142
Additional check for 0 in the case the percentCompleted equals 0.


DaleMcGrew:
👍

Issue #3152, opened 2020-08-22
PositionStatementModal vertical spacing problem in Safari


DaleMcGrew:
1. Go to any ballot.
2. Click on a candidate or measure.
3. Click on the Add Explanation button.

The vertical spacing works in Chrome, but not in Safari. Please fix the spacing in Safari to match how it is in Chrome.

CHROME:
![Screen Shot 2020-08-21 at 5 29 53 PM](https://user-images.githubusercontent.com/7756031/90944508-1b5ef000-e3d4-11ea-8a50-ecba87e3fd02.png)

SAFARI:
<img width="584" alt="Screen Shot 2020-08-21 at 5 29 14 PM" src="https://user-images.githubusercontent.com/7756031/90944529-2d409300-e3d4-11ea-94e3-8fb43d372267.png">

This is file: src/js/components/Widgets/PositionStatementModal.jsx
...

DaleMcGrew:
Resolved.

PR #3153, opened 2020-08-22
added height to outer div

mlevoy:
https://github.com/wevote/WebApp/issues/3152
added height to dialogcontent div


DaleMcGrew:
👍

PR #3169, opened 2020-08-28
Fixes to stop the footer menu from bouncing on first render in Ballot and News in Cordova

SailingSteve:
Added a cordovaOffset for the top of the News page, fix is in ActivityPostAdd (I think where the issue was introduced)


DaleMcGrew:
👍

Issue #3082, opened 2020-07-16
Complete address not saved when you use Google Maps dropdown


DaleMcGrew:
### Please describe the issue (What happens? What do you expect?)
If you start to type your address, the Google autocomplete widget will suggest addresses in a drop down.
![Screen Shot 2020-07-16 at 3 24 13 PM](https://user-images.githubusercontent.com/7756031/87729224-b0d5e700-c779-11ea-9a14-cd810d9e84d7.png)

If before you finish typing the full address you use the down arrow to select an address from the Google autocomplete, and then hit enter, the full address does not get saved. The partial one does.
![Screen Shot 2020-07-16 at 3 24 20 PM](https://user-images.githubusercontent.com/7756031/87729241-b8958b80-c779-11ea-8853-7d1838ff91a5.png)

### Steps to reproduce the problem (1, 2, 3...), including links
1. In desktop mode, go to Ballot page and click on the gear icon next to the election name. A modal will open.
2. Click on the gear icon next to the address that is currently saved to open the edit form.
3. Clear the address and Save.
4. Close the "Address & Elections" modal.
5. Go back to the "Ready" page and you will see an input box for "Full address and ZIP"
6. Start to type an address like "123 Main Street". When the Google Autocomplete starts to offer you an address, use the down arrow to select one, and then hit the "Enter" key on your keyboard.
It is the partial address you have typed that gets saved, instead of the full address suggested by Google auto comlete.

I found the core of this problem is in: src/js/components/Ready/EditAddressOneHorizontalRow.jsx
Related functions: _placeChanged, handleKeyPress, voterAddressSave, saveAddressToApiServer, onBallotStoreChange and these inputProps in "InputBase":
inputProps={{
  onBlur: this.saveAddressToApiServer,
  onChange: this.updateVoterAddress,
  onKeyDown: this.handleKeyPress,
}}

utsab:
@DaleMcGrew Are you cool with me taking on this task?

DaleMcGrew:
Yes that's fine @utsab !

DaleMcGrew:
Verified!

PR #3185, opened 2020-09-05
Fix incomplete address

utsab:
#3082  

Previously, the saveAddressToAPIServer function was called after the onBlur event.  The onBlur event is triggered whenever focus shifts away from the address input.  Unfortunately,. when the user clicks on the auto-completed address (in the dropdown), this also triggers the onBlur event.  At that moment whatever partial address the user manually typed into the input box gets saved to the server.   The events happen as follows, for example: 

1. User types "5000 Forbes Ave, Pitt" 
2. Autocompleted results show "5000 Forbes Ave, Pittsburgh, PA"
3. User clicks on the autocompleted result
4. The onBlur event gets triggered
5. The saveAddressToAPIServer gets triggered
6.  the _placeChanged function gets called (which then saves "5000 Forbes Ave, Pittsburgh, PA" to the state) 

Note that step 5 happens **before** the autocompleted address is saved into the state. 

The fix will trigger a call to saveAddressToAPIServer **after** the user's selected auto-completed address (from the dropdown) is updated into the react state.


DaleMcGrew:
Thank you @utsab for doing this analysis! I'll check it out ASAP.

DaleMcGrew:
Thank you @utsab for doing this investigation. I added some fixes here: https://github.com/wevote/WebApp/pull/3192

PR #3187, opened 2020-09-07
Basic functionality for notifications to iOS app under Cordova is working

SailingSteve:
Uses Firebase Cloud Messaging (FCM) to communicate to apps, and we will
develop a link from our Python servers to send the FCM devices specific
"token" to the Google Firebase cloud.  The token is now gathered in startCordova.js
and sent to the API server via deviceStoreFirebaseCloudMessagingToken
where it is stored in the voter_device_link table.


DaleMcGrew:
👍

PR #3196, opened 2020-09-12
Work around for opening new "tabs" on Android

SailingSteve:
Most of this work is in the WeVoteCordova project.
We now use a fork of EddyVerbruggenEddyVerbruggen/cordova-plugin-safariviewcontroller to open external sites in a "tab" on iOS and InAppBrowser on Android, to work around what appears to be a cordoava-android V9 Java dependency conflict issue.


DaleMcGrew:
👍

PR #3200, opened 2020-09-13
Notification count on bell now is shown on icon after iOS app loads

SailingSteve:
Maintenance on loading of pngs and gifs without cordovaDot()
Set a Cordova only topMargin on the drawers that open on Ballot and Discuss/News


DaleMcGrew:
Awesome @SailingSteve, thank you for doing that! 👍

PR #3217, opened 2020-09-21
Made the Confirm/Save button smaller on the Address Entry component

SailingSteve:
Added a fix for Notifications in Android.


DaleMcGrew:
👍

PR #3222, opened 2020-09-25
Minor fixes for Cordova

SailingSteve:
Made the tiny icon (that duplicates the candidate icon on the "+1 to score" button) square.
Worked around the hardcoded path in _card.scss for the compressed icon, when in Cordova.
Fixed the top margin, when on a candidate via twitter handle page, for iPads.


DaleMcGrew:
👍

PR #3226, opened 2020-10-03
Renamed componentWillReceiveProps to UNSAFE_componentWillReceiveProps globally

SailingSteve:
Also renamed componentWillMount to UNSAFE_componentWillMount
This suppresses pages of JavaScript console logging.
All of these UNSAFE deprecated React methods will need to be removed over time.
Suppressed two lint errors introduced by the renaming, since the method name is
meant to be ugly (by the Facebook React authors), and the renamed methods will
be going away.


DaleMcGrew:
👍

PR #3242, opened 2020-10-09
Cordova changes for next iOS release

SailingSteve:
Lots of header adjustments
Pane size adjustments for StickyHeaders on Candidate and Measures
3rd (and best) way to determine if iOS App is running on the desktop for Apple Silicon (ARM64 processor) -- the variable is now named isIOSAppOnMac, which matches the native code variable name.
Refactored cordovaScrollablePaneTopPadding into its own file
Report the app version in the Device Dialog
Suppressed the "Preview Link in New Window" button in Cordova -- it was opening a second instance of the app, and it was not obvious that you had to close the dialog, and then close the "tab" to return to your context.  And there were multiple visual issues.  Plus the second tab seems completely unrelated to the purpose of the link preview.


DaleMcGrew:
👍

PR #3246, opened 2020-10-22
Cordova changes for next Android release (2.1.0)

SailingSteve:
Lots of header adjustments
For Facebook login, no longer request user_friends permission, since we have not received it from Facebook, and it was
causing warnings.  Same for Facebook login with WebApp.


DaleMcGrew:
👍

PR #3247, opened 2020-10-29
Microsoft Internet Explorer fixes

SailingSteve:
Added String.includes polyfill for IE 11, which is pre-ECMAScript 2015
Suppress Sign In With Apple, if running IE

For wevote/WebApp/issues/3184


DaleMcGrew:
👍

PR #3248, opened 2020-10-29
Microsoft Internet Explorer fixes, Part 2

SailingSteve:
Install polyfillFixes in Root.jsx before first use

For wevote/WebApp/issues/3184


DaleMcGrew:
👍

PR #3249, opened 2020-10-30
Microsoft Internet Explorer fixes, Part 3

SailingSteve:
Install polyfillFixes in index.jsx before almost anything else

For wevote/WebApp/issues/3184


DaleMcGrew:
👍

PR #3250, opened 2020-10-30
Microsoft Internet Explorer fixes, Part 4

SailingSteve:
Call polyfillFixes() from a few places, but have it set a global within the file, so that it only runs once.  Then we should be hardened against variations in the way webpac compiles the bundles.
Lint cleanup.
Moved polyfillFixes() to cordovaUtils.js to avoid a circular dependency.

For wevote/WebApp/issues/3184


DaleMcGrew:
👍

PR #3251, opened 2020-10-31
Microsoft Internet Explorer fixes, Part 5

SailingSteve:
Supporting IE is not worth it.  IE does not support arrow functions, and we have over 100 of them.
Just put up a "We're sorry, we no longer support Internet Explorer... message for IE users.

For wevote/WebApp/issues/3184

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -53,7 +53,7 @@
           top: 0;
           width: 100vw;
           ">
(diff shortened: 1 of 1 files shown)


SailingSteve:
@DaleMcGrew I need to test this PR on Browser Stack

DaleMcGrew:
👍

PR #3252, opened 2020-11-03
Changes for first Apple Silicon release

SailingSteve:
Made the two columns on Ballot (in roughly iPad aspect on Apple Silicon), side-by-side.  Same for News/Discussion for Apple Silicon and iPads.
Cleaned out experimental code.
Made two util functions crash proof, in case their Cordova package dependencies did not load correctly.
No need to deal with the virtual keyboard if Apple Silicon, since there isn't one.
A complete test round, and margin adjustments for iOS.
Fixed horizontal scrolling in four places:  IssueCard, CandidateForDrawer (card), Candidate (card), and Measure (card).
Added a ReadyInformationDisclaimer tab for Cordova, specifically for Googles concerns about our Android app.

### DIFF
--- a/src/js/actions/FacebookActions.js
+++ b/src/js/actions/FacebookActions.js
@@ -58,10 +58,11 @@ export default {
       console.log('FacebookActions.getFacebookData was not invoked, see ENABLE_FACEBOOK in config.js');
       return;
     }
-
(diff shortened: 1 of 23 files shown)


DaleMcGrew:
👍

PR #3253, opened 2020-12-06
Ready page data source information disclosure for Android

SailingSteve:
Moved the Data disclosure pane to the bottom for iOS, and left it higher for Android approval.
(We did get approved by Google with this change in place.)


SailingSteve:
combined with tonights changes in new PR

PR #3257, opened 2020-12-28
Updated the the latest npm packages. pushHistory is only partly upgraded.

SailingSteve:
Updated to the react-router-dom build of react-router V5 (we were at v3).
The latest react-router passes params in a "match" prop, previously as a root prop.
Lint cleanup.
Consolidated to a function, weVoteBrandingOff() from various (mostly) copied implementations, eliminated the passing of the result of weVoteBrandingOff.
Eliminated most passing of pathname, location, and voter - passing copies of these globals is not needed -- globals like window.location, and global stores like VoterStore.  Caused problems since they were stale in some cases, and had been added to state in other cases.
The latest ReactSVG has a different function signature.
Removed ReactLazy load, Dale said it didn't improve performance, and it was causing trouble after the react-router upgrade.
Only the "start-https" run config, has been tested.


DaleMcGrew:
👍 This upgrade project is still in process, and not meant for either the quality or live sites.

PR #3258, opened 2020-12-30
I would love your thoughts on this pull request. Initial updates to make sure I have this right. It seems like we might want to pass "history" into Application in the Root.jsx file, what do you think @SailingSteve? Also the linter is warning about prop spreading and the way around that is to declare props in the new format shown for example on lines 78-80 on OrganizationsToFollowPreview.jsx

DaleMcGrew:
I would love your thoughts on this pull request. Initial updates to make sure I have this right. It seems like we might want to pass "history" into Application in the Root.jsx file, what do you think @SailingSteve? Also the linter is warning about prop spreading and the way around that is to declare props in the new format shown for example on lines 78-80 on OrganizationsToFollowPreview.jsx

### DIFF
--- a/src/js/Application.jsx
+++ b/src/js/Application.jsx
@@ -6,7 +6,7 @@ import AppActions from './actions/AppActions';
 import AppStore from './stores/AppStore';
 import { getApplicationViewBooleans, setZenDeskHelpVisibility } from './utils/applicationUtils';
 import cookies from './utils/cookies';
-import { getToastClass, historyPush, isIOSAppOnMac, isCordova, isWebApp } from './utils/cordovaUtils';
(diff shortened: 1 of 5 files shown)


SailingSteve (commented):
f that change in Root.jsx works it would be great.  HTML5 creates a ‘history’ object that is much simpler and insufficient for our task.  react-router-dom creates a history object that is a superset of the HTML5 built in history.   You can tell in the debugger if you are using the wrong ‘history’ if when opening the object, it does not have a history.push() function (which is what we need and it comes from react-router-dom, is not global, and comes in as a prop.  Having said that, if you found a better way, that really would be great.
If we use the wrong history object, I would expect historyPushV5 to pass the first if statement and then throw and exception of some sort on the history.push().

Otherwise the rest of the change looks perfect!

PR #3260, opened 2020-12-30
Now capture the react-router-dom history object in global.weVoteGlobalHistory, no need to pass history everywhere we need to do a pushHistory

SailingSteve:
Removed all passing of history
Many links still  need to be tested
Some routes in Root still need improved regexes for V5 react-router
The Application.jsx updating the header on path changes is not yet reliably working.

### DIFF
--- a/src/js/components/Navigation/HeaderBar.jsx
+++ b/src/js/components/Navigation/HeaderBar.jsx
@@ -4,7 +4,7 @@ import styled from 'styled-components';
 import { Button, AppBar, Toolbar, Tabs, IconButton, Tooltip } from '@material-ui/core';
 import { Place, AccountCircle } from '@material-ui/icons';
 import { withStyles } from '@material-ui/core/styles';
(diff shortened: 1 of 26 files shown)


src/js/components/Navigation/HeaderBar.jsx
```diff
@@ -344,10 +344,9 @@ class HeaderBar extends Component {
     AppActions.setShareModalStep('');
     const { location: { href } } = window;
     if (stringContains('/modal/share', href) && isWebApp()) {
-      const { history } = this.props;
       const pathnameWithoutModalShare = href.replace('/modal/share', '');  // Cordova
       // console.log('Navigation closeShareModal ', pathnameWithoutModalShare)
-      historyPushV5.push(history, pathnameWithoutModalShare);
+      historyPush.push(null, pathnameWithoutModalShare);
```
DaleMcGrew, line 349: Should the `null` be here?

SailingSteve: good catch!

DaleMcGrew:
👍

PR #3261, opened 2020-12-31
Update to react-router V5, "TwitterHandle only" URLs work, don't render the Footer if it will not be displayed

SailingSteve:
Headers now update on window.location change
Removed some temporary logging that is no longer needed.


DaleMcGrew:
👍

PR #3264, opened 2021-01-04
Update to react-router-5, correct layout on urls that start with twitter handles

SailingSteve:
(Routing and the booleans set for routes, would be so much simpler if twitter handles had some sort of reliable path component before the handle, like wevote.us/twitter/sciam)

### DIFF
--- a/src/js/Application.jsx
+++ b/src/js/Application.jsx
@@ -34,6 +34,17 @@ class Application extends Component {
       // Do not define voter here. We rely on it being undefined
       voter_initial_retrieve_needed: true,
     };
+    // 2021-1-3: This is a workaround for the difficulty of nesting components in react-router V5, it should not be necessary
+    global.weVoteGlobalHistory.listen((location, action) => {
+      if (location.pathname !== this.state.priorPath) {
+        // Re-render the Application if the path changed (Needed for React-router V5)
+        this.setState({ priorPath: window.locationPathname });
+      }
+      if (webAppConfig.LOG_ROUTING) {
+        console.log(`Application: The current URL is ${location.pathname}${location.search}${location.hash}`);
+        console.log(`Application: The last navigation action was ${action}`, JSON.stringify(global.weVoteGlobalHistory, null, 2));
+      }
+    });
   }
 
   componentDidMount () {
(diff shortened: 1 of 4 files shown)


src/js/routes/VoterGuide/OrganizationVoterGuideEdit.jsx
```diff
@@ -101,7 +101,7 @@ export default class OrganizationVoterGuideEdit extends Component {
     const voter = VoterStore.getVoter();
     const voterGuideSaveResults = VoterGuideStore.getVoterGuideSaveResults();
     // console.log('onVoterGuideStoreChange voterGuideSaveResults:', voterGuideSaveResults);
-    if (voterGuideSaveResults && voter && voterGuideSaveResults.organization_we_vote_id === voter.linked_organization_we_vote_id) {
+    if (voterGuideSaveResults && voter && voterGuideSaveResults.organization_we_vote_id !== voter.linked_organization_we_vote_id) {
```
DaleMcGrew, line 104: Hi @SailingSteve why is this change necessary? I worry it changes the logic incorrectly.

DaleMcGrew:
👍

src/js/routes/VoterGuide/OrganizationVoterGuideEdit.jsx
SailingSteve: What was happening is that if `voterGuideSaveResults.organization_we_vote_id` was equal to `voter.linked_organization_we_vote_id` you would goToVoterGuideForDifferentElection with the same we_vote_id, you came back to line 104, and did the compare again, and then went right back to line 104, where they still were equal.  It was an endless loop, that locked up and crashed the session.

SailingSteve:
@DaleMcGrew If you know that that line 104 test was supposed to do, could you test it?  Also, I'm still testing but not finding anything that wasn't pre-existing.  Could you give it a test?  Emphasis on link navigation, and any red log lines.

src/js/routes/VoterGuide/OrganizationVoterGuideEdit.jsx
DaleMcGrew: Hi @SailingSteve, I just merged into develop a fix for the infinite loop and made a couple of other improvements to that component.

PR #3265, opened 2021-01-05
Update to react-router-5, Other package.json scripts working again

SailingSteve:
Commented out some un-needed logging
This task is pretty much done.
Starts up on localhost again (instead of 127.0.0.1, which Chrome does not like).


DaleMcGrew:
👍

PR #3270, opened 2021-01-30
Pulled the (168kb) moment.js out of textFormat.js 

SailingSteve:
textFormat.js is widely imported, and pulling moment out of it, allows us to make a separate smaller more rarely imported chunk that includes moment.js, this will allow us to only pull in moment when actually needed, and especially on the first page load.


DaleMcGrew:
Awesome optimization! 👍

PR #3272, opened 2021-03-08
Stop the menu from jumping 7.5 pixels to left when vertical scroll bar appears

SailingSteve:


DaleMcGrew:
Already merged in another pull request.

PR #3274, opened 2021-04-28
Checkin point on making WebApp having similar performance gains as we have with Campaigns

SailingSteve:
wevote.us has a Lighthouse performance of 27, this PR is at 76 for the first load of Ready.  
There still is low hanging fruit for further improvement.

**There is more to fix.**

Major changes:
* Does not use sass anymore, last generated css is in /src/css/main.css  -- This can be revisited, but eliminating the binary compile of node-sass makes installation and webpack configuration much easier
* Removed useless and problem causing: AppDispatcher.js and RequestDispatcher.js
* Removed semi-abandoned ReadyNoApi.jsx and associated files.
* Application.js, index.js, and root.js are no longer used.
* Backported from Campaigns: App.jsx, index.html, index.js, initializeAppleSDK.js, initializeFacebookSDK.js, initializejQuery.js, initializeMoment.js, .babelrc, webpack.config.js and much of package.json
* Used "replace all" with regex, to make most components lazy loaded -- This left imports interspersed with consts at the top of the file -- used the excellent "Optimize Imports" feature in JetBeans IDEs (Ctrl+Option+o) which did wonders, but sadly took the import of React off of the top line.  This loss is minor, compared to having all the imports optimized and sorted by library path/library name with one keystroke.  Of the roughly 430 files in this PR, most files changes are only import/lazy load changes.
* Fixed a good number of  "Cannot dispatch in the middle of a dispatch" errors
* Replaced abandoned 'react-iframe-resizer-super' with 'iframe-resizer-react' -- test this!
* We are no longer including the larger development libraries for React when building for production
* Removed chai and chai-enzme which we used for a semi-abandoned test from 2017 - they were not React 17 compatible.

Todo:
* Flicker/redraw of header menu, on first time it is clicked for each tab.
* Slow APIs:
  * voterGuidesFromFriendsUpcomingRetrieve takes 8.45 seconds to return two simple records (Ready signed in)
  * issuesUnderBallotItemsRetrieve fetches 34,868 rows of data and executes in the range of 500ms to 10 seconds
* superagent is currently commented out -- fix this


DaleMcGrew:
👍

PR #3280, opened 2021-05-14
Do not lazy load react-bootstrap components

SailingSteve:
They weren't in the main chunk anyways, just some similarly named components from other packages
Added a Suspense component for the header, makes the lazy load look better.
Since I cut back on many unessential API calls for the first load of a page, ValuesToFollowPreview on values/Opinions needed the google_civic_id if it was first page loaded, so had to do a voterBallotItemsRetrieve then a voterGuidesFromFriendsUpcomingRetrieve to populate it.
Restored the 960px max-width of the various pages

**There is still more to fix**

Todo:
* Slow APIs:
  * voterGuidesFromFriendsUpcomingRetrieve takes 8.45 seconds to return two simple records (Ready signed in)
  * issuesUnderBallotItemsRetrieve fetches 34,868 rows of data and executes in the range of 500ms to 10 seconds

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -81,7 +81,7 @@
         "moment-locales-webpack-plugin": "^1.2.0",
         "pre-commit": "^1.2.2",
         "react-hot-loader": "^4.13.0",
-        "stylelint": "^13.9.0",
+        "stylelint": "^13.13.1",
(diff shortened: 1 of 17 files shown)


DaleMcGrew:
👍

PR #3282, opened 2021-05-21
Performance improvements, added apiCalming.js

SailingSteve:
apiCalming.js allows us to pick api queries (mostly fired on componentDidMount) that are duplicative and useless, and suppress follow on identical queries -- without having to untangle the logic behind the "just in case" api calls.
Added back in the serviceWorker for PWA, but need better ssl certs on my local to fully test it.
Started changing from font-awesom icons to Material UI Icons, fa-twitter was the first.
Massive refactor on FollowToggle.jsx -- It was too hard getting the bootstrap menuing going again, so I basically rewrote it in Material UI.
SnackNotifier pulls in a lot of libraries, so to keep it out of the main chunk, I add it back in classes (mostly routes) that it is needed in.  We will probably find some more classes that need it while testing.
Set constants for specific image paths, instead of importing by image path.  Should be the same, but import caused problems with lazy loading.
Reworked HeaderBarSuspense, so that the lazy loading is less jarring.
Moved initializejQuery() to service.js, and eliminated it everwhere else, to avoid a second app redraw on startup. 
More lint fixes.
Replaced numberOfNeedlesFoundInString that had been added to the standard String object prototype, was causing problems in debug. 
 

** There is still more to fix **

Todo:
* Slow APIs:
  * issuesUnderBallotItemsRetrieve via CDN in 10 seconds, (333kb, 34,800 records)
  * voterGuidesFromFriendsUpcomingRetrieve 9.9 seconds, 2 simple records, 4kb via VoterGuideActions.js:106 (we were doing this twice in parallel)
  * issuesUnderBallotItemsRetrieve/CDN 9.9 seconds, 35k records, 334kb via Ballot.js:238
  * voterGuidesFollowedRetrieve in 2 seconds (retrieves 855 records, 804kb)
  * friendList/?kind_of_list=CURRENT_FRIENDS 1.5 seconds (and we were doing it twice in parallel)


DaleMcGrew:
👍

PR #3283, opened 2021-05-26
Added "page-content-container" to some pages (that needed it to be restricted in width)

SailingSteve:
Revived the equivalent of "fas fa-external-link-alt" with ExternalLinkIcon and fa-info-circle with InfoCircleIcon

Misc bug fixes

** There is still more to fix **

Todo:
* Slow APIs:
  * issuesUnderBallotItemsRetrieve via CDN in 10 seconds, (333kb, 34,800 records)
  * voterGuidesFromFriendsUpcomingRetrieve 9.9 seconds, 2 simple records, 4kb via VoterGuideActions.js:106
  * issuesUnderBallotItemsRetrieve/CDN 9.9 seconds, 35k records, 334kb via Ballot.js:238
  * voterGuidesFollowedRetrieve in 2 seconds (retrieves 855 records, 804kb)
  * friendList/?kind_of_list=CURRENT_FRIENDS 1.5 seconds (and we were doing it twice in parallel)


DaleMcGrew:
👍

PR #3287, opened 2021-06-10
Got donations working again.

SailingSteve:
Pulled some of the abandoned files into a new srcDeprecated directory that is not compiled.
Got sign in with SMS working again.
Display the SelectBallotModal automatically on the ballot page when needed, and not when not needed.
Misc bug fixes and style changes.
Worked around that long standing issue, where the HeaderBar menu underline disappears some times.
Enter in the last square of the token verification for email and sms, is equivalent to pressing the confirm button.
IssuesByBallotItemDisplayList was throwing an exception that talked about listeners, but it was really a timer that fired off after the component was dismounted.  We need to handle timers with the same care as listeners.


DaleMcGrew:
👍

PR #3288, opened 2021-06-25
Reworked timers, and replaced AppStore, to avoid multiple run time errors.

SailingSteve:
All these run time errors were bringing my debugging of the lazy loaded version to a crawl, so I fixed them.

**Timers:**
Timers (and Intervals) must be cleared in ComponentWillUnmount to prevent them trying to alter state on an unmounted component, which throws a red error.
Nulling timers when clearing them in ComponentWillUnmount is superfluous since the component is about to be unmounted, and the variables destroyed.  This has been removed everywhere and made "one line" in order to easily find with grep.  Grepping is necessary since the IDE and lint can't catch un-cleared timers and intervals.
Timers that are reused before the component gets unmounted, should be cleared and their tracking variable should nulled, to prevent the "this.timer" tracking variable from being reused and leaving the previous timer detached and un-clearable.

**AppStore/AppActions replacement**
Flux can not handle the interruption of a current dispatch, with another dispatch.
Calling actions from stores exposes us to serious dispatch within a dispatch errors.
The AppStore/AppActions code was an elegant solution to setting state, but overtime we started
calling it in places in the code where other dispatches were occurring.
So I replaced AppStore/AppActions with a simple observable store (AppObservableStore), basically just a globally accessible array of configurations, and wired it up with RxJS so that it would broadcast events when changed.


src/js/components/Navigation/HeaderBar.jsx
```diff
@@ -583,16 +609,6 @@ class HeaderBar extends Component {
       appBarCname += ' page-header__cordova';
     }
 
-    // We want the SelectBallotModal to appear on the ballot page (without a keystroke) if the page is empty and we have a textForMapSearch
```
DaleMcGrew, line 586: Hi @SailingSteve this fix was the result of a bug we found. Is there a reason to remove this?

DaleMcGrew:
👍  Looks great! My only comment is about if we need to remove the code starting on line 586 in src/js/components/Navigation/HeaderBar.jsx. 👍

src/js/components/Navigation/HeaderBar.jsx
SailingSteve: Go to the production wevote.us, clear all cookies, refresh and go to the ballot.  If you accept the default address, you go down a weird path, that doesn't take you to "seeing the ballot".   I suspected that was a bug.  Now you get the ballot select modal instead, and once you pick an election, you can see a ballot.

DaleMcGrew: Thank you for the explanation and the fix.

PR #3291, opened 2021-06-29
Fixed Facebook icon (mui/icons), update header on history push

SailingSteve:
Switched over to the new api friendListsAll -- which speeds up /News page load, due to only using one http channel instead of 6.  Someday we might have a voter with tons of friends, but with minimal friends it only returns about 2kb, and we would have fired off the 6 apis that it replaced anyways.
Fixed layout on back to ballot header
Reverted prior incorrect fix to StickyPopover.onMouseLeave
Removed onlyActiveOnIndex param that is no longer needed (or allowed) for the latest react-router-dom Link
Removed spreading in some files that I was otherwise making changes to.
Solved the flickering header on first visit of the tab, required more granular suspense fallbacks.  Especially in Ready, Ballot, News, Measure, App, and Values -- Indenting changes in each of those files makes it looks like tons of changes in diff.
Fixed Intro.jsx to be able to handle lazy loading of jQuery..
Added functional component LoadingWheelComp, so the loading wheel looks like a component in our code.
Made Measure.jsx respect our narrow side margins.
We had two versions of the WeVoteRouter.jsx -- deleted one of them.


DaleMcGrew:
👍

PR #3294, opened 2021-07-07
Multiple fixes for mobile vs desktop

SailingSteve:
Election countdown was finding an answer on ReadyLight, but not Ready
Fixed header height for office when mobile
Only show Address picker on HeaderBar for ballot and ready in mobile.
Moved 36 unused files to srcDeprecated, 5 of them had spreading warnings -- Classes with references that were commented out in active files were not moved.
resolves wevote/WebApp/pull/3293
Made the functions getters/setters/is/show consistent in AppObservableStore.js
Our homemade startsWith() was mixed in with the one built into JavaScript (that had an indistinguishable method signature!), eliminated our homemade version.


DaleMcGrew:
👍

PR #3295, opened 2021-07-13
Greatly improved polling of Donations after stripe API call to display current results.

SailingSteve:
All classes in the Donation directory plus SettingsVerifySecretCode.jsx are now identical in Campaigns and WebApp.


DaleMcGrew:
👍

PR #3296, opened 2021-07-19
Various fixes to Donate.js and common/component/donations

SailingSteve:
All classes in the Donation directory have been moved to the "common" directory and are identical in Campaigns and WebApp.
Two unused Donate classes moved to srcDeprecated.
There is a bug, where you when you navigate to /more/donate you get the "Thank you" page instead of the new donation page, I will fix this in a week when I get back.


DaleMcGrew:
👍

PR #3303, opened 2021-09-07
New method to build a bundle.js for Cordova

SailingSteve:
React.lazy can't be used in cordova builds since, we need a single bundle.js --  so bldSrcCordova.js has been created to make that single bundle build.  This script makes a temporary copy of the /src directory on the fly at /srcCordova then greps and replaces all instance of React.lazy with equivalent import statements.
The srcCordova directory has been added to .gitignore
A number of the React.lazy import paths had unnecessary path segments which have been cleaned up, a couple of loads did not have the webpackChunkName string which have been fixed, one React.lazy imported a file with ".jsx" at the end, etc.
Various lint issues have been fixed.
A new build target has been defined  ... npm run buildCordova
A new lint target has been defined to lint the srcCordova dir  ... npm run lintCordova

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -158,6 +158,7 @@ staticfiles
 /geo/data/
 /**/migrations/*.py
 !/**/migrations/__init__.py
+srcCordova
 src/js/config.js
 src/js/config-qa.js
 src/js/config-www.js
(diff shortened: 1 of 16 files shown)


DaleMcGrew:
👍

PR #3309, opened 2021-09-29
Top headers are mostly responsive at this point and now (mostly) use common components.  Pretty good in Cordova too.

SailingSteve:
Pulled all background image loading out of css, it couldn't work for both WebApp and Cordova since they have different paths to the images.
Friends still has responsive sizing issues in WebApp, Values has problems in Cordova -- these are the next issues.
Tested in iPhone 12 Max Pro cordova, iPhone 12 Max Pro webapp, and full screen webapp.


DaleMcGrew:
👍

PR #3310, opened 2021-09-30
Prep for cordova release: Layout looks good and links work on iPhones that are "Max" sized

SailingSteve:
Looks good in WebApp
There are many small layout issues to work on, and lots more Cordova, but overall in good shape.


DaleMcGrew:
👍

PR #3316, opened 2021-10-14
Prep for Cordova release: iPhone 13 Mini, 13, 13 Pro, 13 ProMax, SE, 8, 8+, & 11Pro passed a complete testing run

SailingSteve:
- WebApp is in good shape re: headers.
- avatar-generic.png loading had been copied and broken in many instances for Cordova, so has been replaced with a function.
- Made a function component for IPhoneSpacer to eliminate the copied (and broken) code.
- Many more header fixes -- there still are UX inconsistencies in the headers, but all that I could find via navigation look decent now.
- Pulled FriendsTabs out of Friends.jsx so that the HeaderBar can render it as needed.
- Testing sheet: https://docs.google.com/spreadsheets/d/1-_0jthH0coeWqs3KZh4TxSToSgscL-UP64wV7nwkIsI/edit?usp=sharing
-


src/index.html
```diff
@@ -13,7 +13,7 @@
     <meta http-equiv="Cache-Control" content="public" />
     <meta http-equiv="Pragma" content="no-cache" />
     <meta http-equiv="Expires" content="0" />
-    <meta name="viewport" content="width=device-width, initial-scale=1, maximum-scale=5, viewport-fit=cover">
+    <meta name="viewport" content="width=device-width, initial-scale=1, user-scalable=no, maximum-scale=5, viewport-fit=cover">
```
DaleMcGrew, line 16: Hi @SailingSteve, we needed user-scalable to be allowed so people can increase the font size. Was setting it to user-scalable=no critical for a problem you found?

DaleMcGrew:
👍

src/index.html
SailingSteve: I removed that an hour ago on my local, it was in the wrong index.html, so next PR will back that out.

PR #3318, opened 2021-10-30
Addressed Part C: Prop spreading is forbidden

flopmingo:
Our issue:
5 Files - Part C: Prop spreading is forbidden
https://github.com/wevote/WebApp/issues/3232

Pull request comments:

There was a missing file from our issue (Part C):
/Users/dalemcgrew/PycharmProjects/WebApp/src/js/components/VoterGuide/VoterGuideBallot.jsx
- A different version of this file named "VoterGuideBallotItemCompressed.jsx" seems to have been added to Part D.  As it no longer has the same filename, we ignored the "VoterGuideBallot.jsx" problem in Part C.

/Users/dalemcgrew/PycharmProjects/WebApp/src/js/components/VoterGuide/OrganizationPositionItem.jsx
- This file did not have any ESLint errors that we could detect

All the other prop spreading issues in the Part C issue have been addressed.

Thank you for the opportunity to contribute to your project :)


src/js/components/VoterGuide/OrganizationCard.jsx
```diff
@@ -1,295 +1,325 @@
...
-// `;
-
-const TwitterHandleWrapper = styled.span`
-  margin-right: 10px;
-`;
-
-const WebsiteWrapper = styled.div`
-  margin-left: 4px;
-`;
+import { Twitter } from '@material-ui/icons';
```
DaleMcGrew, line 1: Thank you @flopmingo for your pull request! I'm noticing that this entire file is showing every line has changed, and I wonder if it might be different line breaks in your updated file, or if it was deleted and then replaced? This doesn't have to stop the pull request, but I just note that it is odd in case you have any ideas.

mgg143: Thanks, @DaleMcGrew .  We will look into this further and get back to you.  Thanks for being so supportive and for allowing us the opportunity to contribute to your project :)

flopmingo: Sorry for the late reply! We were working on the code through VS code and I believe it may have been because we changed the EOL sequences from CRLF to LF through VS code. As @mgg143 has said, thank you for giving us this learning opportunity!

DaleMcGrew: Thank you for doing the research, and thank you for contributing to We Vote! Could you please update your pull request with the line breaks set back to CRLF? That way we'll be able to see all of the actual changes more easily.

flopmingo: Hello Dale! I have been attempting to revert back to an older version of the file by using git checkout file. However, when I revert back to an older version in VS code the line breaks are still set as LF. I'm afraid that simply just changing the files back to CRLF will still highlight every line as a change, thus still making it difficult to see the actual changes. Should I change the line breaks on my current file and update the PR or should I continue to try to find a way to revert back to an older version?

DaleMcGrew: I suspect VS code has a setting to change the default line breaks to CRLF, so I would focus on that solution. I would suggest a Google search for "visual studio crlf git". This is interesting:
https://developercommunity.visualstudio.com/t/git-detects-line-endings-as-changes/807495

If there isn't a solution in that, it might be fastest to revert the file and then make the changes again from scratch.
PS. I'm not a fan of Visual Studio for open source projects. I recommend the JetBrains tools, and more specifically WebStorm which offers a free student license.

DaleMcGrew:
Addressing in another pull request.

PR #3319, opened 2021-10-31
Fixed spacing bugs on Ballot for mobile and desktop responsive website. Should not affect Cordova.

DaleMcGrew:
Fixed spacing bugs on Ballot for mobile and desktop responsive website. Should not affect Cordova.


SailingSteve (approved):
Some of these look like them might affect cordova, but let's just merge them and I will do another full test run (after getting past the cookie issues I am currently working on).

PR #3323, opened 2021-11-11
Fixed a bldSrcCordova reversion, that caused an exception when clicking ballot immediately after the initial load.

SailingSteve:
Cleared some lint problems
Made the "We're sorry" page more useful.
Released to TestFlight as Version 1.1.1 build 2.2.1.3
Removed SettingsProfilePicture for Cordova
Added privacy policy to SignInModal for Facebook


DaleMcGrew:
👍

PR #3340, opened 2021-11-21
Changes for Android Cordova (will be build 2.2.1.0)

SailingSteve:
Works great on a physical phone, and Android Medium and Large, more variants to be tested.
I had to move some functions into a new hrefUtil.js to avoid a dependency cycle, which forced small include changes in many files.


DaleMcGrew:
I had to do some manual merging, and all of these changes are now merged into develop.

PR #3350, opened 2021-12-11
Good working prototype for getting Apple Contacts in Cordova

SailingSteve:
Loads the contacts from the Apple Contacts API
Most of the look and feel, most of the navigation for wevote/WebApp/issues/3301


DaleMcGrew:
👍

PR #3351, opened 2021-12-22
Sign-in with Twitter is now many times faster

SailingSteve:
Completes in a couple of seconds if, in another tab, you are on the twitter site and signed-in.
The voter photo arrives about 15 seconds later, with the anonymous photo displayed initially.


DaleMcGrew:
👍

PR #3356, opened 2022-01-27
Re-enabled FullStory, upgraded to WebPack 5 (to clear snyk warnings)

SailingSteve:
Manual merge after git issue


src/js/pages/Values/ValuesList.jsx
```diff
@@ -130,6 +131,7 @@ export default class ValuesList extends Component {
       }
     });
 
+    // Cordova Note Jan 26, 2022: The <PageContentContainer> wrapper replaced a few days ago with an empty <ValuesListWrapper> which will probably cause problems.
```
DaleMcGrew, line 134: PageContentContainer was removed from src/js/pages/Values/ValuesList.jsx, since it is not on its own page. It is only included on WebApp/src/js/pages/Values/VoterGuidesUnderOneValue.jsx, which has its own "PageContentContainer" wrapper. src/js/pages/Values/ValuesList.jsx should probably be moved into the Values component folder.

DaleMcGrew:
👍 This is great, thank you!

src/js/pages/Values/ValuesList.jsx
SailingSteve: Ok

PR #3361, opened 2022-02-07
Allows all customized instances to use a consistent index.html as built by the API server

SailingSteve:
Uses common webPackPostBuild.js to create the main.name.js file.
Clears a warning related to WebPack from production builds.
Moved the abandoned Gulpfile.js to srcDeprecated


DaleMcGrew:
👍

PR #3365, opened 2022-02-09
Sends Google Analytics updates on page change

SailingSteve:
Apple Signin shows "loading" on button until api loaded (less than one second, on first presentation)
Apple Signin handles API not loaded, gracefully and with the single initial click
Pulled all of the state updating out of the TwitterSignInProcess render section to clear runtime errors.


DaleMcGrew:
👍

PR #3369, opened 2022-02-16
Fix to have pushHistory display a snack pop up on the destination page.

SailingSteve:
Would have been easier and simpler to put on the App render, but then all that snack stuff would have
been in the main chunk.


DaleMcGrew:
Thank you @SailingSteve! 👍

PR #3378, opened 2022-03-03
Added a GoogleAutoComplete common component

SailingSteve:
Uses ErrorPro/react-google-autocomplete which has the great beneficial feature of not loading the google libraries until needed.
I'm not sure I got all the styles and edge cases right, but will be happy to fix them if there are issues.
Unfortunately recent changes (#3376) to AddressBox and EditAddressInPlace conflicted with this refactor/reimplementation that was in-process, and these changes were not included -- it seems like the same issues for the changes in #3376 are now covered in a different way, or were made moot.  Apologies in advance.
Did not test in Cordova yet.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,8 +1,14205 @@
 {
   "name": "wevoteusa",
   "version": "1.0.0",
-  "lockfileVersion": 1,
+  "lockfileVersion": 2,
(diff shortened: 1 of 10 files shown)


DaleMcGrew:
👍

PR #3381, opened 2022-03-07
Restored GoogleAutoComplete as a functional component reference case

SailingSteve:
This works with all my test cases, it pulls the initial address from the VoterStore, and it allows voter entered arbitrary addresses.


DaleMcGrew:
👍

PR #3382, opened 2022-03-08
Save current voter address in all GoogleAutoComplete instances, when any one of them changes the store

SailingSteve:
To fix the issue with location_guess_closed cookie


DaleMcGrew:
Great! 👍

PR #3388, opened 2022-03-20
[Issue #3234] This commit resolves prop spreading error with the sett…

fdiazq:
…ings in the slick slider. Worked on this with @Johnny01232003.

This commit resolves 8 Files - Part E: Prop spreading is forbidden #3234. This commit resolves only the file IntroNetwork.js's prop spreading error.
The prop spreading operator was removed from the file and replaced with explicit declaration of props from the slick slider's settings.


DaleMcGrew:
Thank you @fdiazq for this check in! Approved. 👍

PR #3390, opened 2022-03-22
Theme needs to be passed into styled-components in MUI5.  Using MUI style breakpoints everywhere.

SailingSteve:
Using MUI style breakpoints avoids a theme passing issue, that probably was solvable, but consistency (and better readability is good).
HowItWorks is now working.

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -13,7 +13,8 @@
     [
       "babel-plugin-styled-components",
       {
-        "displayName": true
+        "displayName": true,
+        "fileName": false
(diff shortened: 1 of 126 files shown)


SailingSteve:
too early to merge

SailingSteve:
The breakpoints seem to work fine.  HowItWorks is still messed up, and the problem is not obvious... I'll keep on it.

DaleMcGrew:
👍

PR #3394, opened 2022-03-25
New style for HeaderBar -- highlight, instead of underbar

SailingSteve:
A few more MUI 5 cleanups.
Cleared some moment warnings in ElectionCountdown
FriendDisplayOuterWrapper was passing a not used 'previewMode' that was throwing warnings
Removed some low value comments in styles, to try to confirm a stackoverflow article that said that comments were confusing the babel-plugin-styled-components plugin (still working on this)
Didn't touch the lint warnings in PositionRowEmpty on the assumption that they are a work in progress.

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -10,6 +10,8 @@
         "loose": true
       }
     ],
+    ["@babel/plugin-proposal-private-methods", { "loose": true }],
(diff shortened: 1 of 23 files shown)


DaleMcGrew:
👍

PR #3396, opened 2022-03-26
 Filter changed files Beta Give feedback New style for HeaderBar -- highlight instead of underbar

SailingSteve:
A few more MUI 5 cleanups.
Cleared some moment warnings in ElectionCountdown
FriendDisplayOuterWrapper was passing a not used previewMode that was throwing warnings in the console
babel-plugin-styled-components plugin is now working with MUI 5 (in almost all cases)
Fixed all lint errors, but not all lint warnings ... warnings could be the result of work in progress


DaleMcGrew:
👍

PR #3401, opened 2022-04-04
Fix to Avatar for Cordova and minor bug fix to Ballot

SailingSteve:
Did a tiny bit of Cordova testing to confirm fix to apache/cordova-plugin-device


DaleMcGrew:
👍

PR #3410, opened 2022-04-22
Resolves prop spreading in 3 files [Issue #3232]

fdiazq:
This resolves errors in pull request Addressed Part C: Prop spreading is forbidden #3318 in Issue #3232

Modifications were made to align spacing and apply changes included in the older pull request

### DIFF
--- a/src/js/components/VoterGuide/OrganizationCard.jsx
+++ b/src/js/components/VoterGuide/OrganizationCard.jsx
@@ -68,6 +68,7 @@ export default class OrganizationCard extends Component {
 
   // eslint-disable-next-line camelcase,react/sort-comp
   UNSAFE_componentWillReceiveProps (nextProps) {
+    const { organizationPositionsRequested } = this.state;
     // console.log('OrganizationCard, componentWillReceiveProps, nextProps:', nextProps);
     if (nextProps.organization && nextProps.organization.organization_we_vote_id) {
       this.setState({
@@ -84,7 +85,7 @@ export default class OrganizationCard extends Component {
         this.setState({
           organizationPosition,
         });
-      } else if (!this.state.organizationPositionsRequested) {
+      } else if (!organizationPositionsRequested) {
         OrganizationActions.positionListForOpinionMaker(nextProps.organization.organization_we_vote_id, true);
         this.setState({
           organizationPositionsRequested: true,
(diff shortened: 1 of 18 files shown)


src/js/components/VoterGuide/OrganizationVoterGuideCandidateList.jsx
```diff
@@ -18,7 +18,18 @@ export default class OrganizationVoterGuideCandidateList extends Component {
               contest_office_name={this.props.contest_office_name}
               link_to_ballot_item_page
               organization_we_vote_id={this.props.organization_we_vote_id}
-              {...child}
+              ballotItemDisplayName={child.ballot_item_display_name}
```
DaleMcGrew, line 17: Thank you @fdiazq, these changes look correct. You will also need to update the variables names coming into `OrganizationVoterGuideCandidateItem.propTypes` in /src/js/components/VoterGuide/OrganizationVoterGuideCandidateItem, and make sure to convert all uses of these incoming prop variable names to camelCase. While you are working on this, please update the remaining incoming snake_case variables to camelCase (ex/ link_to_ballot_item_page).

src/js/components/VoterGuide/OrganizationCard.jsx
```diff
@@ -137,16 +137,45 @@ export default class OrganizationCard extends Component {
       if (this.state.organizationPosition.vote_smart_rating) {
```
DaleMcGrew: I would recommend pulling `organizationPosition` out of the state variable above this if/else block like this first:
```const { organizationPosition } = this.state;```
Then below, you don't need to write out `this.state.organizationPosition` multiple times.

File: src/js/components/VoterGuide/OrganizationCard.jsx [hunk 3]
```diff
@@ -137,16 +137,45 @@ export default class OrganizationCard extends Component {
...
         positionDescription = (
           <PositionRatingSnippet
-            {...this.state.organizationPosition}
+            ballotItemDisplayName={this.state.organizationPosition.ballot_item_display_name}
```
DaleMcGrew: For each of these components (PositionRatingSnippet, PositionSupportOpposeSnippet, PositionInformationOnlySnippet), make sure the explicit incoming variables are listed in `propTypes` within each component, and the updated camelCase variables replace the prior snake_case format throughout the entire components.

DaleMcGrew: Since each of these components is included multiple places in the code (`<PositionInformationOnlySnippet` is shown in 4 places), you will need to resolve the prop type spreading in all of those locations too. Resolving prop type spreading is like a recursive problem.

src/js/components/VoterGuide/VoterGuideDisplayForList.jsx
```diff
@@ -30,28 +30,28 @@ class VoterGuideDisplayForList extends PureComponent {
     const position = this.props;
 
     const {
```
DaleMcGrew: This destructuring from props can be done at the beginning of the render, and then `this.props` can be removed from the start of the variable names.

DaleMcGrew:
Looks great! Excellent work. 👍

PR #3421, opened 2022-05-06
Fix to layout of SignInModal when Virtual Keyboard is displayed.

SailingSteve:


DaleMcGrew:
👍

DaleMcGrew:
On testing, I found that when I click in the email or phone box, the modal shifts to the left, mostly off screen. Prior to this pull request it was shifting mostly off screen to the right.

SailingSteve:
Is the code on the live server now?

SailingSteve:
Those two modals work fine on a phone, but not in the Chrome devtools simulator.
The sign in modal is too high on the phone, but looked good when I checked it in.
I'll take a look tomorrow.

PR #3425, opened 2022-05-10
Removed older SignInModal in favor of newer SignInModalSimple.jsx

SailingSteve:
SignInModalSimple was a copy of SignInModal with changes, that had diverged from the original, in code and style.
SignInModal has been renamed with this PR to SignInModalOriginal.jsx and is deprecated
SignInModalSimple has been renamed SignInModal
SettingsAccount has been renamed to SettingsOptionPanel
Only WebApp has been tested, and a bunch of Cordova constants and code from SignInModalOriginal still needs to be ported over to Cordova.
The fix from Friday with the Email and Phone entry dialogs has been ported over.


DaleMcGrew:
👍

PR #3431, opened 2022-05-15
First pass at a Cordova build that uses relative dimensions for header heights, instead of experimentally determined constants.

SailingSteve:
Works well on iPhone Mini and 13 Max Pro, other TBD, and Androids.  Retested WebApp, still good.
Still relies on the older way (cordovaScrollablePaneTopPadding) for WebApp, but that can and should be refactored out.
Made a common component for the many box-shadow styles
Refactored out the styles for the two versions of the Ready page.
I think this resolves the sign in pane positioning in WebApp, but won't know for sure until it gets on the live server.

### DIFF
--- a/node/buildSrcCordova.js
+++ b/node/buildSrcCordova.js
@@ -84,7 +84,7 @@ fs.remove('./build').then(() => {
               fileRewriterForCordova(path);
             }
           }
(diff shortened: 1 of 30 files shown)


DaleMcGrew:
Thank you! 👍

PR #3432, opened 2022-05-19
Rework of the SettingsDashboard for iPads and Android "fold" phones.

SailingSteve:
Now looks "the same" as it does in desktop mode.
The breakpoints that Bootstrap uses are bad for Cordova, sometimes we are lucky and they are close enough and work, sometimes they are useless.  On the SettingsDashboard, the page goes into side-by-side mode at 768px in the WebApp, but would need something like 850px in Cordova for iPads, so I replaced Bootstrap, with Cordova sensitive code in a few places.
We should not create any more Bootstrap based pages, with Bootstrap that "the page is done" feeling is an illusion.  Changing the breakpoints in Bootstrap is a mess, it would be possible to copy and rewrite the library on the fly for Cordova, but that would be fragile.
Cordova is in decent shape in iOS, Android has some new build dependency problem that needs to be resolved -- those build issues can be very time consuming for a 5% of the time on Android engineer.

### DIFF
--- a/src/js/common/utils/cordovaUtils.js
+++ b/src/js/common/utils/cordovaUtils.js
@@ -755,6 +755,10 @@ export function polyfillFixes (file) {
   }
 }
 
+export function isCordovaWide () {
+  return isIPad() || isAndroidSizeFold();
+}
+
 export function cordovaLinkToBeSharedFixes (link) {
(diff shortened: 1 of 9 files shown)


DaleMcGrew:
👍

PR #3459, opened 2022-06-21
Progress toward reviving the Chrome Extension

SailingSteve:
Still takes 3 minutes to load pages (should take "a few" seconds).
Still has parts that don't fully work.


DaleMcGrew:
👍

PR #3493, opened 2022-08-01
Fixed twitter sign in for WebApp

SailingSteve:
Some early changes for the Cordova release
No longer need to suppress the round avatars in Cordova, so removed that code.
In PositionRowEmpty.jsx, needed to have the static const NUMBER_OF_POSITIONS_REQUIRED_TO_TURN_OFF follow the react.lazy, otherwise the Cordova build script fails.
If the "Ballot for 37 Lake Shore Drive, Oakland, CA 94611, USA" string is too long, use css to truncate it with ellipses, in both Cordova and WebApp.


DaleMcGrew:
👍

PR #3495, opened 2022-08-03
Intermediate checkin #2 for Cordova release

SailingSteve:
Various page size and header size adjustments if Cordova
Reworked HeaderBackTo for Cordova, we now use the same back arrow (and header layout) as WebApp
Reworked SettingsSectionFooter, so that all the links have the same style, removed OpenExternalWebsite for these links in Cordova, since they open completely new contexts that don't share cookies and allow one to navigate to the ballot page in the webapp in those contexts.


DaleMcGrew:
👍

PR #3497, opened 2022-08-04
WebApp Cordova iOS working, a few optional fixes to go

SailingSteve:
Not yet tested on physical devices, only tested on 3 sizes iPhone 13 mini, iPhone 13 Max Pro, and iPad 12"
Header fixes
Did a few Cordova only ui cleanups, a few fixes to both Cordova and Android.
Big changes to "recently" introduced OpenExternalWebsites that were meant to open the same instance of WebApp in another tab, those break Cordova, so added conditional logic for each.
Put the "These are some ways your information is used or shown:" table into a scrollable div, if if could not fit within the window on smaller devices.

### DIFF
--- a/package.json
+++ b/package.json
@@ -55,7 +55,7 @@
     "eslint-config-airbnb": "^18.2.1",
     "eslint-plugin-import": "^2.22.1",
     "eslint-plugin-jsx-a11y": "^6.4.1",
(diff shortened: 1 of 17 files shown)


DaleMcGrew:
👍

PR #3498, opened 2022-08-10
Popovers now close when opening PositionDrawer

### LINKED ISSUE #3492: Opening PositionDrawer should close all modals opened on Ballot
1) Start on the ballot page, and look at one candidate.
Screen Shot 2022-08-01 at 1 06 42 PM

2) Hover over an endorsement (ex/ YIMBY Action) so that the popover opens.
Screen Shot 2022-08-01 at 1 03 57 PM

3) Without moving your mouse, click on the endorsement so that the drawer opens:
Screen Shot 2022-08-01 at 1 05 42 PM

The bug is that the popover should close when the action to open the drawer is clicked.

MGarc758:
Fixes #3492 
Added a reference to StickyPopover, so every time onClickShowPositionDrawer is called, it will also call the closePopover method in the StickyPopover component and close the popover before opening the PositionDrawer.


DaleMcGrew:
Looks great @MGarc758, thank you!

PR #3499, opened 2022-08-10
WebApp Cordova iOS candidate for Apple Store submission

SailingSteve:
Lots of small changes
After spending a day on a really tough bug with the About header (it kept on displaying the default offset even though the debugger would not stop on the default line.  console.logs before the default line would not log, changing the offset in the default line, did adjust what was seen on screen.  Looks almost like a parallel thread is running in Cordova), I just gave up on it for now.  About still has a very old, non-standard header.  Instead, I am just calling the FAQ page where "team" and "about" are called in WebApp.


src/js/components/SignIn/SignInOptionsPanel.jsx
```diff
@@ -477,67 +477,65 @@ export default class SignInOptionsPanel extends Component {
...
+                  <FacebookContainer>
+                    <Facebook fontSize="large" style={{ transform: 'translateY(-3px)' }} />
+                  </FacebookContainer>
+                  <span className="u-margin-left--sm" />
                 </div>
+                )}
               </div>
             ) : null}
-            {!hideVoterPhoneEntry && (
+            {!hideVoterPhoneEntry && isWebApp() && (
```
DaleMcGrew, line 538: Hi @SailingSteve, how does the voter enter a phone or email in Cordova?

DaleMcGrew:
👍

PR #3504, opened 2022-08-19
Cordova release candidate for Android

SailingSteve:
External websites for Cordova Android must use SSL or will fail (need https)
Rewrote getAndroidSize() to use a much more consistent way to determine size.
Lots of header adjustment, more use of calculated offsets, but still not 100%
Constrained the text on OfficeItemCompressed and MeasureItemCompressed to minimize ugly layout on small devices.  This fix will be of great value for iOS too, on the next release.
HowItWorks no longer fills the entire screen in mobile.
Lots of work (tiny code changes) on VoterPhoneEmailCordovaEntryModal to cleanly remove this dialog, after it calls the SettingsVerifySecretCode.
Put the ballot body into a horizontally scrolling container, as a temporary measure, until each section is in a horizontally scrolling container -- this was necessary for Android.
Made the desktop style SettingsDashboard, for androidWide ... tablets and unfoldable wide phones.
Fixed an ugly horizontal scrolling issue on medium sized (pretty much as small as they make them these days) Android phones.
Updated DeviceDialog and Attributions.
Added a app-ads.txt to WebApp, to clear the error/warning in the facebook console.  Probably needed for Android release.
Android sign-in works fine on Browser Stack, but only so-so in the simulators, and not at all in a tethered physical phone with debug apks.


src/js/common/components/FAQBody.jsx
```diff
@@ -49,22 +50,31 @@ export default class FAQBody extends Component {
         <Suspense fallback={<></>}>
           <OpenExternalWebSite
             linkIdAttribute="weVoteEducationWebsite"
-            url="http://WeVoteEducation.org"
+            url="https://www.WeVoteEducation.org"
```
DaleMcGrew, line 53: Unfortunately WeVoteEducation.org is not SSL yet either, and won't be in the near future.

SailingSteve: It is, try it in a browser  https://www.WeVoteEducation.org
I had to add the www nonsense to make it work.  Maybe it is a redirect in the webserver, that handles this.
Screen Shot 2022-08-20 at 1 00 45 PM

DaleMcGrew:
👍

PR #3506, opened 2022-08-23
Device Dialog was showing up in WebApp

SailingSteve:
Added a bit of babel config for Android
Remove sign in modal after verifying email in desktop and mobile.


DaleMcGrew:
👍

PR #3512, opened 2022-09-14
Enabled importing contacts from iPhone in Cordova

SailingSteve:
Android not yet supported.
Works signed in or signed out, but only once per device_id as per existing logic. 
Deleted the old page/Startup prototype for this feature.  Move the pieces of this prototype that did get used, to where they are currently used.
Added support for the new larger front facing camera hole on iPhone 14 plus, pro, and pro max.


DaleMcGrew:
👍

PR #3513, opened 2022-09-18
Added start-win script

SailingSteve:
Submitted from a Windows PC.
The added script is needed for Windows runs.
This looks good, but is an experiment to make sure that Windows CRLF, are converted to LF with the committ.


DaleMcGrew:
👍

Issue #3511, opened 2022-09-14
Error when running WebApp on Windows


fdiazq:
### Please describe the issue (What happens? What do you expect?)
An error occurs when running the command npm start after the setting up environment by following steps for manual installation of WebApp from this link.
https://github.com/fdiazq/WebApp/blob/develop/docs/installing/ENVIRONMENT.md
### Steps to reproduce the problem (1, 2, 3...), including links
Once NodeEnvironment has been setup and code brought into environment.
When running npm start there is an error message " 'PROTOCOL' is not recognized as an internal or external command,
operable program or batch file."

![webapperror](https://user-images.githubusercontent.com/46939004/190037374-5472e07a-4054-429b-874a-ca551005d478.PNG)

SailingSteve:
If you got that far into the setup, you might just be able to use the new `npm run start-win` command in my latest checkin, which is not merged in yet.
For a quick hack, you could just try removing PROTOCOL=HTTP from the "start" line in package.json -- line 22

DaleMcGrew:
Hi @fdiazq could you please test to see if the latest release fixes this problem on your Windows machine?

DaleMcGrew:
Thank you for verifying @fdiazq!

PR #3515, opened 2022-09-18
New WebApp install procedure for Windows

SailingSteve:
This is for a basic install without SSL
There are lots of manual steps that could be scripted.
Tested that Git submits worked from Windows and that the Windows CRLF did not make it up to Git. 
The Windows PowerShell is a lousy substitute for a Bash shell, it probably would be possible to only use PowerShell in this install where absolutely necessary, and use Bash for the rest, and leave the IDE configured for Bash.


SailingSteve:
Should solve https://github.com/wevote/WebApp/issues/3511

DaleMcGrew:
Great instructions, thank you! I will ask someone on the QA team to run through it and provide feedback if there is any needed. 👍

PR #3516, opened 2022-09-20
Fixed the double address in the "Enter your address to find out when" link to dialog

SailingSteve:
More fixes for the iPhone 14 larger cutout, for the settings page Fixed the ambiguity between when the bottom footer menu vs the top header menu appears. All the WebApp code for Cordova iOS build 2.3.1.3, which is now in TestFlight and on the Engineering drive


DaleMcGrew:
👍

PR #3520, opened 2022-09-26
Detect AppleId with no contacts, and put up appropriate message.

SailingSteve:
Was unable to setState, within a native (iOS) callback, so added a workaround. 
Better padding for WeVote logo at different screen sizes


DaleMcGrew:
👍

PR #3529, opened 2022-10-17
Minor changes for Android release

SailingSteve:
About half way through testing, but looking very good
Some changes to avoid "Cannot dispatch in the middle of a dispatch" and "Can't perform a React state update on an unmounted component" in login by Email.
Some style changes for Tablets.


DaleMcGrew:
👍  Thank you @SailingSteve. I'm sorry for all of the hassle around "Cannot dispatch in the middle of a dispatch"!

PR #3558, opened 2022-11-07
Turned on some logging in AppObservableStore for diagnosing mobile browser sign in problem

SailingSteve:
It may be simply that nginx has started to append 'www.' to 'wevote.us' when returning the website URL as seen by the python server.


DaleMcGrew:
👍

PR #3563, opened 2022-11-08
Remove 'www.' from incoming URLs

SailingSteve:
www is not necessary, and messes up oAuth lists of acceptable domains.
Turned off the logging I added over the last few days to diagnose the iOS sign in dialog problem in mobile safari.


DaleMcGrew:
👍

PR #3571, opened 2022-11-21
Style changes and general fixes for Share Modal/Footer in Cordova iOS

SailingSteve:
Moved a lot of copied code from ShareModal and ShareButtonFooter to common files
Fixed the red error on Ballot complaining that lists must have unique keys
Also tested in WebApp and Mobile


DaleMcGrew:
This is great refactoring @SailingSteve, thank you!

PR #3572, opened 2022-11-21
Wrap too long election titles like 'District of Columbia General Election'

SailingSteve:
Update babel-loader to the latest version to clear https://github.com/wevote/WebApp/pull/3570
Deduped the multiple identical results when searching for candidates on the ballot page (OfficeItemCompressed.jsx)
Made multiple styled components have unique names, so that I can find the right one when searching backward from the dom while testing Cordova.
Put more Facebook api startup behind the webAppConfig.ENABLE_FACEBOOK for debugging startup issues in Cordova.
A bunch of conditional and computed styling for the the Friends sub-tabs for iOS.
Handle long election names for mobile like 'District of Columbia General Election'
Moved getPageKey() into its own file to avoid circular inclusion.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,14 +1,18366 @@
 {
   "name": "wevoteusa",
   "version": "1.0.0",
-  "lockfileVersion": 1,
+  "lockfileVersion": 2,
   "requires": true,
+  "packages": {
+    "": {
+      "name": "wevoteusa",
+      "version": "1.0.0",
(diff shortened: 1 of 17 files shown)


DaleMcGrew:
I believe I captured all of the changes in https://github.com/wevote/WebApp/pull/3575
I did have to fix device.isVirtual crash to get WebApp to load.

PR #3580, opened 2022-11-30
Test build for Cordova iOS 2.3.1

SailingSteve:
Facebook sign in is not working (again) in Cordova, same with share with facebook.
Otherwise ready to go


DaleMcGrew:
👍

PR #3584, opened 2022-12-04
Allow Facebook signin from a browser that has not previously signed into facebook.com

SailingSteve:
New manual procedure in SECURE_CERTIFICATE.com for fully signing out of facebook.com
Change to webpack.config.js to allow https://wevotedeveloper.com:3000 instead of localhost on developer computers, for Sign in with Facebook.


DaleMcGrew:
👍

PR #3593, opened 2022-12-15
Drastic simplification of Sign in with Facebook

SailingSteve:
This is all prep work for reviving Cordova Sign in with Facebook.
Cuts the Sign in with Facebook duration to about 1/3 as long with far fewer API calls I left a lot of commented out code in place, until I'm sure that there aren't any edge cases that are not covered.

Tested:  First login without a prior facebook login in the browser, Login after a logout, and Login when already logged in with Twitter.

Added a updateHosts.js script, so that less experienced developers won't have to manually edit linux/macOs system files.


DaleMcGrew:
👍

PR #3595, opened 2022-12-27
Added Axios to make requests uniform across browser and Node.js. Added a script that generates capabilities for all BrowserStack-supported devices.

nyette:
### Changes

Added Axios to make requests uniform across browser and Node.js. Added a script that generates capabilities for all BrowserStack-supported devices.


DaleMcGrew:
Thank you @Nyette! I wonder if tests/browserstack_automation/conf/capabilities.json should be renamed to something like "defaultCapabilities.json". That way it could be used if the "capabilities.json" file hasn't been updated locally, and yet not be overridden locally. Unless you think developers should periodically check in the updates they make locally to capabilities.json when they run your new script? If not, then capabilties.json should be added to the .gitignore file.

nyette:
We don't know how often BrowserStack updates their list of supported devices, so you're right; it might be better to rename the pushed `capabilities.json` file like you said and ignore more recent copies. I will push these changes up later today.

PR #3596, opened 2022-12-29
Set default capabilities in the event that capabilities.json does not exist.

nyette:
### Changes

Set default capabilities in the event that `capabilities.json` does not exist.


DaleMcGrew:
Nice and clean solution! Thank you. 👍

PR #3601, opened 2022-12-31
Fixed buildCapabilities.js, so that mobile capabilities without app URLs would no longer be generated.

nyette:
### Changes

Fixed `buildCapabilities.js`, so that mobile capabilities without app URLs would no longer be generated.


DaleMcGrew:
👍

PR #3616, opened 2023-02-01
Added maximum email length requirement in validateEmail function

tranngocsongtruc:
Now the User can only enter the email address which is 6 to 254 characters long.

Added 254 character maximum email address length to validateEmail function in WebApp/src/js/utils/regex-checks.js


DaleMcGrew:
👍  Thank you @tranngocsongtruc for adding this new requirement!

PR #3625, opened 2023-02-27
Removed the 0 that appears whenever the user has no contacts with accounts

nyette:
### Changes

Removed the `0` that appears whenever the user has no contacts with accounts.

Thank you for your help with this fix, @tranngocsongtruc.


DaleMcGrew:
Good catch! Thank you @Nyette and @tranngocsongtruc. 👍

PR #3650, opened 2023-05-03
candidate-for-extension api, changes saved to DB from extension, overrides changes sent in on the URL

SailingSteve:


DaleMcGrew:
👍

PR #3652, opened 2023-05-06
[WV-78] Fixed a bug that allows the user to create posts without content

nyette:
### Changes

Fixed a bug that allows the user to create posts without content


DaleMcGrew:
👍

Issue #51, opened 2016-02-18
Reconcile service.js functions


lisamcho:
There are functions in service.js such as "candidatesRetrieve" that seem to be duplicated by BallotAPIWorker in BallotStore.js
Needs to delete the ones we are not using.

nf071590:
Hey Lisa, we should be migrating towards jQuery $ajax function. I will work on cleaning up these requests.

DaleMcGrew:
It looks like @nf071590 resolved this.

PR #3665, opened 2023-05-31
[Issue #51] Redirect user to home page after sign out in Security & Sign In tab

Anirudhx234:
JIRA Issue #51

Sign out in the Security and Sign In tab in user settings now redirects the user to the home page rather than to the ballot page.


DaleMcGrew:
👍

PR #3668, opened 2023-06-02
[Issue #73] Updated Twitter handle to componentDidUpdate

Anirudhx234:
JIRA Issue #73

Updated UNSAFE_componentWillReceiveProps() to componentDidUpdate

### DIFF
--- a/src/js/pages/TwitterHandleLanding.jsx
+++ b/src/js/pages/TwitterHandleLanding.jsx
@@ -52,35 +52,17 @@ export default class TwitterHandleLanding extends Component {
   }
 
   // eslint-disable-next-line camelcase,react/sort-comp
-  UNSAFE_componentWillReceiveProps (nextProps) {
-    // console.log('TwitterHandleLanding componentWillReceiveProps');
-    let activeRoute = '';
-    let nextTwitterHandle;
-    // eslint-disable-next-line react/prop-types
-    if (nextProps.match && nextProps.location) {
-      // eslint-disable-next-line react/prop-types,camelcase
-      const { location: { pathname }, match: {params: { twitter_handle} }  } = nextProps;
-      activeRoute = pathname;
-      nextTwitterHandle = twitter_handle;
-      // eslint-disable-next-line react/prop-types
-    } else if (this.props && this.props.location) {
-      // eslint-disable-next-line react/prop-types
-      const { location: { pathname } } = this.props;
-      activeRoute = pathname;
-    }
-
-    const { twitterHandle } = this.state;
-    this.setState({
-      activeRoute,
-    });
-    // console.log('TwitterHandleLanding componentWillReceiveProps activeRoute:', activeRoute);
-    if (nextTwitterHandle && twitterHandle.toLowerCase() !== nextTwitterHandle.toLowerCase()) {
-      // We need this test to prevent an infinite loop
-      console.log('TwitterHandleLanding componentWillReceiveProps, different twitterHandle: ', nextProps.match.params.twitter_handle);
+  componentDidUpdate(prevProps) {
+    // if-statement prevents infinite rendering
+    if (prevProps.match.params.twitter_handle !== this.props.match.params.twitter_handle ||
+        prevProps.location.pathname !== this.props.location.pathname) {
+      const { match: { params: { twitter_handle: nextTwitterHandle } }, location: { pathname: activeRoute } } = this.props;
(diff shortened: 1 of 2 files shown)


src/js/pages/TwitterHandleLanding.jsx
```diff
@@ -52,35 +52,17 @@ export default class TwitterHandleLanding extends Component {
...
-
-    const { twitterHandle } = this.state;
-    this.setState({
-      activeRoute,
-    });
-    // console.log('TwitterHandleLanding componentWillReceiveProps activeRoute:', activeRoute);
-    if (nextTwitterHandle && twitterHandle.toLowerCase() !== nextTwitterHandle.toLowerCase()) {
-      // We need this test to prevent an infinite loop
-      console.log('TwitterHandleLanding componentWillReceiveProps, different twitterHandle: ', nextProps.match.params.twitter_handle);
+  componentDidUpdate(prevProps) {
```
DaleMcGrew, line 55: It seems like it would be more efficient to only call:
      TwitterActions.resetTwitterHandleLanding();
      TwitterActions.twitterIdentityRetrieve(nextTwitterHandle);
...when nextTwitterHandle changes, as opposed to also making that call when the pathname changes. I recommend splitting those two different conditions into two blocks. (and two different setStates as well)

Anirudhx234: Alright, I'll make those changes

src/js/pages/TwitterHandleLanding.jsx
```diff
@@ -52,36 +52,19 @@ export default class TwitterHandleLanding extends Component {
...
-      nextTwitterHandle = twitter_handle;
-      // eslint-disable-next-line react/prop-types
-    } else if (this.props && this.props.location) {
-      // eslint-disable-next-line react/prop-types
-      const { location: { pathname } } = this.props;
-      activeRoute = pathname;
+  componentDidUpdate(prevProps) {
+    const { match: { params: { twitter_handle: twitterHandle } }, location: { pathname: activeRoute } } = this.props;
+    // if-statements prevents infinite rendering
+    if (prevProps.location.pathname !== this.props.location.pathname) {
```
DaleMcGrew: I seem to remember having a problem with some situations where .location wasn't present, so trying to access an element of an empty variable throws an error. I recommend creating a "prevPathname" and "currentPathname" variable that will exist whether or not "prevProps.location" or "this.props.location" exist, before you do this comparison:
if (prevProps.location.pathname !== this.props.location.pathname)

This same comment applies to the next comparison:
if (prevProps.match.params.twitter_handle !== this.props.match.params.twitter_handle)

src/js/pages/TwitterHandleLanding.jsx
```diff
@@ -52,36 +52,19 @@ export default class TwitterHandleLanding extends Component {
...
-    // console.log('TwitterHandleLanding componentWillReceiveProps activeRoute:', activeRoute);
-    if (nextTwitterHandle && twitterHandle.toLowerCase() !== nextTwitterHandle.toLowerCase()) {
-      // We need this test to prevent an infinite loop
-      console.log('TwitterHandleLanding componentWillReceiveProps, different twitterHandle: ', nextProps.match.params.twitter_handle);
-      TwitterActions.resetTwitterHandleLanding();
-      TwitterActions.twitterIdentityRetrieve(nextTwitterHandle);
-      this.setState({
-        twitterHandle: nextTwitterHandle,
-      });
+    if (prevProps.match.params.twitter_handle !== this.props.match.params.twitter_handle) {
```
DaleMcGrew: Please make sure both versions of the twitter_handle are compared in lower case.

src/js/pages/TwitterHandleLanding.jsx
```diff
@@ -22,7 +22,7 @@ const DelayedLoad = React.lazy(() => import(/* webpackChunkName: 'DelayedLoad' *
 
 // The component that gets called for paths like https://localhost:3000/sierraclub
 // A Twitter handle is the username that appears at the end of your unique Twitter URL
-export default class TwitterHandleLanding extends Component {
+export default class  TwitterHandleLanding extends Component {
```
DaleMcGrew, line 25: This can be done in a future pull request, but I notice there is an extra space before "TwitterHandleLanding" which should be removed.

DaleMcGrew:
Great work! Approved. 👍

PR #3669, opened 2023-06-06
Renamed and reorganized files, so that the role of each file is easily understood at a glance. Updated tests.

nyette:
### Changes

Renamed and reorganized files, so that the role of each file is easily understood at a glance. Updated tests.

### Notes

GA4 requires IDs to be placed directly on the elements that users interact with, so IDs may need to be moved down from wrapper nodes to any descending text nodes.


DaleMcGrew:
👍

.gitignore
```diff
@@ -179,6 +179,6 @@ web_app/build/*
 server.csr
 tests/browserstack/wdio.conf.js
 tests/browserstack/wdio.conf.template
-tests/browserstack*/conf/browserstack.conf.js
-tests/browserstack*/conf/capabilities.json
+tests/browserstack*/config/browserstack.config.js
```
DaleMcGrew, line 182: Hi @Nyette in the WebApp installation process, we tell engineers how to set up an engineer-specific custom version of this config file. Could you please hunt that down and update it to match the new name? Thank you!

PR #3672, opened 2023-06-07
postMessage to enclosing iFrame in Chrome Extension on successfully saving an edit of a  candidate voterGuidePossibilityPosition

SailingSteve:
Allows the dialog containing the iFrame to automatically close on successful save


DaleMcGrew:
👍  Thank you @SailingSteve. Am I correct that you only need this on quality.wevote.us for testing at this time?

PR #3676, opened 2023-06-12
Added dynamic button text during deletion to reassure the user about the possible delay

nyette:
### Changes

Added dynamic button text during deletion to reassure the user about the possible delay


DaleMcGrew:
👍

PR #3685, opened 2023-07-03
Added ability to bypass 2FA via cid query parameter

nyette:
### Changes

Added ability to bypass 2FA via `cid` query parameter

### Related Issue

[WV-82](https://wevoteusa.atlassian.net/browse/WV-82)

### References

[react-router getting this.props.location in child components](https://stackoverflow.com/questions/37516919/react-router-getting-this-props-location-in-child-components)

### DIFF
--- a/src/App.jsx
+++ b/src/App.jsx
@@ -16,6 +16,7 @@ import { renderLog } from './js/common/utils/logging';
 import Header from './js/components/Navigation/Header';
 import HeaderBarSuspense from './js/components/Navigation/HeaderBarSuspense';
 import webAppConfig from './js/config';
+import VoterSessionActions from './js/actions/VoterSessionActions';
 import VoterActions from './js/actions/VoterActions';
 import AppObservableStore, { messageService } from './js/common/stores/AppObservableStore';
 import VoterStore from './js/stores/VoterStore';
@@ -309,7 +310,7 @@ class App extends Component {
     const cid = query.get('cid');
     const voterDeviceId = VoterStore.voterDeviceId();
     if (cid && cid !== voterDeviceId) {
-      VoterStore.setVoterDeviceIdCookie(cid);
+      VoterSessionActions.setVoterDeviceIdCookie(cid);
       VoterActions.voterRetrieve();
     }
   }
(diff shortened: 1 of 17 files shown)


src/App.jsx
```diff
@@ -283,6 +303,17 @@ class App extends Component {
     this.setState({ showReadyLight: false });
   }
 
+  bypass2FA () {
+    const queryString = this.props.location.search;
+    const query = new URLSearchParams(queryString);
+    const cid = query.get('cid');
+    const voterDeviceId = VoterStore.voterDeviceId();
+    if (cid && cid !== voterDeviceId) {
+      VoterStore.setVoterDeviceIdCookie(cid);
```
DaleMcGrew: I think we want to call VoterSessionActions.setVoterDeviceIdCookie instead of VoterStore.setVoterDeviceIdCookie.
EDIT: I can see that the VoterStore.setVoterDeviceIdCookie has the benefit of removing earlier cookies, but it always makes me nervous to set variables directly in the Store because then the change doesn't get picked up by listening components. I'm also wonder if there will be a race condition with the next line "VoterActions.voterRetrieve()". That is, will VoterActions.voterRetrieve() ever try to run *before* the cookies get set?

src/App.jsx
```diff
@@ -4,19 +4,21 @@ import React, { Component, Suspense } from 'react';
...
 import TagManager from 'react-gtm-module';
-import { Redirect, Route, Switch } from 'react-router-dom';
+import PropTypes from 'prop-types';
+import { Redirect, Route, Switch, withRouter } from 'react-router-dom';
 import styled from 'styled-components';
 import muiTheme from './js/common/components/Style/muiTheme';
-import ErrorBoundary from './js/common/components/Widgets/ErrorBoundary';
 import LoadingWheelComp from './js/common/components/Widgets/LoadingWheelComp';
-import WeVoteRouter from './js/common/components/Widgets/WeVoteRouter';
+import historyPush from './js/common/utils/historyPush';
```
DaleMcGrew, line 14: @SailingSteve can you think of any reason moving "WeVoteRouter" to the src/index.jsx would cause problems, either in WebApp or Cordova?

DaleMcGrew:
Hi @SailingSteve this is a checkin from @Nyette as a solution to this issue: https://wevoteusa.atlassian.net/browse/WV-82
I think it looks good to me, but I wanted to request your opinion since it could affect the whole bundle.
Thank you @Nyette for this pull request!

PR #3687, opened 2023-07-05
[WV-89] Added code to make candidate lists and representative lists scroll on button click

rileychou:
[WV-89] The arrow(s) is not clickable (Candidates Page)

Clickable Left and Right arrow buttons that scroll 2 cards at a time. Left arrow hidden when you can't scroll left.


DaleMcGrew:
Thank you @rileychou! Great work. 👍

PR #3689, opened 2023-07-05
In candidate-for-extension don't overwrite good data from url parmas if info is not found by candidateWeVoteId

SailingSteve:
Don't send the placeholder word 'null' to the screen as the Endorsement text, if endorsement text is undefined.


SailingSteve:
No rush on getting this to quality

DaleMcGrew:
👍

PR #3690, opened 2023-07-06
Added ability to bypass 2FA via cid query parameter

nyette:
### Related Issue

[WV-82](https://wevoteusa.atlassian.net/jira/software/projects/WV/issues/WV-82)

### DIFF
--- a/src/App.jsx
+++ b/src/App.jsx
@@ -16,6 +16,7 @@ import { renderLog } from './js/common/utils/logging';
 import Header from './js/components/Navigation/Header';
 import HeaderBarSuspense from './js/components/Navigation/HeaderBarSuspense';
 import webAppConfig from './js/config';
+import VoterSessionActions from './js/actions/VoterSessionActions';
 import VoterActions from './js/actions/VoterActions';
 import AppObservableStore, { messageService } from './js/common/stores/AppObservableStore';
 import VoterStore from './js/stores/VoterStore';
@@ -309,7 +310,7 @@ class App extends Component {
     const cid = query.get('cid');
     const voterDeviceId = VoterStore.voterDeviceId();
     if (cid && cid !== voterDeviceId) {
-      VoterStore.setVoterDeviceIdCookie(cid);
+      VoterSessionActions.setVoterDeviceIdCookie(cid);
       VoterActions.voterRetrieve();
     }
   }
--- a/src/App.jsx
+++ b/src/App.jsx
@@ -4,19 +4,21 @@ import React, { Component, Suspense } from 'react';
 import FullStory from 'react-fullstory';
 import ReactGA from 'react-ga4';
 import TagManager from 'react-gtm-module';
-import { Redirect, Route, Switch } from 'react-router-dom';
+import PropTypes from 'prop-types';
+import { Redirect, Route, Switch, withRouter } from 'react-router-dom';
 import styled from 'styled-components';
 import muiTheme from './js/common/components/Style/muiTheme';
-import ErrorBoundary from './js/common/components/Widgets/ErrorBoundary';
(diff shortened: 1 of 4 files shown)


src/App.jsx
```diff
@@ -4,19 +4,22 @@ import React, { Component, Suspense } from 'react';
...
 import TagManager from 'react-gtm-module';
-import { Redirect, Route, Switch } from 'react-router-dom';
+import PropTypes from 'prop-types';
+import { Redirect, Route, Switch, withRouter } from 'react-router-dom';
 import styled from 'styled-components';
 import muiTheme from './js/common/components/Style/muiTheme';
-import ErrorBoundary from './js/common/components/Widgets/ErrorBoundary';
 import LoadingWheelComp from './js/common/components/Widgets/LoadingWheelComp';
-import WeVoteRouter from './js/common/components/Widgets/WeVoteRouter';
+import historyPush from './js/common/utils/historyPush';
```
DaleMcGrew, line 12: @SailingSteve can you think of any reason moving "WeVoteRouter" to the src/index.jsx would cause problems, either in WebApp or Cordova?

src/index.jsx
```diff
@@ -33,7 +35,11 @@ function redirectToStandardizedWeVoteUrl () {
 
 function startReact () {
   ReactDOM.render(
-    <App />,
+    <ErrorBoundary>
```
DaleMcGrew, line 38: Hi @SailingSteve this is a checkin from @Nyette as a solution to this issue: https://wevoteusa.atlassian.net/browse/WV-82
I think it looks good to me, but I wanted to request your opinion since it could affect the whole bundle. Thank you @Nyette for this pull request!

src/App.jsx
```diff
@@ -283,6 +304,17 @@ class App extends Component {
     this.setState({ showReadyLight: false });
   }
 
+  bypass2FA () {
+    const queryString = this.props.location.search;
+    const query = new URLSearchParams(queryString);
+    const cid = query.get('cid');
+    const voterDeviceId = VoterStore.voterDeviceId();
+    if (cid && cid !== voterDeviceId) {
+      VoterSessionActions.setVoterDeviceIdCookie(cid);
```
DaleMcGrew, line 313: @SailingSteve, do you think it is better to use `VoterSessionActions.setVoterDeviceIdCookie` here, or `VoterStore.setVoterDeviceIdCookie`? I'm always nervous about setting variables directly in the Store because then the change doesn't get picked up by listening components, but if we go through VoterSessionActions, and then on the next line call "VoterActions.voterRetrieve()", will there be a race condition such that VoterActions.voterRetrieve() runs before the voter_device_id cookie gets set?

DaleMcGrew (commented):
Hi @SailingSteve, this pull request will affect Cordova, and I wanted to get your thoughts before approving it.

src/index.jsx
SailingSteve: This should be fine

src/js/common/components/Widgets/ErrorBoundary.jsx
SailingSteve, line 1: Don't know without debugging.  If this works, then fine!

src/App.jsx
SailingSteve: I think this change should be fine

src/App.jsx
SailingSteve: I think in this case, it is fine to set the variable directly in the store. Presumably the bypass will happen before the detailed renders, and if it breaks something it will only be in a test script.

SailingSteve (approved):
I think this is going to be fine.  I'm good with it.

DaleMcGrew:
Thank you @SailingSteve and @Nyette! 👍

PR #3707, opened 2023-07-20
Wv86 riley

rileychou:
WV-86: scrolling container should not have vertical motion

I made the container not wiggle on mobile. I think it works, but will have to test it on an actual phone. 
Also it's telling me that the changes I made for WV-104 are also in this PR, even though I only pushed ScrollingStyles.jsx for this commit.


DaleMcGrew:
👍

DaleMcGrew:
Thank you @rileychou, your latest code is now live on the quality site. It seems like your fix worked, but I'll let you also confirm.

PR #3712, opened 2023-07-25
Updated OpenReplay tracker, so that it only sets userID once

nyette:
### Changes

Updated OpenReplay tracker, so that it only sets `userID` once

### DIFF
--- a/src/App.jsx
+++ b/src/App.jsx
@@ -268,10 +268,7 @@ class App extends Component {
               resourceBaseHref: 'https://wevote.us/',
             });
           }
-          const voterWeVoteId = VoterStore.getVoterWeVoteId();
-          tracker.start({
-            userID: voterWeVoteId,
-          });
+          tracker.start();
           AppObservableStore.setOpenReplayTracker(tracker);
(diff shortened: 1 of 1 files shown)


src/App.jsx
```diff
@@ -268,10 +268,7 @@ class App extends Component {
               resourceBaseHref: 'https://wevote.us/',
             });
           }
-          const voterWeVoteId = VoterStore.getVoterWeVoteId();
-          tracker.start({
-            userID: voterWeVoteId,
-          });
+          tracker.start();
```
DaleMcGrew, line 271: Hi @Nyette, I don't see anywhere else in the code where the userId is set. Are you sure we want to remove this?

nyette: `userID` is being set twice, unless I misunderstood. See [this line](https://github.com/wevote/WebApp/blob/develop/src/js/stores/VoterStore.js#L1158).

DaleMcGrew:
Hi @Nyette, you are correct that a fix was needed. Thank you! I implemented another solution, which I will check in later, so you can roll back this change.

PR #3718, opened 2023-08-01
Wv89 riley

rileychou:
WV-89

Stuff that works:
- The scrolling function speed and animation is much smoother
- On arrow click, card stops before sliding into shaded area
- Clicking arrow centers the cards instead of remaining off centered
- On initial load, arrows appear for “current representatives” and “More politicians”
- When I switch to a different US state with the drop down menu, the arrows are shown/hidden correctly
- **Mobile:** I made the disabled arrows opaque instead of removing it so it doesn’t shift over. 
- CampaignsListRoot includes scrollable arrows now.

What broke or is still broken:
- Clicking “Show more” does not scroll the cards with the new scrolling function
- The arrow states arent being shown/hidden correctly after Using the search bar to filter

### DIFF
--- a/src/js/common/components/Campaign/CampaignCardList.jsx
+++ b/src/js/common/components/Campaign/CampaignCardList.jsx
@@ -104,6 +104,15 @@ class CampaignCardList extends Component {
     });
   }
 
+  loadMoreHasBeenClicked = () => {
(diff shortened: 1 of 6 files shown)


DaleMcGrew:
👍  Thank you @rileychou, I will check it out!

PR #3720, opened 2023-08-01
removed stylelint -- we havent been using it in years, and it was causing a security warning

SailingSteve:
Also ran main.css through css-optimize to reduce its size

resolves https://wevoteusa.atlassian.net/browse/WV-112


DaleMcGrew:
👍  Thank you @SailingSteve!

PR #3721, opened 2023-08-03
Wv89 riley

rileychou:
WV-89

- "func is not a function" fixed. that was an issue of me not merging the files correctly
- As far as I know, the arrow states are being set correctly when the list changes
- temporary workaround for "show more" button click function until i figure out how to make it scroll as intended.


DaleMcGrew:
Thank you @rileychou! 👍

PR #3722, opened 2023-08-04
Performance improvements for the new candidate page   @SailingSteve

SailingSteve:
For candidate Adam Schiff, Nancy Pelosi's endorsement on the right was the last visible element to be rendered and was appearing at about 4.5 seconds
After these changes the page is fully rendered and Nancy appears (to the right and below) at about 1.1 seconds.  In the more typical situation where you don't start the app on this page, it loads in about 950ms.
The https://github.com/wevote/WeVoteServer/pull/2191 siteConfigurationRetrieve improvement knocks about a half second off the page load time.
Separately we were calling retrievePositionListsForOneCandidate for six elections for Adam Schiff (a waste of time) now we load one... either a future election, or the most recent one in the past.
Did not remove all the `<DelayedLoad />` in PoliticsDetailsPage.jsx, just set them conditionally to zero.  This knocked at least a second off the load, and since the rest of the page was loading so quickly, these delays seemed to have little value.
Unfortunately Lighthouse does not give reliable results for this page.  The "Performance" and LDP/FCP numbers are wrong probably due to the facebook graph API at 2.4 seconds and the activityNoticeListRetrieve activityNoticeListRetrieve at 4 seconds.  Other timers expiring later (without doing anything) might also be confusing Lighthouse.
Without Facebook and activityNotice, and monitoring the "Performance" tab, the screen is completely rendered at 950ms and no frame updates occur beyond 950ms.

### DIFF
--- a/src/js/common/components/Campaign/CampaignCardForList.jsx
+++ b/src/js/common/components/Campaign/CampaignCardForList.jsx
@@ -94,9 +94,6 @@ class CampaignCardForList extends Component {
     this.appStateSubscription.unsubscribe();
     this.campaignStoreListener.remove();
     this.campaignSupporterStoreListener.remove();
-    if (this.timer) {
-      clearTimeout(this.timer);
-    }
   }
 
   onAppObservableStoreChange () {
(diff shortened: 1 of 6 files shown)


DaleMcGrew:
👍

Issue #115, opened 2016-02-29
Return key should submit, after entering address


DaleMcGrew:
1. On 'Ballot Location' page, enter an address
2. Hit the "Return/Enter" key
3. The form should submit (currently you must click the "Save Location" button with your mouse to submit)

lisamcho:
Fixed with 519f853

pertrai1:
:100:

PR #3728, opened 2023-08-09
Fixed HowItWorksModal slide change isssue for new viewers

Anirudhx234:
JIRA Issue #115

Commented line 211 in WebApp/src/js/pages/HowItWorks.jsx which seems to be causing users who visit the How It Works modal for the first time to experience a glitch when moving from the first slide to the second slide.


DaleMcGrew:
👍

PR #3746, opened 2023-09-18
Moved the Donorbox script to the head of the Donate page via Helmet

nyette:
### Changes

Moved the Donorbox script to the head of the Donate page via Helmet


DaleMcGrew:
👍

PR #3748, opened 2023-09-22
Support iPhone 15 for Cordova

SailingSteve:
Some work from before my break to reduce storms of apiCalming in position lists.  (Was making debug very difficult).
No longer attempt to calculate the height of the footer bar in WebApp, was throwing exceptions, still works in Cordova, was not needed in WebApp.
Turned off Donate and Squads in footer for Cordova.
Cordova can't handle the Suspense keyword in comments, so renamed them to CommentedOutSupense instead of deleting them in case these bits of commented out code are ever revived.
Messaging is still temporarily disabled -- has problems

### DIFF
--- a/.npmrc
+++ b/.npmrc
@@ -1,0 +1,2 @@
+init-author-name="We Vote Dev Team"
+init-author-email="info@wevote.us"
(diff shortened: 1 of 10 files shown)


DaleMcGrew:
👍

PR #3751, opened 2023-10-07
Sending of emails running again in Cordova iOS

SailingSteve:
Lots of plugins updated for iOS 17
Bring the candidate name down to the sending of email title, so it is no longer simply "Vote for"


DaleMcGrew:
👍

PR #3762, opened 2023-10-27
add optional chaining operator to organization and guide information …

andykb9b13:
…to prevent string  methods from trying to execute with a null value

This is fixing issue WV-158 "The Search bar on "All Endorses" for any Topic rejects user queries"

The only change was including the optional chaining operator to lines 214-216 and 223-225. The toLowerCase() string method was trying to operate on the details for 'organization' and 'guide' before that data was available. The optional chaining operator prevented it from running before there was a value present. 

### Additional Notes
I would also add that in my local environment, not "every" topic created this bug behavior. The only two I could find at a quick glance that did it for me were the "Pro Choice" topic and "2nd Amendment Gun Rights" topics. Many other topics did not display the error behavior for me/

### DIFF
--- a/src/js/pages/Values/OneValue.jsx
+++ b/src/js/pages/Values/OneValue.jsx
@@ -211,18 +211,18 @@ class OneValue extends Component {
         });
         organizationsForValue = filter(organizationsForValueModified,
           (organization) => (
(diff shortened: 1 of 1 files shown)


DaleMcGrew:
👍  Thank you @andykb9b13!

PR #3765, opened 2023-11-02
Fixed a bug that prevents client-side data from clearing after the user requests data deletion

nyette:
### Related Issue(s)

[WV-96](https://wevoteusa.atlassian.net/browse/WV-96)


DaleMcGrew:
👍  Thank you @nyette!

PR #3771, opened 2023-11-07
replace '.replace' with '.replaceAll' for WV-129

cbarnes0:

- WV-129

- changed 4 occurrences of replace with replaceAll. 
- WebApp\src\js\pages\Campaigns\CampaignHomeLoader.jsx 
- WebApp\src\js\pages\Campaigns\CampaignHome.jsx.

These changes allows District of Columbia to load on the first click.


cbarnes0:
There is one additional place where .replace was converted to .replaceAll for safe measure in each file.

DaleMcGrew:
Thank you @cbarnes0 and Christina! 👍

PR #3772, opened 2023-11-08
More Cordova offset adjustments for iOS

SailingSteve:
Resolves https://wevoteusa.atlassian.net/browse/WV-155 which is not in V2.4.2
Fixes an iPad headroom offset which has no jira, and the fix is in V2.4.2


DaleMcGrew:
👍

PR #3788, opened 2023-12-01
WV-136: Ensuring we only see the Displaying '' of '' if we have 6 or more endorsements.

sunnykarim:
WV-136
Ensuring we only see the Displaying '' of '' if we have 6 or more endorsements.


DaleMcGrew:
Thank you @sunnykarim -- great fix. As we discussed, I can approve this when you remove the package-lock.json file from this Pull Request.

PR #3791, opened 2023-12-05
Do not display tooltips in mobile or Cordova, for Jira issue 174

SailingSteve:
Resolves https://wevoteusa.atlassian.net/browse/WV-174
Removes sign in with Facebook from Cordova Android (temporary).

### DIFF
--- a/src/js/common/components/Widgets/ToolBar.jsx
+++ b/src/js/common/components/Widgets/ToolBar.jsx
@@ -4,6 +4,7 @@ import { styled as muiStyled } from '@mui/styles';
 import withStyles from '@mui/styles/withStyles';
 import React, { Suspense } from 'react';
 import styled from 'styled-components';
+import isMobileScreenSize from '../../utils/isMobileScreenSize';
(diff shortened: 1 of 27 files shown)


src/js/common/components/Widgets/ToolBar.jsx
```diff
@@ -24,7 +25,7 @@ function ToolBar (params) {
             className="u-no-underline"
             url="https://twitter.com/WeVote"
             target="_blank"
-            body={(
+            body={isMobileScreenSize() ? (<span />) : (
```
DaleMcGrew, line 28: @SailingSteve Instead of just showing a span in mobile screensize, could we show the IconButton block without the Tooltip wrapper? (for all of the places in this file where you set isMobileScreenSize to remove the Tooltip) I could be mistaken, but it looks like your change removes the icon in mobile view as well as the Tooltip.

DaleMcGrew:
👍  Thank you @SailingSteve!

PR #3796, opened 2023-12-12
[WV-166] Fix

cbarnes0:
[WV-166]

A new function is added to addressFunctions.js made in collaboration with Dale. A new set of conditional operations was added to CampaignsHomeLoader.jsx. With each of these in conjunction, if a stateCode isn't recognized (for example, someone is out of the country) they will then be routed to /cs instead of -candidates/cs. There is not longer an issue with the header not appearing and the expected behavior is happening.

### DIFF
--- a/src/js/common/utils/addressFunctions.js
+++ b/src/js/common/utils/addressFunctions.js
@@ -88,6 +88,18 @@ export function convertStateCodeFilterToStateCode (stateCodeFilter) {
   return '';
 }
 
+export function isValidStateCode (incomingStateCode) {
+  if (incomingStateCode) {
+    // Log incoming State Code
+    console.log('incomingStateCode:', incomingStateCode);
+    const incomingStateCodeUpper = incomingStateCode.toUpperCase();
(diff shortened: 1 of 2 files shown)


src/js/common/utils/addressFunctions.js
```diff
@@ -88,6 +88,18 @@ export function convertStateCodeFilterToStateCode (stateCodeFilter) {
   return '';
 }
 
+export function isValidStateCode (incomingStateCode) {
+  if (incomingStateCode) {
+    // Log incoming State Code
+    console.log('incomingStateCode:', incomingStateCode);
```
DaleMcGrew, line 94: Please comment out console.log's like this in future commits.

DaleMcGrew (approved):
Thank you @cbarnes0! 👍

src/js/common/utils/addressFunctions.js
cbarnes0: Will do! Sorry about that. I thought I got them all!

PR #3813, opened 2024-01-04
Updated package.json to eliminate all of the deprecated libraries

SailingSteve:
Updated included libraries as needed to eliminate deprecated dependent libraries.
Updated react-router-dom from v5 to v6 using the compatibility layer `<CompatRouter>` to eliminate the deprecated package `mini-create-react-context@0.4.1` that was being included in V5.
Fully updating all the `<Route>` commands is a bigger task, that I'll start on.  Then `<CompatRouter>` can be removed.
Did not test the upgrade to the latest version of puppeteer or mocha

Resolves https://wevoteusa.atlassian.net/jira/software/projects/WV/issues/WV-191
except for react-router-dom@6


DaleMcGrew:
👍 Thank you @SailingSteve!

PR #3815, opened 2024-01-13
Changes for Cordova issue fixes

SailingSteve:
Un-copied styles from SquadsPageCommonStyles.jsx.  
In Cordova the developer has to go backwards from the styles in the DOM to the code, so it is very important that every styled component has a unique name.  And the fewer components the better, so please reuse instead of copying. 
Fixes https://wevoteusa.atlassian.net/browse/WV-77 
Fixes https://wevoteusa.atlassian.net/jira/software/projects/WV/issues/WV-47

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9660,9 +9660,9 @@
       }
     },
     "node_modules/axios": {
-      "version": "1.6.4",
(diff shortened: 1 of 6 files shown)


DaleMcGrew:
👍

PR #3828, opened 2024-02-14
Added building and access via webapp to the Storybook design inventory

SailingSteve:
There is no detail in this issue, on how this is going to be used, so I'm shooting in the dark a bit.

`npm start` is for development, but I added the capability there anyways.  `npm build` and `npm prod`, are (I think) what the production servers use, so I also added the capability to those scripts.

It looks like anyone who navigates to https://wevote.us/storybook will be able to modify the statically built copy of the inventory, but those changes do not seem to effect the master copy (wherever that is stored), since when you regenerate the static files, those local changes are overwritten.  The downside is that those local changes seem to persist between sessions (between different users), so accidental changes could be confusing, and purposeful changes would be lost.

I think there is no harm in making this live, but it will probably need more analysis re: what we are using it for, and how it will be used.  And do we need to restrict access?

Implements https://wevoteusa.atlassian.net/browse/WV-206


DaleMcGrew:
👍

SailingSteve:
The local storage was on the url. as passed parameters, so no problem with local changes effecting the actual setup.

PR #3832, opened 2024-02-16
Added building and access via webapp to the Storybook design inventory (Part 2)

SailingSteve:
Now do the redirect in App.jsx.
New https requests will be intercepted before they get to the React Router, and the DOM window.location.href and if pointed at /storybook will be reset to the desired /storybook/index.html

Works fine locally (as did the prior approach), we will see how it does on the production servers.

Implements https://wevoteusa.atlassian.net/browse/WV-206


DaleMcGrew:
👍

PR #3836, opened 2024-02-20
Updated webpack to use wevotedeveloper.com SSL if configured

SailingSteve:
Works on Safari and Firefox for webapp, but not on Chrome. 
Works on all three for local Python API server (which uses a different local http server program).
Works with Chrome and and the Express webserver ... `node serverWevotedeveloper.js`, but not with Chrome and the Webpack devServer.
Will keep working on chrome.


DaleMcGrew:
👍

PR #3839, opened 2024-02-21
Misc changes to use wevotedeveloper.com SSL

SailingSteve:
Updated packages related to Webpack (to no avail)
Still fails to load in SSL mode in Chrome with the Webpack dev server, but works with Firefox and Safari.
Added serverWevotedeveloper.js to prove usability of Chrome with the express webserver server (this works).


DaleMcGrew:
👍

PR #3841, opened 2024-02-22
[WV-131]  Sign In with Twitter button becomes unresponsive after navigating back MERGE READY

smcvey7:
WV-131 Twitter login button disabled when using back button after redirect in Safari

Simplified variables used when determining styling of button and when it is disabled.
Changed to only disable button in iOS due to issue in Safari browser.
Use reloadApp function to reload page instead of repeating code.


DaleMcGrew:
👍

PR #3845, opened 2024-02-25
Preliminary changes for an Android Cordova release

SailingSteve:
Better detection of Android Tablets, for necessary style substitutions. Had to move isAndroid() into a file that does not include other js files, to avoid circular dependencies.
Added the display of the user agent, since it will be needed to detect Android tablets beyond the few predominant developers that we support with this change:  Samsung, Google Pixel, and Lenovo.

There are still a handfull of showstopper issues, that prevent release.  On to iOS for now. https://docs.google.com/document/d/1pl0qsci2kcqGMY1UsQdn72NyMjsmqxuWussbrq41hs0/edit?usp=sharing

A minor change to the EmptyBallotNotice styled component fixes Stas's follow on comment in https://wevoteusa.atlassian.net/browse/WV-47


DaleMcGrew:
👍

PR #3847, opened 2024-02-28
Disable Sign In with Twitter 

SailingSteve:
This only leaves Sign In With Apple (except in Android), and sign in with SMS and email.


DaleMcGrew:
👍  Thank you @SailingSteve for jumping on this so quickly.

PR #3866, opened 2024-03-19
[WV-105] TEAM REVIEW Added open new window for "i'll chip in" in cordova mobile 

jpena925:
Adds icon and checks if isCordova then opens new window outside of app for donations.


src/js/common/pages/CampaignSupport/CampaignSupportPayToPromote.jsx
```diff
@@ -288,7 +291,31 @@ class CampaignSupportPayToPromote extends Component {
...
+                                  style={{
+                                  }}
+                                />
+                              </Button>
+                            )}
+                          />
+                        </Suspense>
+                      </CampaignSupportMobileButtonPanel>
+                    ) : (
                       <Button
```
DaleMcGrew: Hi @jpena925 I forgot when we met that you also have to wrap the Button in the "else" block with `<CampaignSupportMobileButtonPanel>`. You can also delete the "style" attribute in `<Launch`. Finally, OpenExternalWebsite should be included using Lazy, instead of a direct import.

jpena925: Hi @DaleMcGrew I made the changes here. Let me know if this works or if I should make a new PR. Thanks for the feedback and help!

DaleMcGrew:
👍

PR #3867, opened 2024-03-20
WV-4 Indicate primary email/phone number can't be deleted TEAM REVIEW

smcvey7:
### Changes included this pull request

-Add grayed-out trash can icon to primary email/phone numbers
-Add tooltip popup to trash cans with message indicating primary email/phone cannot be deleted
-Change variable name emailOwnershipIsVerified to smsOwnershipIsVerified in phone component

###Questions to address
-trash icon placement (There was originally some padding in the email component, but not the phone. I just eyeballed it with additional padding to line up with existing trash icons)
-double check that key used for div created in map is ok (I'm pretty sure it is)


DaleMcGrew:
👍

PR #3881, opened 2024-04-08
[WV-320] fixes

ivanko22:
Make changes according to design review


src/js/components/CompleteYourProfile/CompleteYourProfile2024.jsx
```diff
@@ -365,7 +365,7 @@ const HowItWorksContainer = styled('div')`
   alignItems: 'center';
   // height: 168px;
   border-radius: 10px;
-  border: 1px solid ${Colors.lightGrey};
+  border: 1px solid #B0B0B0;
```
DaleMcGrew, line 368: I really like using a Colors object like you did before instead of putting a color directly in the component. Can we go back to that, but choose another standardized color?

DaleMcGrew (approved):
Thank you for these changes @ivanko22! I have requested one change you can add to your next pull request. 👍

PR #3883, opened 2024-04-10
[WV-328] Sign Up: Signed in when clicking the "Try a different email" button MERGE READY

smcvey7:
### Changes
Changes digit 6 input onBlur to use handleBlur function instead of handleDigit6Blur
### Notes
Variable cancelingVerifyModal is only referenced in the handleDigit6Blur function which is no longer being called but was left in for reference.


DaleMcGrew:
👍  Great detective work @smcvey7!

PR #3892, opened 2024-04-23
[WV-321] TEAM REVIEW: styling for notifications with designtokencolors

jpena925:
This adds styling to notifications for when clicked vs not clicked and used some DesignTokenColors

### Potential Issues - 
Marked with comments, line 61 and 62, I get no notification data from ".allActivityNotices" but partial data from "allActivity". Dale took screenshots of the server code to mark what might be going on.

### DIFF
--- a/src/js/common/components/Style/DesignTokenColors.js
+++ b/src/js/common/components/Style/DesignTokenColors.js
@@ -1,9 +1,13 @@
 const primitiveColorNames = {
   blue50: '#E6F3FF',
+  blueUI50: '#E2F2FD',
+  blueUI900: '0943A3',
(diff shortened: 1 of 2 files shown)


src/js/common/components/Style/DesignTokenColors.js
```diff
@@ -1,9 +1,13 @@
 const primitiveColorNames = {
   blue50: '#E6F3FF',
+  blueUI50: '#E2F2FD',
+  blueUI900: '0943A3',
```
DaleMcGrew, line 4: I will fix this, but it should have included "#" at the start. Otherwise approved! 👍

PR #3899, opened 2024-05-04
Copied Django templates included files: css, js, and icons to WebApp for static serving to the API Server

SailingSteve:


DaleMcGrew:
👍

PR #3901, opened 2024-05-04
Copied Django templates included js files:  Needed a webpack.config copy rule

SailingSteve:


DaleMcGrew:
👍 Thank you @SailingSteve!

PR #3923, opened 2024-05-27
Added pauses to test cases and updated discuss page test cases

fdiazq:
Added pauses to slow down videos on BrowserStack, commented out test cases that do not pass


DaleMcGrew:
👍

PR #3933, opened 2024-06-01
Sign in with Twitter fixes for iOS Cordova

SailingSteve:
DelayedLoad and Suspense tags have to be on a single line in the render() markup for Cordova.
Images need to be imported with normalizedImagePath() if they are going to work in Cordova.


DaleMcGrew:
👍

PR #3974, opened 2024-08-09
WV-316 After pressing cancel, redirects user to sign in options [TEAM REVIEW]

omgitsmiles:

WV-316 - 'We should aim to eliminate unnecessary steps and if a user opts to use the 'Cancel' button, they should be redirected back to one step.'

Commented back in logic that pressing cancel redirects user sign in options rather than clearing the number form.

### DIFF
--- a/src/js/components/Settings/VoterPhoneVerificationEntry.jsx
+++ b/src/js/components/Settings/VoterPhoneVerificationEntry.jsx
@@ -302,12 +302,12 @@ class VoterPhoneVerificationEntry extends Component {
     // console.log('cancelShouldCloseModal:', cancelShouldCloseModal);
     if (cancelShouldCloseModal) {
       this.closeSignInModalLocal();
-    } else if (isMobileScreenSize()) {
-      if (this.props.showAllSignInOptions) {
-        this.props.showAllSignInOptions();
-      } 
     } else if (isCordova()) {
-      // WV-316: seperated Cordovoa and Mobile screen size logic, Mobile only shows all sign in options.
+      if (this.props.showAllSignInOptions) {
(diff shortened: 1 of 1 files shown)


DaleMcGrew:
Hi @omgitsmiles could you please separate the "if Cordova" logic from the "if isMobileScreenSize". We can showAllSigninOptions with "isMobileScreenSize" (i.e. Uncommenting the code that is there), but there may be a reason @SailingSteve needed that commented out for cordova. Please add a note in the code mentioning WV-316 so that when @SailingSteve is in the code updating the Cordova version next he can consider uncommenting that code in that Cordova logic branch too.

src/js/components/Settings/VoterPhoneVerificationEntry.jsx
```diff
@@ -302,10 +302,12 @@ class VoterPhoneVerificationEntry extends Component {
...
       this.closeSignInModalLocal();
-    } else if (isCordova() || isMobileScreenSize()) {
-      // if (this.props.showAllSignInOptions) {
-      //   this.props.showAllSignInOptions();
-      // }
+    } else if (isMobileScreenSize()) {
+      if (this.props.showAllSignInOptions) {
+        this.props.showAllSignInOptions();
+      } 
+    } else if (isCordova()) {
```
DaleMcGrew: Hi @omgitsmiles unfortunately "else if" isCordova has to be first so that the Cordova devices that are mobile don't get picked up by the first "else if (isMobileScreenSize)".

omgitsmiles: Understood, will fix.

DaleMcGrew:
Thank you @omgitsmiles! 👍

PR #4011, opened 2024-09-03
Solved WV-409- Audited the privacy policy page for ADA compliance issues and made changes to the code to make it ADA compliant.

Almee98:
- WV-409

- Made the table (scrollable element) reachable through keyboard
- Removed the user-scalable="no" parameter from the content attribute of the <meta name="viewport"> element in order to allow zooming.
- Included `aria-label="profile avatar"` in the profile icon button so that it has a name that is identifiable by screen readers.

### Changes that might be needed:
- The mailing link in the 
   `<p>
          <span>
            If you have any questions about this Privacy Policy, you may contact us by sending an e-mail to
            {' '}
            <a href="mailto:info@WeVote.US" target="_blank" class="class u-link-color" rel="noopener noreferrer">info@WeVote.US</a>
            {' '}
          </span>
          <br />
          <span>You can also write to us at the following address:</span>
  </p>`
Is not recognizable even after changing the colors. The only way it is distinguishable is if there is some text that immediately follows 'info@WeVote.US' link.

### DIFF
--- a/src/js/common/components/ChallengeInviteFriends/ViewInviteeDetails.jsx
+++ b/src/js/common/components/ChallengeInviteFriends/ViewInviteeDetails.jsx
@@ -7,68 +7,73 @@ import React from 'react';
 import DesignTokenColors from '../Style/DesignTokenColors';
 import ModalDisplayTemplateA, { templateAStyles, TextFieldWrapper } from '../../../components/Widgets/ModalDisplayTemplateA';
 
-const ViewInviteeDetails = ({ show, setShow, setAnchorEl }) => {
+const ViewInviteeDetails = ({ inviteeId, show, setShow, setAnchorEl }) => {
+  const handleClose = () => {
+    setShow(false);
+    setAnchorEl(null);
+  };
 
- const handleClose = () => {
-  setShow(false);
-  setAnchorEl(null);
- };
+  const dialogTitleText = "Jane's Invitation History";
 
- const dialogTitleText = "Jane's Invitation History";
-
- const textFieldJSX = (
-  <TableContainer components={Paper} sx={{ paddingBottom: '5px' }}>
-    <TableWrapper>
-     <Table aria-label="simple table">
-      <TableHead>
-        <StyledTableRow>
-          <StyledTableHeaderCell>STATUS</StyledTableHeaderCell>
-          <StyledTableHeaderCell align="right">DATE</StyledTableHeaderCell>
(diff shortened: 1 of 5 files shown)


src/js/common/components/PrivacyBody.jsx
```diff
@@ -84,7 +84,7 @@ export default class PrivacyBody extends Component {
             programs, and to send you timely election reminders.
             You may opt-out of receiving messages from WeVote by sending any email to
             {' '}
-            <a href="mailto:info@WeVote.US" target="_blank" rel="noopener noreferrer">info@WeVote.US</a>
+            <a href="mailto:info@WeVote.US" target="_blank" class="class u-link-color" rel="noopener noreferrer">info@WeVote.US</a>
```
DaleMcGrew: Hi @Almee98 thank you for pushing this issue forward! I notice this is where the extra "class " is that I mentioned in our meeting yesterday. (It can be removed)

src/js/common/components/PrivacyBody.jsx
```diff
@@ -322,8 +322,9 @@ export default class PrivacyBody extends Component {
           <span>
             If you have any questions about this Privacy Policy, you may contact us by sending an e-mail to
             {' '}
-            <a href="mailto:info@WeVote.US" target="_blank" rel="noopener noreferrer">info@WeVote.US</a>
+            <a href="mailto:info@WeVote.US" target="_blank" class="class u-link-color" rel="noopener noreferrer">info@WeVote.US</a>
```
DaleMcGrew: Another extra "class " text.

src/js/components/Navigation/HeaderBackTo.jsx
```diff
@@ -1,325 +1,325 @@
...
-  },
-  stackedReturnAndShare: {
-    flexDirection: 'column',
-    alignItems: 'flex-start',
-  },
-});
-
-export default withStyles(styles)(HeaderBackTo);
-
+import { AccountCircle } from '@mui/icons-material';
```
DaleMcGrew: Hi Almee, could you please remove this file from your pull request? I don't see any changes, but Github is reacting as if you deleted this file and then added it beck in again. I would recommend spending some time researching how to fix this by searching Google. It might be a little bit complicated, but is worth it to learn.

DaleMcGrew (changes requested):

PR #4036, opened 2024-09-17
[MERGE READY] WV-538 - Ballot: Clicking anywhere on candidate which causes hover, should open the drawer (same as clicking on Candidate name)

eaw-pid:
Created new function handleContainerClick to better target elements


DaleMcGrew:
👍

PR #4070, opened 2024-09-30
chore(config): update WebdriverIO local and browserstack configs

asjwbl:
Updated WebdriverIO configuration to run test between local browser and BrowserStack.

Test should be executed on local browser (chrome) instead of BrowserStack WHEN we run test(s) locally using 
`(WebAppEnv) $ npm run wdio-local -- --spec [specs_dir_path]/[spec].js` which make it easier for debugging during creating automated test script.

To run test locally on Browserstack, just run command: `(WebAppEnv) $ npm run wdio-local-bs -- --spec [specs_dir_path]/[spec].js`


DaleMcGrew:
👍  Thank you @Aisanjiang-W!

PR #4071, opened 2024-09-30
[MERGE READY] WV-531 Create List Item for Opinion about candidate

jedwardmook:

https://wevoteusa.atlassian.net/browse/WV-531

-Starts to change over buttons and h3s to styled components but stops because they will be taken care of in thumbsUpThumbsDown widget
-changes layout for popover, uses anchorReference rather than anchorOrigin to place popover above and to the left of button
-adds classes for styling of popover
-button for view source with open in new tab


DaleMcGrew:
Looks great. Thank you! 👍

PR #4072, opened 2024-09-30
[WV-522]Added DesignTokenColors instead of css color

itcreativeusa:
Added DesignTokenColors instead of css color


DaleMcGrew:
👍

PR #4073, opened 2024-10-01
WV-560 Reduced height and adjusted centering of text of TIP chip, WV-417 Added additonal test case top navigation TEAM REVIEW

fdiazq:
Reduced size of chip
Set fixed height size


DaleMcGrew:
Hi @fdiazq, this is the correct height for the Chip, but the "TIP" text should be vertically centered within the chip.
Screenshot 2024-10-01 at 5 33 49 AM

DaleMcGrew:
👍

PR #4084, opened 2024-10-04
WV-416 and WV-544 scripts

shan14T:

WV-416 and WV-544

Add new scripts for  'Who's Running for office" Naviation links in footer section
Verified existing scripts for all tc's  and made changes to run on the current build. One test script having locator issues, so skipping the run for that( tc#2 in testcases document).


shan14T:
will be creating new pull requests for the above

PR #4089, opened 2024-10-06
WV-492 ADA Accessibility - resolve ARIA role issues TEAM REVIEW

tejasaditya:

WV-492 ADA Accessibility - resolve ARIA role issues

The custom TabWithPushHistory functional component was modified use the Tab component directly with the root element being set as Link rather than using a Link component wrapper. This allows the hierarchy of components to be direct and each tab is now a direct child of the flexContainer element which has a role of tablist.


DaleMcGrew:
Great solution. 👍

PR #4091, opened 2024-10-07
[WV-514]rename <img> to <SvgImage>, h2 and button changed to custom components 

itcreativeusa:
change h2 and button to custom style name component

remove css for img since we use svgimage now

### DIFF
--- a/src/js/common/components/Navigation/ChallengeInviteSteps.jsx
+++ b/src/js/common/components/Navigation/ChallengeInviteSteps.jsx
@@ -198,21 +198,6 @@ const StepOneIconAndText = styled('div')`
   margin-right: 25px;
   text-align: center;
   width: 169px;
-
-  a {
-    font-weight: ${({ isActive }) => (isActive ? '600' : 'normal')};
-    color: ${({ isActive }) => (isActive ? DesignTokenColors.primary500 : 'inherit')};
-
-    &:hover {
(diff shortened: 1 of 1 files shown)


src/js/common/components/Navigation/ChallengeInviteSteps.jsx
```diff
@@ -208,10 +204,6 @@ const StepOneIconAndText = styled('div')`
   text-align: center;
   width: 169px;
 
-  img {
-    content: url(${({ isActive }) => (isActive ? StepOneActiveIcon : StepOneIcon)});
-  }
-
   a {
```
DaleMcGrew, line 201: Are these "a {" styles still needed?

DaleMcGrew (approved):
Please run `npm run lint` to make sure the eslinter doesn't have any complaints about formatting. Thank you! 👍

src/js/common/components/Navigation/ChallengeInviteSteps.jsx
itcreativeusa: Yes, we still need <a>  because < Link> renders as an <a> tag in the browser. We use it to directly style the <a> tag nested inside the StepOneIconAndText component. 
 But also, I can try to rewrite it and put those styles directly inside the < Link>. 
Would it be better to rewrite it? Or leave it as is?

PR #4098, opened 2024-10-08
 WV-41 [TEAM REVIEW] Set open state on Unfollow menu 

JosephSzpigiel:
[Unfollow dropdown option shifted to top of screen on Web and Mobile devices. ](https://wevoteusa.atlassian.net/browse/WV-41)

Added open state change when Unfollowing to keep the menu from appearing in the upper right hand corner when re-following.


DaleMcGrew:
👍  We can also review in our next meeting. Thank you @JosephSzpigiel

PR #4101, opened 2024-10-09
[TEAM REVIEW] - WV 538 - Clicking anywhere on candidate which causes hover, should open the drawer

eaw-pid:
Changed pointer to cursor

The only place you can't click is on the BallotMatchIndicator2024 span/button - I changed that from cursor to default pointer


DaleMcGrew:
👍 Thank you @eaw-pid!

PR #4106, opened 2024-10-13
Replace Link with NavLink and style for active state in ChallengeInvi…

itcreativeusa:
…teSteps component. Added StyledNavLink component using styled-components to ensure consistent styling and handle active class.


DaleMcGrew:
👍

PR #4109, opened 2024-10-14
[WV-522] Created switcher for border in Joined and Days left

itcreativeusa:
Fixed design token colors to correct ones from file DesignTokenColors.js, and created a switcher for border joined and days left that turns off and on borderSwitcher={borderSwitcher} and borderSwitcher={False}
Fixed DesignTokensColors in JoinedAndDaysLeft.jsx as previously were undefined, changed to correct ones and created a switcher for border joined and days left

Renamed style prop to borderSwitcher in JoinedAndDaysLeft and ChallengeHeaderSimple components

Applied borderswitcher for ChallengeHeaderSimple.jsx and changed "border: none"  to To  just "none" in a ternary expression that dynamically applies CSS styles based on the value of the borderSwitcher prop.


DaleMcGrew:
👍 Thank you @itcreativeusa.

PR #4110, opened 2024-10-14
[WV-547] Fixed Tab Key navigation for the sign-in page [Team Review]

anup2307:
SignInOptionsPanel.js - added showEmailOnlysignIn() to VoterPhoneVerificationEntry.js as prop

VoterPhoneVerificationEntry.jsx - In OnBlur() and OnCancel() functions added code to set the focus back on the correct field.

VoterEmailAddressEntry.jsx -  In OnBlur() and OnCancel() functions added code to set the focus back on the correct field.


DaleMcGrew:
👍

PR #4117, opened 2024-10-15
Fixes for Cordova Android build 2.5.0.0

SailingSteve:
Still need to get Sign in with Twitter working
Some more removal of stripe calls.
Some snyk fixes
Removed a listener from OrganizationModal.jsx on close. 
"Fixed" the missing address on the ready page for mobile devices and Cordova.

### DIFF
--- a/src/js/common/components/CampaignSupport/PayToPromoteProcess.jsx
+++ b/src/js/common/components/CampaignSupport/PayToPromoteProcess.jsx
@@ -2,35 +2,26 @@ import { LockOutlined } from '@mui/icons-material';
 import { Button, InputAdornment, TextField } from '@mui/material';
 import { styled as muiStyled } from '@mui/styles';
 import withStyles from '@mui/styles/withStyles';
-import { Elements } from '@stripe/react-stripe-js';
-import { loadStripe } from '@stripe/stripe-js';
 import PropTypes from 'prop-types';
 import React, { Component } from 'react';
 import styled from 'styled-components';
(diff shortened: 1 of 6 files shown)


src/js/common/components/CampaignSupport/PayToPromoteProcess.jsx
```diff
@@ -354,58 +345,58 @@ class PayToPromoteProcess extends Component {
             </ContributeGridWrapper>
           </InnerWrapper>
         </OuterWrapper>
-        <PaymentWrapper>
```
DaleMcGrew, line 357: Hi @SailingSteve we actually are using this code for a UX test and can't strip this out at this time. I'd rather leave PayToPromoteProcess untouched this close to the election. I'm happy to jump on a screenshare if you have any questions.

src/js/pages/More/Donate.jsx
```diff
@@ -380,16 +377,16 @@ class Donate extends Component {
                           </ContributeGridSection>
                         </ContributeGridWrapper>
                         <PaymentWrapper joining={joining}>
-                          <PaymentCenteredWrapper>
```
DaleMcGrew, line 383: I'd also like to leave this page unchanged this close to the election. This code shouldn't be reachable on our live site, but there are too many moving parts to make this change right now.

DaleMcGrew (changes requested):

SailingSteve:
Replaced by https://github.com/wevote/WebApp/pull/4157

PR #4126, opened 2024-10-18
[MERGE READY] WV-538 - Ballot- Clicking anywhere on candidate should trigger modal

eaw-pid:
Updated handleContainerClick function to use data attributes on elements instead of unique ids.


DaleMcGrew:
👍 Great work @eaw-pid!

PR #4127, opened 2024-10-19
Updated all instances of "Sign in with Twitter" to "Sign in with X" in SignIn Page TestScripts

lavishaGit:
Fix is in  SignIn page change in TwitterButton text from "Sign in with Twitter" to "Sign in with X"

Updated few locaters on Signin Page for more stable and straightforward
Updated test script with require validation for "Sign in with X" instead of "SignIn Twitter"


tests/browserstack_automation/page_objects/signin.page.js
```diff
@@ -15,10 +15,10 @@ class SignIn extends Page{
     }
 
     get signInWithTwitterTextElement(){
-        return $('//span[text()="Sign in with Twitter"]');
+        return $('.csbvaL');
```
fdiazq: Could you locate the element with an id instead of class name?

fdiazq (changes requested):

tests/browserstack_automation/page_objects/signin.page.js
lavishaGit: Thanks for reviewing the changes, @fdiazq! I tried adding the ID for the "Sign in with Twitter" element, but  the WebApp running on localhost port 3000 is not displaying the Sign in with X/Twitter option on the SignIn page. As a result, I opted for the class name strategy, which is more stable .
Please have a look .

lavishaGit:
@fdiazq @DaleMcGrew  I updated the changes that are requested for SignIn page .Please review it .

fdiazq (approved):
@lavishaGit  👍

PR #4131, opened 2024-10-20
WV-652 Add Challenge started line

Succorro:
Update to Challenge About section
- Captured SVG from Figma and create file for it. 
- Created state for showDaysLeft, challengeCreator, challengeIsSupporting 
- set methods for new state additions are called within onChallengeStoreChange (needs to be updated, has dummy text) 
- Updated challengeDates text to better handle boolean state for showDaysLeft


DaleMcGrew:
Hi @Succorro, thank you for submitting this! If you believe your code won't break anything, please don't submit the pull request as a "Draft" pull request. If you want to signal that you would like to look at the code together, please add "TEAM REVIEW" to the name of the pull request. Do you think this code is ready for me to connect to live data?

Succorro:
Good to know for the future! Yes, the code won't have any breaking changes. The only note I would have for the live data would be the type of values we are receiving for this component from the store. In my implementation, I am assuming the values will be strings for the challenge creator and supporting names. If we need to update anything I'll be available

DaleMcGrew:
👍

PR #4135, opened 2024-10-20
Profile SignedIn testcases automation

Ayushee28:
Automated testcases from Profile SignedIn
Added ids for automation for some of the elements not having ids

@DaleMcGrew @fdiazq @mythrienjam @lavishaGit

### DIFF
--- a/tests/browserstack_automation/specs/SignedIn.js
+++ b/tests/browserstack_automation/specs/SignedIn.js
@@ -4,9 +4,18 @@ import ReadyPage from '../page_objects/ready.page';
 
 const waitTime = 14000;
 
+async function switchWindow(currentWindowHandle, allWindows) {
+  for (const window of allWindows) {
+    if (window !== currentWindowHandle) {
+      await driver.switchToWindow(window);
+      break;
+    }
+  }
+}
+
 describe('Profile(Signed In)', () => {
 
-    ProfileSignedIn_001
+    // ProfileSignedIn_001
     it.only('verifySignInButton', async () => {
       await ReadyPage.load();
       await driver.pause(waitTime);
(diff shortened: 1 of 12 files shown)


tests/browserstack_automation/specs/SignedIn.js
```diff
@@ -0,0 +1,319 @@
...
+      await driver.pause(waitTime);
+      await SignedIn.enterPhoneNumberInPhoneNumberInput();
+      await SignedIn.clickOnSendCodeOption();
+      await driver.pause(waitTime);
+      await SignedIn.enterVerificationCode();
+      await SignedIn.clickOnVerifyButtonForCodeVerification();
+      await driver.pause(waitTime);
+      await SignedIn.clickOnProfilePhotoPlaceholder();
+      await driver.pause(waitTime);
+      await expect(driver).toHaveUrl(expect.stringContaining('profile'));
```
lavishaGit, line 44: @Ayushee28 Just a suggestion correct me if I am wrong but there is no need to use driver—all commands and functionality are attached to the browser object by default.So you can use :
await expect(browser).toHaveUrl(expect.stringContaining('profile'));

Ayushee28: @lavishaGit I am not sure about that. I had tried with browser and it gives error sometimes for me. Only driver works fine. Also I saw others have used driver as well in the other specs so added this way

fdiazq: @Ayushee28 The use of driver is fine

fdiazq:
@Ayushee28 The test cases look good. I recommend adding the new SignedIn.js to the config file

Ayushee28:
> @Ayushee28 The test cases look good. I recommend adding the new SignedIn.js to the config file

Will make sure to add it in next pull request

PR #4138, opened 2024-10-22
[WV-523] Push voter_we_vote_id to GA4 from within WebApp user data [TEAM REVIEW]

StevenSmith-code:
It now attaches the wevoterid to the datalayer using the TagManager datalayer. Did you mean for the GTMId to be the GOOGLE_ADS_TRACKING_ID @DaleMcGrew I made a seperate key value pair in the config but can adjust that if you prefer the former.


DaleMcGrew:
👍

PR #4144, opened 2024-10-24
[WV-522]Updated ChallengeHeaderSimple.jsx with custom padding for JoinedAndDa…

itcreativeusa:
Added props for **border** to JoinedAndDaysLeft component and updated ChallengeHeaderSimple.jsx.


itcreativeusa:
Also, I can add a prop for margin for JoinedAndDaysLeft in ChallengeHeaderSimple if we need to change the margin.

DaleMcGrew:
👍  Thank you @itcreativeusa.

PR #4158, opened 2024-10-31
[WV-691]Updated toolbar layout Moved MoreInfoIcon to the right with absolute positioning. 

itcreativeusa:
Moved MoreInfoIcon to the right with absolute positioning. Removed previously added **margin** from styles as this is a new approach.
Styles added: 
**Toolbar** style={{ display: 'flex', alignItems: 'center', position: 'relative', width: '100%' }}
**Tabs:** style={{ flexGrow: 1 }}
**MoreInfoIcon:**  style={{ position: 'absolute', right: '16px', display: 'flex', alignItems: 'center' }}


DaleMcGrew:
👍

PR #4185, opened 2024-11-27
WV-706 Ballot search fixes, updated/refactored search function [Team Review]

Succorro:
WV-706 Ballot search function 
- FilterBaseSearch: Refactored filter function to check for search value and filter out candidates that did not match, similar functionality to /CampaignListRoot `onFilterOrListChange` search matching. 
- Ballot/BallotItemCompressed/OfficeItemCompressed: Passed `foundInSearchWords` prop to ensure when item is found the component will properly rerender the new item. Was running into an issue where the initial search candidate would work but when searching for another candidate in the same category (ie. presidential candidate) the second search would not update.


DaleMcGrew:
👍

PR #4187, opened 2024-12-02
[wv-758]Mobile Footer in middle of page in certain situations

itcreativeusa:
[wv-758 ]Set min-height: 100vh to ensure full viewport height for the WeVoteBody styled component Defaulted bottom to '57px' for consistent appearance when no prop is provided.


DaleMcGrew:
Thank you @itcreativeusa! 👍

PR #4205, opened 2024-12-13
WV-490 parse politician name and state, render on before load

omgitsmiles:

Google  “de-lists” the politician page when there is a delay in retrieving politician information. The information does seem to eventually load, but too late for Google. I parsed the politician name and state from URL to render on suspense before api loads. 

- extractPoliticianDetailsFromUrl function pulling politician name and state from URL
- adding a suspense with politician details on load

To Dos:

- [ ] Code cleanup / ES Lint


src/js/compileDate.js.backup
```diff
@@ -0,0 +1 @@
+module.exports = ['10/9/2024, 5:54:14 PM'];
```
DaleMcGrew: Please add this compileDate.js.backup file to .gitignore so it isn't checked in.

src/js/common/pages/Politician/PoliticianDetailsPage.jsx
```diff
@@ -901,6 +942,14 @@ class PoliticianDetailsPage extends Component {
           <meta name="description" content={politicianDescriptionLimited} />
         </Helmet>
         <PageWrapper>
+        <Suspense fallback={(
```
DaleMcGrew: Hi @omgitsmiles, we don't want to add a different place to display this information, we want the politicianNameParsedFromURLBeforeLoad and politicianStateParsedFromURLBeforeLoad placed into the components where the name and state currently appear once the data loads.

DaleMcGrew (changes requested):
This is a good start @omgitsmiles, thank you. Comments included.

src/js/common/pages/Politician/PoliticianDetailsPage.jsx
omgitsmiles: Thanks for the clarification. Will update.

DaleMcGrew:
Thank you @omgitsmiles! 👍

PR #4206, opened 2024-12-13
updated 'whatIsHappeningTitle' id to camel case in CandidatesPage tes…

nadiavaida:
…tscript

updated 'whatIsHappeningTitle' id to camel case in Candidates Page test script. The test case was failing due to mismatch in id name on webApp and the one used in test script.


DaleMcGrew:
👍

PR #4207, opened 2024-12-14
[wv-691] changed breakpoint for invited friends tab

itcreativeusa:
Updated the tab label for small screens (sm) to display a shorter version ("Invited") instead of the full label ("Invited friends").


DaleMcGrew:
👍

PR #4240, opened 2025-01-31
[MERGE READY] - Wv-937 - ADA/Ballot - Keyboard accessibility issues

eaw-pid:
Ballot.jsx 
- Moved the onClick into the Chip component for accessilibty

HeartFavoriteToggleBase.jsx
- Changed LikeContainer and DislikeContainer to styled 'buttons' and added 'hover' behavior so can be identified during tabbing


DaleMcGrew:
👍  Thank you @eaw-pid! @charanya-QA I will be checking in a follow-up pull request to move the id "ballotBadgeMobileAndDesktop-All" to the new onClick element, but our two pull requests might still break WebApp/tests/browserstack/specs/ballotMainTest.js under "should click on ballot filler badges at top of page"

PR #4243, opened 2025-02-05
Search bar 2024 handle key press

seanquinn917:

Captures Data from Challenges page search bar including keyword and user info

Refactoring of handleKeyPress() function within SearchBar2024 to include Google tag to capture the content being searched.  "trackSearch" prop passed from ChallengesHomeFilter (parent component) to localize functionality in generic component.

### DIFF
--- a/src/js/common/components/Challenge/ChallengesHomeFilter.jsx
+++ b/src/js/common/components/Challenge/ChallengesHomeFilter.jsx
@@ -18,6 +18,8 @@ function ChallengesHomeFilter (props) {
           searchFunction={props.searchFunction}
           clearFunction={props.clearSearchFunction}
           searchUpdateDelayTime={500}
+          trackSearch={true}
+          parentComponentName="ChallengesHomeFilter"
         />
       </SearchBarWrapper>
     </ChallengesHomeFilterWrapper>
--- a/src/js/common/components/Search/SearchBar2024.jsx
+++ b/src/js/common/components/Search/SearchBar2024.jsx
@@ -56,12 +56,30 @@ class SearchBar2024 extends Component {
     }
   }
 
-  handleKeyPress () {
-    if (this.timer) clearTimeout(this.timer);
-    this.timer = setTimeout(() => {
-      this.props.searchFunction(this.state.searchString);
-    }, this.props.searchUpdateDelayTime);
+  
+ handleKeyPress = () => {
+  if (this.timer) {
+    clearTimeout(this.timer);
   }
+  this.timer = setTimeout(() => {
+    const { searchString } = this.state;
+    if (searchString.length === 0) {
+      return;
+    }
+    this.props.searchFunction(searchString);
+    if(this.props.trackSearch){
+    const dataLayerObject = {
+      event: 'KeyWord Search',
+      search_keyword: searchString,
+      voterWeVoteId: VoterStore.getVoterWeVoteId(),
+    };
+    //console.log(dataLayerObject)
+    TagManager.dataLayer({dataLayer: dataLayerObject});
+  }
+  }, 3000);
+  const { searchString } = this.state;
+  this.props.searchFunction(searchString);
+};
 
   clearQuery () {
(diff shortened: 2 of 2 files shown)


src/js/common/components/Search/SearchBar2024.jsx
```diff
@@ -56,12 +58,30 @@ class SearchBar2024 extends Component {
...
   }
+  this.timer = setTimeout(() => {
+    const { searchString } = this.state;
+    if (searchString.length === 0) {
+      return;
+    }
+    this.props.searchFunction(searchString);
+    if(this.props.trackSearch){
+    const dataLayerObject = {
+      event: 'KeyWord Search',
```
DaleMcGrew, line 74: I think event needs to be a camelCase single word, like "searchString" instead of "KeyWord Search"

src/js/common/components/Search/SearchBar2024.jsx
```diff
@@ -56,12 +58,30 @@ class SearchBar2024 extends Component {
...
+      search_keyword: searchString,
```
DaleMcGrew, line 75: I believe 'search_keyword' needs to be camelCase like: 'searchKeyword'. Also, I added searchKeyword to our list of Property Names. https://docs.google.com/spreadsheets/d/1KO4_AOmddKvIpA79XrZdaVBdRkcP1HitqrvQkCo2A34/edit?gid=1828322090#gid=1828322090

src/js/common/components/Search/SearchBar2024.jsx
```diff
@@ -56,12 +58,30 @@ class SearchBar2024 extends Component {
...
+      voterWeVoteId: VoterStore.getVoterWeVoteId(),
```
DaleMcGrew, line 76: This should be:
user: {
  voterWeVoteId: VoterStore.getVoterWeVoteId(),
}

src/js/common/components/Challenge/ChallengesHomeFilter.jsx
```diff
@@ -18,6 +18,8 @@ function ChallengesHomeFilter (props) {
           searchFunction={props.searchFunction}
           clearFunction={props.clearSearchFunction}
           searchUpdateDelayTime={500}
+          trackSearch={true}
```
DaleMcGrew, line 21: Did you mean to implement trackSearch in SearchBar2024?

DaleMcGrew (approved):
Hi Sean, congratulations on checking in this pull request! I'm approving it and merging it, but added some requested improvements you can add with your next pull request.

src/js/common/components/Search/SearchBar2024.jsx
seanquinn917: ### Just so I understand, currently "searchString" is the value of searchKeyword.  Would it be helpful to have a name for the event that isn't the same? I know on Tuesday we discussed having a generic term for events, like "Search." Would this be a moment to implement that?  
event: "Search",
searchKeyword: searchString

DaleMcGrew: For event names, they need to be camelCase, so "Search" wouldn't work. I think within Google Tag manager interface we would use a "starts with" regex looking for keywords like "search". Within WebApp I would unique enough event names that we can search the code to find them. So "searchString" or "searchKeyword" instead of having to search for "search" (3.331 results). Also, after more conversations, we can do search-and-replace if we decide on better event names.

seanquinn917: Awesome, understood. thank you Dale!

src/js/common/components/Challenge/ChallengesHomeFilter.jsx
seanquinn917: ### The Google Tag was previously being triggered by every parent component using the SearchBar2024. To resolve this, I've implemented trackSearch on line 72 of SearchBar2024, limiting its scope to this specific instance of the component.

PR #4258, opened 2025-02-19
added scroll func for few tcs wv-579

shan14T:
wv-579
added scroll func for few tcs wv-579 as they seem to fail when running on browser stack

ReadyPage.js


charanya-QA (approved):

charanya-QA:
Thank you!

PR #4265, opened 2025-02-26
WV-986/WBV-987 Add Data Layer To "Show More" buttons on '"Fine Print" and "We Vote helps you" [CHANGES NEEDED]

seanquinn917:
Add Google Tag Manager data layer to two "Show More" instances: "fine print" and "we vote helps you.  
Specific logic changes include passing props from ReadyIntroduction.jsx and ReadyFinePrint.jsx components to create the instances where GTM fires.  Adding "trackInGTM={true}" down to other Show more instances will create data layers in those as well if future needs require.

### DIFF
--- a/src/js/components/Widgets/ShowMoreButtons.jsx
+++ b/src/js/components/Widgets/ShowMoreButtons.jsx
@@ -5,9 +5,45 @@ import PropTypes from 'prop-types';
 import React from 'react';
 import styled from 'styled-components';
 import { renderLog } from '../../common/utils/logging';
+import TagManager from 'react-gtm-module';
+import VoterStore from '../../stores/VoterStore';
 
 
 class ShowMoreButtons extends React.Component {
+  handleClick=()=>{
+    //console.log('click')
+    const { showMoreId, showMoreButtonWasClicked, showMoreCustomText, trackInGTM } = this.props;
+
+    if (trackInGTM && !showMoreButtonWasClicked) {  
+      // console.log("Show More clicked (tracked):", showMoreId);
+      // const eventName = showMoreCustomText || 'show more';
+      const dataLayerObject = {
+        event: 'Show More Button Clicked',
+        element_id: showMoreId,
+        source:{
+          pageName: window.location.href,
+          pageType: "homepage",
+          pathName: window.location.pathname,
+        },
+        user:{
+          weVoteVoterId: VoterStore.getVoterWeVoteId(),
+          userStatus:"",
+          method:""
+        },
+        // button_text: eventName,
+      };
+    //  console.log("dataLayerObject:", dataLayerObject)
+      TagManager.dataLayer({ dataLayer: dataLayerObject });
+    } else if (trackInGTM && showMoreButtonWasClicked){
+       console.log("Show Less clicked (not tracked):", showMoreId);
+    }
+    else {
+      console.log("Show More/Less clicked (not tracked):", showMoreId);
+    }
+
+    this.props.showMoreButtonsLink(); 
+  };
+
   render () {
     renderLog('ShowMoreButtons');  // Set LOG_RENDER_EVENTS to log all renders
     const { classes, showLessCustomText, showMoreButtonsLink, showMoreButtonWasClicked, showMoreCustomText, showMoreId } = this.props;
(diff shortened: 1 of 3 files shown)


src/js/components/Widgets/ShowMoreButtons.jsx
```diff
@@ -5,9 +5,45 @@ import PropTypes from 'prop-types';
...
 class ShowMoreButtons extends React.Component {
+  handleClick=()=>{
+    //console.log('click')
+    const { showMoreId, showMoreButtonWasClicked, showMoreCustomText, trackInGTM } = this.props;
+
+    if (trackInGTM && !showMoreButtonWasClicked) {  
+      // console.log("Show More clicked (tracked):", showMoreId);
+      // const eventName = showMoreCustomText || 'show more';
+      const dataLayerObject = {
+        event: 'Show More Button Clicked',
```
DaleMcGrew, line 21: I believe we want "event" to be a camelCase event name we can listen for in GTM, instead of a sentence.

src/js/components/Widgets/ShowMoreButtons.jsx
```diff
@@ -5,9 +5,45 @@ import PropTypes from 'prop-types';
...
+        element_id: showMoreId,
```
DaleMcGrew, line 22: Let's add "element_id" to the pageDetails dictionary. Please also make it camelCase.

src/js/components/Widgets/ShowMoreButtons.jsx
```diff
@@ -5,9 +5,45 @@ import PropTypes from 'prop-types';
...
+        source:{
```
DaleMcGrew, line 23: Please convert "source" to "pageDetails".

src/js/components/Widgets/ShowMoreButtons.jsx
```diff
@@ -5,9 +5,45 @@ import PropTypes from 'prop-types';
...
+          pageName: window.location.href,
+          pageType: "homepage",
+          pathName: window.location.pathname,
+        },
+        user:{
```
DaleMcGrew, line 28: Please change "user" to "userDetails".

src/js/components/Widgets/ShowMoreButtons.jsx
```diff
@@ -5,9 +5,45 @@ import PropTypes from 'prop-types';
...
+          weVoteVoterId: VoterStore.getVoterWeVoteId(),
+          userStatus:"",
+          method:""
+        },
+        // button_text: eventName,
+      };
+    //  console.log("dataLayerObject:", dataLayerObject)
+      TagManager.dataLayer({ dataLayer: dataLayerObject });
+    } else if (trackInGTM && showMoreButtonWasClicked){
+       console.log("Show Less clicked (not tracked):", showMoreId);
```
DaleMcGrew, line 38: Only indent by 2 spaces instead of 3. I think we *should* track "show less" clicks.

src/js/components/Widgets/ShowMoreButtons.jsx
```diff
@@ -5,9 +5,45 @@ import PropTypes from 'prop-types';
 import React from 'react';
 import styled from 'styled-components';
 import { renderLog } from '../../common/utils/logging';
+import TagManager from 'react-gtm-module';
+import VoterStore from '../../stores/VoterStore';
 
 
 class ShowMoreButtons extends React.Component {
+  handleClick=()=>{
```
DaleMcGrew, line 13: Please change the generic function name `handleClick` to a function name that tells other programmers what this function is doing. For example `sendDataToGTMAndNotifyParent`.

DaleMcGrew (changes requested):
Great work @seanquinn917!

src/js/components/Widgets/ShowMoreButtons.jsx
seanquinn917: to clarify, it should be a string still, but camelCase. Is that correct?  
for example:
event: 'showMoreButtonClicked'

DaleMcGrew: That is correct:
`event: 'showMoreButtonClicked'`

PR #4266, opened 2025-02-26
WV-??? Add Source Object into GTM logic SearchBar2024.jsx [CHANGES NEEDED]

seanquinn917:
Update SearchBar2024.jsx GTM logic to include updated 'source' object.

### DIFF
--- a/src/js/actions/IssueActions.js
+++ b/src/js/actions/IssueActions.js
@@ -5,6 +5,12 @@ import Dispatcher from '../common/dispatcher/Dispatcher';
 // let uniqueKeyIssuesFollowedRetrieve = '';
 
 export default {
+  updateFollowersForIssue (issueWeVoteId, followersChangeAmount) {
+    Dispatcher.dispatch({
+      type: 'updateFollowersForIssue',
+      payload: { issueWeVoteId, followersChangeAmount },
+    });
+  },
   issueDescriptionsRetrieve () {
     /*
     April 2021: We were firing off issueDescriptionsRetrieve four times upon loading the Ready page.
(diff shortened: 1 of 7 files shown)


src/js/common/components/Search/SearchBar2024.jsx
```diff
@@ -73,6 +73,11 @@ class SearchBar2024 extends Component {
     const dataLayerObject = {
       event: 'searchKeyword',
       searchKeyword: searchString,
+      source:{
```
DaleMcGrew: Update to `source` to `pageDetails`.

src/js/common/components/Search/SearchBar2024.jsx
```diff
@@ -73,6 +73,11 @@ class SearchBar2024 extends Component {
...
+        pageName: window.location.href,
+        pageType: "challenge",
+        pathName: window.location.pathname,
+      },
       user:{
```
DaleMcGrew: Update `user` to `userDetails`.

src/js/common/components/Search/SearchBar2024.jsx
```diff
@@ -73,6 +73,11 @@ class SearchBar2024 extends Component {
     const dataLayerObject = {
       event: 'searchKeyword',
       searchKeyword: searchString,
+      source:{
+        pageName: window.location.href,
+        pageType: "challenge",
```
DaleMcGrew: We want the pageName & pageType to come from the parent component that calls `searchBar20204`, but instead of adding props we need to pass from the parent, I think another approach will be more maintainable. Please create a function in `/js/utils/calculateGTMPageType.js` which takes in the pathName, and returns one of our pageTypes: 
ballot, politician, candidate, donation, homepage, ballot, account, challenge, issue
...and then an even more specific pageName. Feel free to reach out with questions!

DaleMcGrew (changes requested):
Thank you @seanquinn917! 👍

PR #4270, opened 2025-03-03
[WV-948] fix followers count cdn

itcreativeusa:
1)Added updateFollowersForIssue action in IssueStore.js
2)Fixed followers count not updating after follow/unfollow
3)State updates correctly without direct mutation
4)Handled followers_count initialization to prevent undefined issues
5)Addressed CDN caching causing outdated data


DaleMcGrew:
👍 Thank you @itcreativeusa!

PR #4280, opened 2025-03-13
WV-937 [TEAM REVIEW] Keyboard accessibility issues: Ballot Page #4,5.

ilonagog:
Keyboard accessibility issues: Ballot Page
Issue 4: Make sure you can reach “Remaining Choices” and “Chosen” (you must choose one candidate on your ballot to see those options)
Issue 5: Click on the election name to find a prior election that has State & Local ballot items. With only “All” and “Federal” you have to tab twice after “All” to get to “Federal” and that may be because those elements are on the page but not visible?


DaleMcGrew:
👍

PR #4288, opened 2025-03-19
Made a larger clickable area for the search magnifying glass for Android in FilterBaseSearch

SailingSteve:
Does not affect WebApp, what left to do should only affect Cordova

Also fixes a "Can't perform a React state update on an unmounted component" in BallotTitleHeader

This is the code in Android 2.6.0

Addresses a number of lint issues in FilterBaseSearch

Resolves https://wevoteusa.atlassian.net/issues/WV-299?filter=-1


DaleMcGrew:
Thank you @SailingSteve! 👍

PR #4296, opened 2025-03-23
Stopped the app from jumping horizontally 7 pixels when search box expands

SailingSteve:
Fix also stops the TopHeader from jumping horizontally when other tabs are selected.

We use a deprecated version of Dialog from '@mui/material', and should replace it with the new version in '@base-ui-components/react/dialog' Added some code to ShareModal, that mostly resolves the problem in the old version. From the doc: "@mui/base has been deprecated and has been replaced by Base UI. We strongly recommend using the new package instead." The change would be... import Dialog from '@base-ui-components/react/dialog' instead of '@mui/material';
Created issue https://wevoteusa.atlassian.net/browse/WV-1044 for this upgrade

Resolves https://wevoteusa.atlassian.net/issues/WV-1028
Resolves https://wevoteusa.atlassian.net/issues/WV-1019

### DIFF
--- a/src/css/main.css
+++ b/src/css/main.css
@@ -60,7 +60,7 @@ body {
   font-weight:normal;
   letter-spacing:.15px;
   line-height: 1.4;
-  margin:0 auto;
+  /* margin:0 auto; Removed 3/19/25 https://wevoteusa.atlassian.net/issues/WV-1028 */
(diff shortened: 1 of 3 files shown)


DaleMcGrew:
👍

PR #4335, opened 2025-05-18
[WV-943] Candidates Page - Candidates under current representatives header is not loading

itcreativeusa:
Previously, the code checked numberOfRepresentativeResults > 0 and displayed a CandidateListRootPlaceholder, even when representativeListShownAsRepresentatives was empty. This could mislead users by suggesting there were representatives when none were actually rendered.

Now, the UI only shows the section if representativeListShownAsRepresentatives has items. This avoids confusing UX and unnecessary rendering.

I added a console.log to compare the actual list length with numberOfRepresentativeResults to help catch any future data mismatches.

Tested with the state of Hawaii, which does have representatives, to confirm the updated logic works as intended.

Screenshot 2025-05-17 at 6 24 58 PM


DaleMcGrew:
Thank you @itcreativeusa 👍

PR #4342, opened 2025-05-26
[WV-1118]-manual-test-comments-not-posting

itcreativeusa:
Fix Avatar child error by destructuring initials and sx props

- Destructure `{ sx, children } = speakerDisplayNameToInitials(name)`
- Pass `sx` to `<Avatar>` and `children` as its content


DaleMcGrew:
Thank you @itcreativeusa this solution looks good! 👍  As I was doing a code review, I noticed that the pattern you are fixing in src/js/components/Activity/ActivityTidbitComments.jsx also still exists in these other areas that should be updated as well:
src/js/common/components/CampaignSupport/MostRecentCampaignSupport.jsx
src/js/components/Activity/ChildCommentList.jsx

Could you please fix (and test to verify) the problem in those places too?

itcreativeusa:
Sure, I'll fix it all!

PR #4358, opened 2025-05-29
WV-706 Ballot: Search by candidate name is also showing other candidates in same race, and should only show candidate in that race matching search [TEAM REVIEW]

jedwardmook:
https://wevoteusa.atlassian.net/browse/WV-706

-uses props being passed from FilterBaseSearch to render list instead list in onCanidateStoreChange to use up to date state


DaleMcGrew:
👍

PR #4401, opened 2025-06-20
[WV-1205]Added destination tracking for dataLayer on Politician Page for the PositionForBallotItem [MERGED]

pi-sen:

### DIFF
--- a/src/js/common/components/Position/PositionForBallotItem.jsx
+++ b/src/js/common/components/Position/PositionForBallotItem.jsx
@@ -1,5 +1,6 @@
 import PropTypes from 'prop-types';
 import React, { Suspense, useState } from 'react';
+import { Link } from 'react-router-dom';
 import styled from 'styled-components';
 import { BlockOutlined, CheckOutlined, Launch, MoreHoriz } from '@mui/icons-material';
 import Popover from '@mui/material/Popover';
@@ -12,8 +13,9 @@ import speakerDisplayNameToInitials from '../../utils/speakerDisplayNameToInitia
 import {
   getDateFromUltimateElectionDate, getTodayAsInteger, timeFromDate,
 } from '../../utils/dateFormat';
-import AppObservableStore from '../../stores/AppObservableStore';
+// import AppObservableStore from '../../stores/AppObservableStore';
 import stringContains from '../../utils/stringContains';
+import lookupPageNameAndPageTypeDict from '../../../utils/lookupPageNameAndPageTypeDict';
 
 const OpenExternalWebSite = React.lazy(() => import(/* webpackChunkName: 'OpenExternalWebSite' */ '../Widgets/OpenExternalWebSite'));
 const ReadMore = React.lazy(() => import(/* webpackChunkName: 'ReadMore' */ '../Widgets/ReadMore'));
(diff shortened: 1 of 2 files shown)


src/js/common/components/Position/PositionForBallotItem.jsx
```diff
@@ -96,6 +96,8 @@ function PositionForBallotItem ({ classes, linksOpenExternalWebsite, position })
             target="_blank"
             trackingOn
             url={speakerLinkExternal}
+            destinationPageName="PoliticianDetailsPage"
+            destinationPageType="politician"
```
DaleMcGrew: Hi @pi-sen the speakerLinkExternal in this context is always going to be a page on WeVote, so instead of passing these destination variables into OpenExternalWebSite, we can update src/js/utils/lookupPageNameAndPageTypeDict.js to deal with the path in speakerLinkExternal. If it is a path pattern that we don't recognize (like https://WeVote.US/npr, we assume it is a Twitter handle. See line 742: src/App.jsx
```
<Route path="/:twitter_handle" exact component={TwitterHandleLanding} />
```
The destinationPageName calculated in lookupPageNameAndPageTypeDict should be: TwitterHandleLanding
destinationPageType: "endorser"

DaleMcGrew (changes requested):

src/js/common/components/Position/PositionForBallotItem.jsx
pi-sen: Changes updated

DaleMcGrew:
Hi @pi-sen we use OpenExternalWebsite when we are opening a new tab, even if the new tab contains the WeVote site. We can't use `Link` because our Cordova mobile app deals with new tabs differently than a typical browser. If you roll back all of your changes in src/js/common/components/Position/PositionForBallotItem.jsx except for these lines, I believe your changes will be correct. 

Please also add a dataLayer within the triple dot menu, when a person clicks "View source of opinion".

pi-sen:
Changes done.

src/js/utils/lookupPageNameAndPageTypeDict.js
```diff
@@ -89,7 +89,10 @@ function calculatePageNameAndPageTypeDict (path) {
   let settingsPageName = 'notSet'; // Per our naming convention for pageName, this would normally be 'NotSet' but I think the value of having settingsPageName being identical to settingsPageType will save us grief in the future.
   let settingsPageType = 'notSet';
 
-  if (path.startsWith('/ballot')) {
+  if (path && (path.includes('http') || path.includes('https')) && !path.includes('wevote.us')) {
```
SailingSteve: For the URL `https://wevote.us/steve-garvey-politician-from-california/-/` the path is `/steve-garvey-politician-from-california/-/`

lookupPageNameAndPageTypeDictionary should NOT receive URLs, it should only receive paths  so if the path starts with `https://` or `http://` it was due to a coding mistake and it should console.error a message that includes the path that was sent in, and then it return the default values of 'notSet' for settingsPageName and settingsPageType.

If the path contains `/-/` then it is an Opinion and it should return 'OpinionSource' and 'reference'

SailingSteve (changes requested):
comment obsolete

src/js/utils/lookupPageNameAndPageTypeDict.js
SailingSteve: Fixed.

SailingSteve (approved):

SailingSteve (approved):
Looks good!

PR #4402, opened 2025-06-20
[WV-1138] Add dataLayer to "Open in new tab symbol" on Candidate side panel [MERGED]

dboudagian:
WV-1138

- passed candidateDetails to dataLayer + propTypes
- Added candidateDetails prop to CardForListBody.jsx and passed it into OpenExternalWebSite.jsx
- Updated sendExternalLinkInfoToGTM to include candidateDetails in dataLayer for enhanced tracking
- Defined PropTypes for candidateDetails in OpenExternalWebSite.jsx
- Cleaned up all lint errors and warnings in related files

### DIFF
--- a/src/js/common/components/CardForListBody.jsx
+++ b/src/js/common/components/CardForListBody.jsx
@@ -90,7 +90,7 @@ function CardForListBody (props) {
                     <LaunchIconWrapper>
                       <Suspense fallback={<></>}>
                         <OpenExternalWebSite
-                          linkIdAttribute="openPoliticianPage"
+                          linkIdAttribute="openCandidateInNewTab"
                           url={politicianDetailsURL}
                           target="_blank"
                           className="open-web-site open-web-site__no-right-padding"
@@ -109,6 +109,14 @@ function CardForListBody (props) {
                           destinationPageName={destinationPage.pageName}
                           destinationPageType={destinationPage.pageType}
                           trackingOn
+                          candidateDetails={{
+                            candidateWeVoteId: candidateWeVoteId || politicianWeVoteId,
+                            candidateName: ballotItemDisplayName,
+                            officeName,
+                            politicianWeVoteId,
+                            politicalParty,
+                            stateCode,
+                          }}
                         />
                       </Suspense>
                     </LaunchIconWrapper>
--- a/src/js/common/components/CardForListBody.jsx
+++ b/src/js/common/components/CardForListBody.jsx
@@ -62,7 +62,7 @@ function CardForListBody (props) {
     politicalPartySvgNameWithPath = '../../img/global/svg-icons/political-party-working-families.svg';
   }
   const politicianDetailsURL = `${webAppConfig.WE_VOTE_URL_PROTOCOL + webAppConfig.WE_VOTE_HOSTNAME}${politicianBasePath}`;
-  const destinationPage = lookupPageNameAndPageTypeDict(politicianDetailsURL);
+  const destinationPage = lookupPageNameAndPageTypeDict(politicianBasePath);
   // console.log('politicianBasePath:', politicianBasePath);
   // console.log('CardForListBody politicianDetailsURL:', politicianDetailsURL, ', destinationPage: ', destinationPage);
   const location = useLocation();
--- a/src/js/common/components/CardForListBody.jsx
+++ b/src/js/common/components/CardForListBody.jsx
@@ -62,7 +62,7 @@ function CardForListBody (props) {
     politicalPartySvgNameWithPath = '../../img/global/svg-icons/political-party-working-families.svg';
   }
   const politicianDetailsURL = `${webAppConfig.WE_VOTE_URL_PROTOCOL + webAppConfig.WE_VOTE_HOSTNAME}${politicianBasePath}`;
-  const destinationPage = lookupPageNameAndPageTypeDict(politicianBasePath);
+  const destinationPage = lookupPageNameAndPageTypeDict(politicianDetailsURL);
   // console.log('politicianBasePath:', politicianBasePath);
   // console.log('CardForListBody politicianDetailsURL:', politicianDetailsURL, ', destinationPage: ', destinationPage);
   const location = useLocation();
@@ -109,8 +109,6 @@ function CardForListBody (props) {
                           destinationPageName={destinationPage.pageName}
                           destinationPageType={destinationPage.pageType}
                           trackingOn
-                          candidateWeVoteId={candidateWeVoteId}
-                          politicianWeVoteId={politicianWeVoteId}
                         />
                       </Suspense>
                     </LaunchIconWrapper>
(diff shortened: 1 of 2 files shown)


dboudagian:
@DaleMcGrew This PR adds a dataLayer to "Open in new tab symbol" on Candidate side panel. Ready for review!

src/js/common/components/CardForListBody.jsx
```diff
@@ -90,7 +90,7 @@ function CardForListBody (props) {
                     <LaunchIconWrapper>
                       <Suspense fallback={<></>}>
                         <OpenExternalWebSite
-                          linkIdAttribute="openPoliticianPage"
+                          linkIdAttribute="openCandidateInNewTab"
```
DaleMcGrew: In general I would be cautious about changing existing ids, because the QA Automation team may have used the id in their tests. In this case they aren't using it yet, but I think the `openPoliticianPage` terminology is better.

src/js/common/components/Widgets/OpenExternalWebSite.jsx
```diff
@@ -23,9 +23,11 @@ export default class OpenExternalWebSite extends Component {
       // console.log('External link clicked:', this.props.url);
       const dataLayerObj = {
         actionDetails: {
+          actionType: 'navigate',
           buttonId: linkIdAttribute,
+          ...(candidateDetails && { candidateDetails }),
```
DaleMcGrew: We want to add a candidateDetails dictionary at the same level as actionDetails and destinationDetails, as opposed to with the actionDetails Dictionary.

Also, instead of always adding candidateDetails to dataLayerObject, let's only add it to dataLayerObject if the candidateWeVoteId prop exists. Ex/ (this is off the top of my head, so double-check my syntax)
```
if (candidateWeVoteId) {
  dataLayerObject.candidateDetails = {
    <DETAILS HERE>
  };
}
if (politicianWeVoteId) {
  dataLayerObject.politicianDetails = {
    <DETAILS HERE>
  };
}
```
I think it makes sense to populate both the candidateDetails dictionary AND the politicianDetails.

src/js/common/components/Widgets/OpenExternalWebSite.jsx
```diff
@@ -23,9 +23,11 @@ export default class OpenExternalWebSite extends Component {
...
         },
-        event: 'click',
+        event: 'action',
```
DaleMcGrew: When leaving the current site/window (which is always the case in this component), the Google Analytics standard for event is 'click' instead of 'action'.

src/js/common/components/Widgets/OpenExternalWebSite.jsx
```diff
@@ -114,4 +116,12 @@ OpenExternalWebSite.propTypes = {
   title: PropTypes.string,
   trackingOn: PropTypes.bool,
   url: PropTypes.string.isRequired,
+  candidateDetails: PropTypes.shape({
```
DaleMcGrew: Instead of passing in a complex package of data, let's pass in both candidateWeVoteID and politicianWeVoteId, and then retrieve all of these attributes within the OpenExternalWebSite component with CandidateStore.getCandidateByWeVoteId(yyy) or PoliticianStore.getPoliticianByWeVoteId(yyy).

src/js/common/components/CardForListBody.jsx
```diff
@@ -109,6 +109,14 @@ function CardForListBody (props) {
                           destinationPageName={destinationPage.pageName}
                           destinationPageType={destinationPage.pageType}
                           trackingOn
+                          candidateDetails={{
+                            candidateWeVoteId: candidateWeVoteId || politicianWeVoteId,
```
DaleMcGrew: Instead of passing in a complex package of data here, let's pass in both the candidateWeVoteId and politicianWeVoteId, and then retrieve all of these values within the OpenExternalWebSite component with src/js/stores/CandidateStore.js CandidateStore.getCandidateByWeVoteId(yyy) or src/js/common/stores/PoliticianStore.js .getPoliticianByWeVoteId(yyy)

DaleMcGrew (changes requested):
You have a good start! I have some requested changes.

dboudagian:
@DaleMcGrew This PR makes changes from your feeback!

dboudagian:
@DaleMcGrew  Sorry for adding the upstream pull to get the new codebase update. Please check commits from today (June 24) - those are the accurate changes for this ticket.
Summary of actual changes made:

Updated OpenExternalWebSite component per code review feedback
Changed event type from 'action' to 'click' for external link tracking
Pass candidateWeVoteId and politicianWeVoteId as separate props instead of candidateDetails object
Retrieve candidate/politician data within OpenExternalWebSite component using respective stores
Added conditional logic to populate both candidateDetails and politicianDetails in dataLayerObject
Removed unused PropTypes and corrected import paths

SailingSteve:
@dboudagian This PR includes lots of code changes -- 25 files!  Most of which are not related to your work.   I'm guessing that you might not have 

1. stashed your changes
2. checked out develop
3. git pull upstream develop
4. git push origin develop
5. create a new branch
6. unstash your changes

One way to fix this is to 

1. open your branch (in WebStorm, if you use that)
2. git compare your branch with remote branch upstream develop
3. Go through the list of files with differences, and for each file if you are sure that you did not make any changes in that file, right click on it and select "Get" to overwrite your diverged file with the latest from Git.  You can also double click on those differing files on the list, to see what is different before overwriting the changes.
4. Then when you are done, confirm that just your intended changes are left, and git/commit directory
5. Then in github at https://github.com/wevote/WebApp/pull/4402 instead of 25 files, you should just have one or two, and then change the state of the PR in the title, back to [MERGE READY]

dboudagian:
@DaleMcGrew and @SailingSteve Here are the changes with just the commits for these two files

src/js/common/components/CardForListBody.jsx
```diff
@@ -62,7 +62,7 @@ function CardForListBody (props) {
     politicalPartySvgNameWithPath = '../../img/global/svg-icons/political-party-working-families.svg';
   }
   const politicianDetailsURL = `${webAppConfig.WE_VOTE_URL_PROTOCOL + webAppConfig.WE_VOTE_HOSTNAME}${politicianBasePath}`;
-  const destinationPage = lookupPageNameAndPageTypeDict(politicianBasePath);
+  const destinationPage = lookupPageNameAndPageTypeDict(politicianDetailsURL);
```
SailingSteve: Thanks for cleaning up the other files!

lookupPageNameAndPageTypeDict doesn't take a URL, it takes a path ... so for example instead of https://wevote.us/ballot it just takes /ballot.

If you look into the lookupPageNameAndPageTypeDict function, all of those `path.startsWith` would fail if the path was a URL and started with https -- they assume paths.

So the code was correct before this proposed change.

SailingSteve (changes requested):
Change needed.

dboudagian:
@SailingSteve I made the change back to politicianBasePath

SailingSteve:
Looks good!

PR #4406, opened 2025-06-22
WV-1081: Ballot now scrolls to the top, when link from 'preview your Ballot' is clicked.

SailingSteve:


DaleMcGrew:
Good catch. 👍

PR #4411, opened 2025-06-24
[WV-1089]  Add dataLayer for all links under "More candidate information" and Update lookup...External URL function.

seanquinn917:
Verify that GTM layer is built and functioning in the More Candidate Information Links that direct to external URL's.  Add conditionals into lookupPageNameAndTypeDictForExternalUrls to include Twitter, Youtube, Instagram, Wiki, X, Google, Bing and a candidates personal website.


SailingSteve (approved):
Ready to go

PR #4420, opened 2025-07-01
WV-924: Cordova errors seen in Android console

SailingSteve:
Was getting errors in OverlayTrigger, so upgraded react-bootstrap from v1 to v2, checked code against migration guide and it looks ok. 

Then disabled OverlayTrigger in Cordova via bldSrcCordova.


DaleMcGrew:
👍

PR #4421, opened 2025-07-01
Fixed settings issue where /settings/securityAndSignIn was broken

SailingSteve:
Too many overlapping changes to find where this came from, but we use the two routes interchangeably, and we were not picking it up on the SettingsDashboard.jsx

As a result you could not get to /settings/securityAndSignIn

Did not want to make a change in components/Navigation/SettingsPersonalSideBar.jsx which is in the middle of a series of changes, by another developer.


DaleMcGrew:
👍

PR #4423, opened 2025-07-01
WV-1197: Android "backbutton" suppressed in Chrome on Android device and in Cordova

SailingSteve:
It was too messy to fix as requested, and would have required a app-wide refactor of pushHistory, plus messy one-off fixes when navigating in a dialog, when not using routing/pushHistory.

"backbutton" which is an Android only feature, no longer kicks you out of Cordova, nor does it kick you out of Chrome when browsing the WebApp on an Android device. 

The "backbutton" kicking you out of Chrome when browsing the WebApp on an Android device, is a current bug, that is resolved by this PR.

Some lint cleanup in Donate.


DaleMcGrew:
👍

PR #4424, opened 2025-07-02
[WV-1198] Add dataLayer for all "non-socials links" on FAQ Page

fnoble1305:
added dataLayer code (action) for all internal clicks and added dataLayer for all external links. also fixed ESLint warnings.


src/js/common/components/FAQBody.jsx
```diff
@@ -135,18 +173,18 @@ export default class FAQBody extends Component {
...
         {' '}
-        <Suspense fallback={<></>}>
-          <OpenExternalWebSite
-            linkIdAttribute="weVoteAboutUsPage"
-            url={`${webAppConfig.WE_VOTE_URL_PROTOCOL + webAppConfig.WE_VOTE_HOSTNAME}/more/about`}
-            target="_blank"
-            body="volunteer board members"
-          />
-        </Suspense>
+        <Link
```
SailingSteve: The code before you changed it has a bug.  This should be a historyPush, otherwise by linking to wevote as if it were an external site, will cause the session to be destroyed and recreated, potentially losing data. 
So in this case at this line you should call a onCick function like 
`routeToAbout () {
    this.pushDataLayer('/more/about', 'faq-more-about')
    historyPush(''/more/about');
}
`

src/js/common/components/FAQBody.jsx
```diff
@@ -257,7 +299,7 @@ export default class FAQBody extends Component {
           <>
             If you like WeVote,
             {' '}
-            <Link to="/donate" className="u-cursor--pointer u-link-color">
+            <Link to="/donate" className="u-cursor--pointer u-link-color" onClick={() => this.pushDataLayer('/donate', 'faq-donate')}>
```
SailingSteve: Same issue as above, this should not be a <Link

src/js/common/components/FAQBody.jsx
```diff
@@ -322,18 +365,19 @@ export default class FAQBody extends Component {
...
             className="open-web-site open-web-site__no-right-padding"
             body="sign up to volunteer"
+            trackingOn
           />
         </Suspense>
         {' '}
         and
         {' '}
-        <Link to="/donate" className="u-cursor--pointer u-link-color">
+        <Link to="/donate" className="u-cursor--pointer u-link-color" onClick={() => this.pushDataLayer('/donate', 'faq-donate-monthly')}>
```
SailingSteve: Same issue as above, this should not be a <Link

src/js/common/components/FAQBody.jsx
```diff
@@ -322,18 +365,19 @@ export default class FAQBody extends Component {
...
           donate $5 monthly
         </Link>
         .
         <br />
         <br />
-        <Link to="/" className="u-cursor--pointer u-link-color">
+        <Link to="/" className="u-cursor--pointer u-link-color" onClick={() => this.pushDataLayer('/', 'faq-home')}>
```
SailingSteve: Same issue as above, this should not be a <Link

SailingSteve (commented):
Needs changes, but mostly there!

PR #4426, opened 2025-07-02
[WV-1458]Enhanced data layer with additional user details for Ballot link[TEAM REVIEW]

pi-sen:
Updated new details to be tracked under user details for Ballot link in the existing datalayer


SailingSteve (approved):
Looks good!

PR #4436, opened 2025-07-08
[WV-1463] Add dataLayer on FAQ page for Landing event

tahn1234:
WV-1463

- Added TagManager.dataLayer in componentDidUpdate to fire the landing event with: pageDetails (pageName, pageType, pathname), userDetails (voterWeVoteId, stateCode, userCohort)
- Ensures it fires only once and only after voter data has been retrieved.
- Confirmed parameters align with analytics requirements.

### DIFF
--- a/src/js/pages/More/FAQ.jsx
+++ b/src/js/pages/More/FAQ.jsx
@@ -1,10 +1,13 @@
 import React, { Component } from 'react';
+import TagManager from 'react-gtm-module';
 import { Helmet } from 'react-helmet-async';
 import styled from 'styled-components';
 import FAQBody from '../../common/components/FAQBody';
 import { isCordova } from '../../common/utils/isCordovaOrWebApp';
 import { renderLog } from '../../common/utils/logging';
 import { PageContentContainer } from '../../components/Style/pageLayoutStyles';
+import VoterStore from '../../stores/VoterStore';
+import lookupPageNameAndPageTypeDict from '../../utils/lookupPageNameAndPageTypeDict';
 
 export default class FAQ extends Component {
   static getProps () {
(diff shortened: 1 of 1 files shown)


src/js/pages/More/FAQ.jsx
```diff
@@ -1,10 +1,13 @@
...
 import styled from 'styled-components';
 import FAQBody from '../../common/components/FAQBody';
 import { isCordova } from '../../common/utils/isCordovaOrWebApp';
 import { renderLog } from '../../common/utils/logging';
 import { PageContentContainer } from '../../components/Style/pageLayoutStyles';
+import VoterStore from '../../stores/VoterStore';
+import lookupPageNameAndPageTypeDict from '../../utils/lookupPageNameAndPageTypeDict';
 
 export default class FAQ extends Component {
   static getProps () {
```
SailingSteve: Remove the getProps function, it is a leftover from an old approach

Replace it with 
```
  constructor (props) {
    super(props);
    this.state = {
      dataLayerFired: false,
    };
  }
```

You need to set an initial value for dataLayerFired, otherwise when this page loads componentDidMount() is called and crashes in checkAndFireDataLayer since dataLayerFired is null

src/js/pages/More/FAQ.jsx
```diff
@@ -13,6 +16,38 @@ export default class FAQ extends Component {
 
   componentDidMount () {
     window.scrollTo(0, 0);
+    this.checkAndFireDataLayer();
+  }
+
+  componentDidUpdate () {
+    this.checkAndFireDataLayer();
```
SailingSteve: This call to checkAndFireDataLayer() is unnecessary, since this is a landing event, which would be handled by componentDidMount.  In general unnecessary code should be removed.

src/js/pages/More/FAQ.jsx
```diff
@@ -13,6 +16,38 @@ export default class FAQ extends Component {
...
+
+      TagManager.dataLayer({
+        dataLayer: {
+          event: 'landing',
+          pageDetails: {
+            pageName: currentPage.pageName,
+            pageType: currentPage.pageType,
+            pathname: currentPathname,
+          },
+          userDetails: {
```
SailingSteve: Replace the userDetails part with Dales new function
`userDetails: VoterStore.getAnalyticsUserDetails(),`
(It didn't exist when you wrote this code!)

SailingSteve (changes requested):

SailingSteve (approved):
Looks good!

PR #4437, opened 2025-07-08
[WV-1460]Added data layer tracking for Challenges button

pi-sen:
Added data layer tracking for Challenges button

### DIFF
--- a/src/js/components/Navigation/HeaderBar.jsx
+++ b/src/js/components/Navigation/HeaderBar.jsx
@@ -4,6 +4,7 @@ import withStyles from '@mui/styles/withStyles';
 import PropTypes from 'prop-types';
 import React, { Component, Suspense } from 'react';
 import styled from 'styled-components';
+import TagManager from 'react-gtm-module';
 import OrganizationActions from '../../actions/OrganizationActions';
 import VoterActions from '../../actions/VoterActions';
 import VoterGuideActions from '../../actions/VoterGuideActions';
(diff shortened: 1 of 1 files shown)


src/js/components/Navigation/HeaderBar.jsx
```diff
@@ -577,7 +578,35 @@ class HeaderBar extends Component {
                     <TabWithPushHistory
                       classes={isWebApp() ? { root: classes.tabRootDonateDesktop } : { root: classes.tabRootDonate }}
                       value={squadsValue}
-                      change={this.handleTabChange}
+                      change={(value) => {
```
SailingSteve: This works fine, but it is too much inline code to put in a render (functional component return).  This is too different from the way we do it elsewhere.  Please move it to a function like onClickChallenges, or you could add it conditionally to handleTabChange.

src/js/components/Navigation/HeaderBar.jsx
```diff
@@ -577,7 +578,35 @@ class HeaderBar extends Component {
...
+                              pageType: currentPage.pageType,
+                              pathname: currentPathname,
+                            },
+                            destinationDetails: {
+                              pageName: destinationPage.pageName,
+                              pageType: destinationPage.pageType,
+                              pathname: destinationPathname,
+                              stateCode: VoterStore.getVoterStateCode(),
+                            },
+                            userDetails: {
```
SailingSteve: Dale recently made a common function for userDetails, which is much better than building the data in each dataLayer object.  Line 600 becomes  simply  `userDetails: VoterStore.getAnalyticsUserDetails(),`

SailingSteve (changes requested):
Works well, just needs some "consistency with other code changes"!

pi-sen:
Moved the datalayer tracking code into handleTabChange
Updated the implementation as per the review feedback

src/js/components/Navigation/HeaderBar.jsx
pi-sen: Done

src/js/components/Navigation/HeaderBar.jsx
pi-sen: Done

SailingSteve (approved):
Looks good!

PR #4439, opened 2025-07-08
[WV-1445] Add dataLayer on Topic page for endorser logo or name

YousraElzamzami:
WV-1445 

- Created function `sendEndorserClickEvent` that takes in `buttonId `
- Added onClick handlers on `organizationDisplayLogo` and `organizationDisplayName` to call this function.

\***Question***
When an organization has a Twitter handle, the URL takes the form /{twitterHandle}, and destinationDetails reports for example: 
Screenshot 2025-07-08 at 4 31 16 PM

If the organization doesn’t have a Twitter handle, the fallback URL is /voterguide/{organizationId}, and destinationDetails for example, looks like:
Screenshot 2025-07-08 at 4 31 07 PM

Should I leave this distinction as-is, or try to normalize these values so they follow a consistent structure regardless of the URL format?

### DIFF
--- a/src/js/components/Organization/OrganizationDisplayForList.jsx
+++ b/src/js/components/Organization/OrganizationDisplayForList.jsx
@@ -1,6 +1,7 @@
 import PropTypes from 'prop-types';
 import React, { Component, Suspense } from 'react';
 import { Link } from 'react-router-dom';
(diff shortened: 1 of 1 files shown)


SailingSteve:
Looks good!

PR #4440, opened 2025-07-08
Multiple changes for Cordova

SailingSteve:
Added jsconfig.json if Cordova so that
OverlayTrigger can be swapped out for CordovaOverlayTrigger without having to calculate the path for import for each file in bldSrcCordova. 
Conditioned display of Import Contacts with ENABLE_NEXT_RELEASE_FEATURES since there is no destination page for that link.


DaleMcGrew:
👍  Thank you @SailingSteve!

PR #4442, opened 2025-07-09
Upgrade Storybook, to clear related npm high severity vulnerabilities

SailingSteve:
Tested, and it works fine.


SailingSteve:
All are now cleared, except for  2 from pre-commit, which has not been maintained in 9 year, but is still widely used.
Consider replacing it with Husky (13M weekly downaloads) 
https://wevoteusa.atlassian.net/browse/WV-1559

PR #4451, opened 2025-07-10
Made a getPageDetails() convenience function to get rid of tons of inconsistent boilerplate

SailingSteve:
1) pageDetails: getPageDetails(): I left about 5 of the old ones in place in case their code did something valuable. 
2) destinationDetails could use one too.
3) lookupPageNameAndPageTypeDict(currentPathname); very rarely needs a path to be passed in (unless destination).  That could eliminate some more boilerplate

### DIFF
--- a/src/js/common/components/CampaignShare/ShareByCopyLink.jsx
+++ b/src/js/common/components/CampaignShare/ShareByCopyLink.jsx
@@ -5,13 +5,13 @@ import React, { Component } from 'react';
 import CopyToClipboard from 'react-copy-to-clipboard';
 import TagManager from 'react-gtm-module';
 import styled from 'styled-components';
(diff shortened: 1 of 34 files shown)


SailingSteve:
@DaleMcGrew if you are not around, I'll merge this tomorrow to avoid other developer's overlapping merges.
I spot checked about 15 of them, and all were fine... pretty low risk.

DaleMcGrew:
Hi @SailingSteve this refactoring makes a lot of sense. Good call! The places where we have to break out the pageDetails dictionary locally are the exception (not the rule), in cases like in this file: src/js/common/components/Widgets/OpenExternalWebSite.jsx
...which you left in place. So I like it! 👍

PR #4453, opened 2025-07-12
Cordova "Custom photo" button now is (and stays) selected after uploading photo.

SailingSteve:
Also fixes same bug in WebApp


DaleMcGrew:
👍

PR #4344, opened 2025-05-27
WV-1098 Add dataLayer for "Search" on Ballot Page [TEAM REVIEW]

ilonagog:
Added handleClick in src/js/components/Filter/FilterBaseSearch.jsx

PR #4472, opened 2025-07-23
[WV-1098] Add dataLayer for Search on Ballot Page [TEAM REVIEW]

YousraElzamzami:
WV-1098

- Added function `sendSearchEvent` that takes in `searchText`
- Updated `searchNewItems` to call `sendSearchEvent` when a new unique search term is entered.

In the previous PR (#4344), dataLayer was recommended to be placed inside the uniqueness check (`if (!searchTextAlreadyRetrieved.includes(searchText))`). However, the outer guard `if (opinionsAndBallotItemsSearchMode)` seems to always be undefined, so the whole block never runs. I wasn’t able to determine why `opinionsAndBallotItemsSearchMode` remains undefined here, so I’d appreciate any insight!

I’ve commented out that guard to allow data fetching and event sending for every unique search term per session, preserving uniqueness.

If keeping `opinionsAndBallotItemsSearchMode` is necessary, I can move the event call outside that check, but that would lose the uniqueness control and trigger on every search. Appreciate any feedback!

### DIFF
--- a/src/js/components/Filter/FilterBaseSearch.jsx
+++ b/src/js/components/Filter/FilterBaseSearch.jsx
@@ -2,6 +2,7 @@ import { Close, Search } from '@mui/icons-material';
 import { IconButton, InputBase } from '@mui/material';
 import withStyles from '@mui/styles/withStyles';
 import withTheme from '@mui/styles/withTheme';
+import TagManager from 'react-gtm-module';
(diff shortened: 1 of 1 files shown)


DaleMcGrew:
👍 Great work @YousraElzamzami!

PR #4477, opened 2025-07-28
[WV-1137]Filter undefined values in data layer tracking on Candidate modal[TEAM REVIEW]

pi-sen:
Added condition that if any field in politicianDetails are absent or undefined, it wont be sent to the datalayer


DaleMcGrew:
👍

PR #4480, opened 2025-07-29
[WV-1447] Add dataLayer on Topic page for Landing event [Team Review]

seanquinn917:
DataLayer added to topics (Values) landing page.  Fires after user clicks on value/topic and lands on that topic's page. 

Additionally, there were many lint errors that we fixed when creating this request. The new code is found only in OneValue.jsx. The remaining 24 files that were changed are all simply lint corrections.


DaleMcGrew:
Thank you @seanquinn917! 👍

PR #4484, opened 2025-07-30
[WV-1010] [WV-1460] Add dataLayer code to the Header items, when they are clicked [NEEDS REVIEW]

Terry-z04:

[WV-1010](https://wevoteusa.atlassian.net/issues/?filter=10060&selectedIssue=WV-1010)

1. Added dataLayer function in `TabWithPushHistory` to send click event and its corresponding tag information for HeaderBar items.
2. Fixed `ChallengesHome` landing event dataLayer
 - Commented out original code in `HeaderBar`
 - Added landing dataLayer in `ChallengesHome`

[WV-1010]: https://wevoteusa.atlassian.net/browse/WV-1010?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/js/common/pages/Challenge/ChallengesHome.jsx
+++ b/src/js/common/pages/Challenge/ChallengesHome.jsx
@@ -4,12 +4,14 @@ import PropTypes from 'prop-types';
 import React, { Component, Suspense } from 'react';
 import { Link } from 'react-router-dom';
 import styled from 'styled-components';
+import TagManager from 'react-gtm-module';
 import ActivityActions from '../../../actions/ActivityActions';
 import BallotActions from '../../../actions/BallotActions';
 import IssueActions from '../../../actions/IssueActions';
 import OrganizationActions from '../../../actions/OrganizationActions';
 import SupportActions from '../../../actions/SupportActions';
 import ChallengeStore from '../../stores/ChallengeStore';
+import { getPageDetails } from '../../../utils/lookupPageNameAndPageTypeDict';
 import apiCalming from '../../utils/apiCalming';
 import arrayContains from '../../utils/arrayContains';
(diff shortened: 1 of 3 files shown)


src/js/common/pages/Challenge/ChallengesHome.jsx
```diff
@@ -91,6 +94,28 @@ class ChallengesHome extends Component {
     // AnalyticsActions.saveActionOffice(VoterStore.electionId(), params.office_we_vote_id);
   }
 
+  componentDidUpdate () {
```
DaleMcGrew, line 97: Hi @Terry-z04 this looks correct, great work. We are finding that some components mount (componentDidMount) and then don't have to update, which will lead to the dataLayer not getting sent. Please move the dataLayer code into a new function, and then check to send it in both componentDidMount and componentDidUpdate. Thank you!

src/js/components/Navigation/TabWithPushHistory.jsx
```diff
@@ -1,22 +1,45 @@
...
+
+// DataLayer helper function
+function pushDataLayer (destinationPath, buttonId = '') {
+  const destinationPage = lookupPageNameAndPageTypeDict(destinationPath);
+  const dataLayerObject = {
+    actionDetails: {
+      actionType: 'navigate',
+      buttonId,
+    },
+    event: 'click',
```
DaleMcGrew, line 24: Since this is an internal event (instead of going to another website), this should be:
```
event: 'action',
```

DaleMcGrew (approved):
👍

PR #4485, opened 2025-07-30
[WV-1161] Align EditBanner width with menu container and style

itcreativeusa:
WV-1161: Align EditBanner width with menu container and style

- Added EditBannerWrapper with horizontal padding and centered layout
- Set EditBanner max-width to 960px to match menu width

image


DaleMcGrew:
👍

PR #4500, opened 2025-08-10
wv-1483 and changes for wv-1455

fnoble1305:
I had originally been creating a dataLayer from scratch, and I had some trouble with the destination piece. Then I realized it was already connected to OpenExternalWebsite, so I enabled the tracking for this action.


DaleMcGrew:
👍

PR #4507, opened 2025-08-13
[WV-1736] Prevent photo area from getting too narrow on Politician Campaign Details page

itcreativeusa:
WV-1736: Prevent photo area from getting too narrow on Politician Campaign Details page

Comment out min-width in ColumnTwoThirds and note its purpose for layout alignment


DaleMcGrew:
👍

PR #4513, opened 2025-08-22
Moved NotificationBannerAboveHeader into its own file. Now able to know from API server data if the voter can edit candidate's profile. Opening modal with six digit code when a candidate is claiming a profile by email address.

DaleMcGrew:
Moved NotificationBannerAboveHeader into its own file. Now able to know from API server data if the voter can edit candidate's profile. Opening modal with six digit code when a candidate is claiming a profile by email address.


DaleMcGrew:
👍

PR #4515, opened 2025-08-25
Cordova iOS only, for the first 30 seconds send all console.log()s to CloudWatch

SailingSteve:
To hopefully find out why some iPhones are crashing with the latest released version.
Have to release this in the App Store to get it on developers/users phone without sideloading
Will take this out once I figure out the problem in https://wevoteusa.atlassian.net/browse/WV-1181


DaleMcGrew:
👍

PR #4549, opened 2025-10-02
[WV-1733]-more-header-menu

itcreativeusa:
Make 'Candidates I’m managing' menu item highlight correctly by matching both 'manage' and 'more/manage'.


DaleMcGrew:
👍

PR #4567, opened 2025-10-14
[WV-1152] Changes for Cordova iOS 2.7.3.15

SailingSteve:
Major "header spacing in iOS" rework, I found lists of all iOS internal model IDs and am now using those to determine which Apple device is running.
There is a small chance that the iOS blinking issue could be resolved by one fix.
Also fixes https://wevoteusa.atlassian.net/browse/WV-1152


DaleMcGrew:
👍

PR #4575, opened 2025-10-23
[WV-1733] Fix(header): correct tab highlight after navigation (sync ta…

itcreativeusa:
…bs with route) Body

- Fix tabs no longer stays highlighted after clicking the WeVote logo to return home; the active tab now matches the current route.
- Move getTabsValueFromPage and syncTabsToRoute into HeaderBar as class fields
- Call syncTabsToRoute() in componentDidMount and whenever the URL page changes
- Remove redundant handleTabChange bind; avoid double highlight calls
- Keep jQuery-based customHighlightSelector but drive it from route-derived state

### DIFF
--- a/src/js/components/Navigation/HeaderBar.jsx
+++ b/src/js/components/Navigation/HeaderBar.jsx
@@ -72,29 +72,29 @@ class HeaderBar extends Component {
   }
 
   getTabsValueFromPage = () => {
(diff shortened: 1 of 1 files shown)


DaleMcGrew:
Thank you @itcreativeusa! 👍

PR #4589, opened 2025-11-09
Big Change in Cordova Headers for iOS

SailingSteve:
iOS had the toughest header offset issues for Cordova.  
Eliminated the tables of offsets, and replaced them with calculated values based on the heights of the actual rendered DOM objects in the various header components.

Version:  2.7.4.0    is in TestFlight


DaleMcGrew:
👍

PR #4592, opened 2025-11-12
[WV-2157] Fix: Hide endorsement tooltip for candidates with zero endorsements

lyndabanh:
Fixes WV-2157

* Fixed bug where endorsement tooltip appeared on candidates with 0 endorsements
* Updated condition in `PositionRowListCompressed.jsx` to check `filteredPositionList.length > 0` instead of just checking if array exists

### Testing
* Tested on localhost with candidates that have 0 endorsements → tooltip no longer appears
* Tested with candidates that have 1+ endorsements → tooltip still appears correctly


DaleMcGrew:
Great catch, thank you @lyndabanh! 👍

PR #4593, opened 2025-11-12
Fix- endorsers count mismatch-WV-1470[MERGE READY]

anup2307:


DaleMcGrew:
Thank you @anup2307! 👍

PR #4617, opened 2025-12-03
GoogleAutoComplete working again in WebApp

SailingSteve:
This had nothing to do with API keys or permissions, and almost for sure had nothing to do with Cordova.
It was some sort of css problem related to the pop-up address dialog -- I just overrode the css for now.  
This is a short lived fix, because we will have to upgrade to Google places.PlaceAutocompleteElement() at some point in the future, to avoid using "Places Services (Legacy)". 
Ideally we can wait for 'react-google-autocomplete' to get updated.

This console message is true, but isnot the cause of the problem:  
**As of March 1st, 2025, google.maps.places.Autocomplete is not available to new customers. Please use google.maps.places.PlaceAutocompleteElement instead. At this time, google.maps.places.Autocomplete is not scheduled to be discontinued, but google.maps.places.PlaceAutocompleteElement is recommended over google.maps.places.Autocomplete. While google.maps.places.Autocomplete will continue to receive bug fixes for any major regressions, existing bugs in google.maps.places.Autocomplete will not be addressed. At least 12 months notice will be given before support is discontinued. Please see https://developers.google.com/maps/legacy for additional details and https://developers.google.com/maps/documentation/javascript/places-migration-overview for the migration guide.**

------------
Also added some lint cleanup
Some header size code for iPads (already in the released version)
And fully removed the historyPush('/settings/hamburger'), which has been replaced with a drawer, but was still incorrectly tripping in some circumstances.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -22,6 +22,7 @@
         "@storybook/addon-actions": "^7.6.20",
         "@stripe/react-stripe-js": "^1.3.0",
         "@stripe/stripe-js": "^1.13.0",
+        "@vis.gl/react-google-maps": "^1.7.1",
         "classnames": "^2.3.1",
         "clsx": "^1.1.1",
         "d3-geo": "^2.0.1",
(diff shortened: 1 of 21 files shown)


DaleMcGrew:
Thank you @SailingSteve for chasing this down! 👍

PR #4621, opened 2025-12-05
[WV-2285] Added condition to handle empty voter opinion in SupportStore.jsx

Almee98:
This PR addresses the issue where deleting an opinion from the “Delete Opinion” menu did not immediately update the UI.
The problem occurred because SupportStore.js did not properly handle the case where statement_text was an empty string.
As a result, the local state was never updated when a voter deleted their opinion, and the deleted comment would remain visible until the page was refreshed.

This fix introduces logic within the voterPositionCommentSave case to explicitly check for statement_text === ''.
When this condition is true, the voter’s opinion entry is removed from the voter_statement_text dictionary (for both ballotItemWeVoteId and politicianWeVoteId, if present).


DaleMcGrew:
Thank you Almee! This looks like a good solution. 👍

PR #4625, opened 2025-12-12
WV-130 frontend React store and HearFavoriteToggleBase fix so that correct follow count is shown even after voter delete

Arjunsivakumar28:
- Updated organization store to resetState() on voter delete
- Updated campaign store to change supporters count on voter delete and resetVoterSpecificData()
- Ensure current prop values re used on props change at HeartFavoriteToggleBase.jsx


DaleMcGrew:
Great detective work @Arjunsivakumar28! Thank you for this fix. 👍

PR #4629, opened 2025-12-17
[WV-2048] Invite voters: "Invite selected" modal (Manage My Candidates) [TEAM REVIEW]

ric0de:
Some notes:

- **IMPORTANT** The "Paste list" modal doesn't let you import voters without an email, so in order to test for voters with missing emails, I hardcoded a list of invitees that overrides whatever list you currently had selected. To enable this, simply set const TEST_MODE to true.

- We might want to add a 'Back to invite' button to "Preview invitation" modal when accessing it from the "Invite selected" modal

- Currently, the first column on the invitees list (their names) has a fixed length of 95px, so any long names get cut off with ellipses. This column can be adjusted to be longer, and the subsequent columns will shift accordingly with no issue, but an idea is to maybe have every voter's last name just be their initial so we can leave the first column short for mobile. Currently any name longer than "John Smith" gets cut off.

- When there's only 1 invitee left, its respective "Remove from list" button becomes un-selectable to prevent an empty list

- Minor change, but I fixed the font for all the other modals' dialogTitleJSX since they had broken during the migration to Template A

### DIFF
--- a/src/js/components/More/ConfirmCloseModal.jsx
+++ b/src/js/components/More/ConfirmCloseModal.jsx
@@ -1,0 +1,135 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import styled, { createGlobalStyle } from 'styled-components';
+import { Info } from '@mui/icons-material';
+import DesignTokenColors from '../../common/components/Style/DesignTokenColors';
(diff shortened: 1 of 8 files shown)


DaleMcGrew:
Thank you @ric0de! 👍

PR #4634, opened 2025-12-24
Android Header Fixes

SailingSteve:
Added new Android plug-in that reports when the phone has a notch.
Consolidated IOSTopOfScreenSpacer with a new unified CordovaTopOfScreenSpacer that works consistently for both platforms.
Added donations to Android.
Made sure the Drawer "x" buttons are visible and work on Android phones with a notch. 
Broke up a dependency loop in ShareButton
Disambiguated some copied code.


DaleMcGrew:
Thank you @SailingSteve! 👍

PR #4637, opened 2026-01-03
[WV-754] Privacy add default visibility

mjacquot1:
Added a component to allow for selection of default visibility for your profile. This is not currently used in `SettingsYourData.jsx`  as the actual functionality isn't added yet, this is just the UI.

Text coloring might seem strange, but that is because I tried to match text colors used in `SettingsYourData.jsx`.

Testing:
Tested in StoryBook as well as in the local browser embedded in the expected location (Video included).
https://github.com/user-attachments/assets/9f7608d3-4fda-4217-9bb7-d1b349ab44ef


DaleMcGrew:
Thank you @mjacquot1! 👍

PR #4638, opened 2026-01-09
WV-2058: Add click animation when URL parameter is passed

Arjunsivakumar28:
- Added a ripple animation that triggers on click when the url parameter "?flashCursor=1" is passed (case insensitive). 
- When anything else is passed, the effect deactivates. 
- Animation triggers on any click on the web page


DaleMcGrew:
Really great work @Arjunsivakumar28! 👍

PR #4644, opened 2026-01-12
WV-23 - Added a Smart Tooltip wrapper around React Tooltip to use for different tooltip behaviors. Added tooltip on hover for Ballot tabs.

Almee98:
WV-23

Added tooltip for Ballot tabs.

Added a tooltip wrapper around React tooltip to help with different tooltip behaviors. Below is some information about what it is and how to use it:

`SmartToolTip.jsx` is a small wrapper around react-bootstrap’s OverlayTrigger that standardizes tooltip behavior across devices. It does two things that the base tooltip doesn’t handle well for us:

- Responsive trigger logic: it chooses hover on desktop and click/tap on mobile, and honors the triggerType prop (hover, click, both) so components can declare the intended behavior.
- Mobile UX cleanup: it closes the tooltip on outside tap and when another tooltip opens, which prevents multiple tooltips lingering on touch devices.
- Usage: wrap the target element with `<SmartTooltip title="..." triggerType="hover|click|both" placement="top|bottom|...">…</SmartTooltip>`

So it exists to give us consistent, predictable tooltip behavior without each component re‑implementing device detection and tap‑to‑close logic.

### DIFF
--- a/src/js/common/components/Widgets/ClaimedProfileIcon.jsx
+++ b/src/js/common/components/Widgets/ClaimedProfileIcon.jsx
@@ -1,78 +1,33 @@
-import React, { useState, useEffect, useRef } from 'react';
-import { OverlayTrigger, Tooltip } from 'react-bootstrap';
+import React, { useRef } from 'react';
 import SvgImage from './SvgImage';
+import SmartTooltip from './SmartToolTip';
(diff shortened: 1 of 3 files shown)


DaleMcGrew:
Thank you @Almee98 this is a great refactoring, and addition of SmartToolTip as reusable code. Great work! 👍

PR #4648, opened 2026-01-16
WV-2058: Added fixes to arrow flash on click based on testing team input

Arjunsivakumar28:
- Added a 0.5 second delay to the flash cursor effect
- Made sure that external links open after flash cursor effect is seen (might have pop up block issue)


DaleMcGrew:
👍

PR #4665, opened 2026-01-27
Updates To README_TESTING.md for changes needed in browserstack.config.js

nadiavaida:

When we create browserstack.config.js as a copy of browserstack.config.template.js, the test execution fails with below error:
WebApp/tests/browserstack_automation/config/wdio.config.js: Cannot read properties of undefined (reading 'NAME')

**Added the below note in  README_TESTING.md file related to the issue:**

As the process of migrating to ESM is not completed for all the files [https://wevoteusa.atlassian.net/browse/WV-2254], we need to make changes in the 'browserstack.config.js' file for now. After you create the local file browserstack.config.js as a copy of browserstack.config.template.js, please update the below in your local browserstack.config.js before running the tests:
module.exports = { browserStackConfig: {

instead of
export default {


charanya-QA:
Thanks Nadia!

PR #4673, opened 2026-01-30
Additional changes so cursor flashes on any button or link click

Arjunsivakumar28:
Now cursor flashes on click of most buttons and links before button or link navigation action takes place. Navigation to external links and new tabs wait until cursor flash animation is complete.


DaleMcGrew:
Very interesting work @Arjunsivakumar28. 👍

PR #4675, opened 2026-02-03
[WV-2472] Correct 'More...' footer button behavior on mobile

atarrh:

https://wevoteusa.atlassian.net/browse/WV-2472

Basically two changes; first, simply ignore the "showMoreMenu" state when determining which tab is shown as selected. Second, add an anchor to the "More..." button, so that it can be "ignored" when a user clicks outside of the ClickAwayListener overlay, after it has already been opened. This prevents the overlay from being reopened (when I believe a user re-clicking the "More..." button to close the overlay was double-counted - once as "clicking away from the listener" and again as "clicking the More... button").


DaleMcGrew:
Thank you @atarrh 👍

PR #4682, opened 2026-02-07
UI Fixes for Negative Feedback Bypass

SailingSteve:
Non-anonymous emails now have an id_string (header)
Standard buttons have been added to ReviewAppModal.jsx
Cancel of ReviewAppModal works
We now default to "checked" in ReviewAppModal if we have an email address.

Replaced old isMobile (that relied on user agent detection, which only works sometimes) with isMobileScreenSize()

Fixes https://wevoteusa.atlassian.net/browse/WV-2473


DaleMcGrew:
Thank you @SailingSteve! 👍

PR #4691, opened 2026-02-13
replace inline wrapper CSS with styled trigger div in SmartToolTip.jsx

Almee98:

- Replaced duplicated inline trigger styles with TooltipTriggerWrapper in SmartToolTip.jsx.
- Removed extra nested click-mode wrapper so hover/click paths share the same trigger element.


DaleMcGrew:
👍 Great upgrade @Almee98!

PR #4699, opened 2026-02-24
[WV-2475] Extend the itemActionBar to include "make public" interface [TEAM REVIEW]

ric0de:
Added visibility row + candidate staff section to item action bar

Wired up the chat bubble to toggle the make public row inline, added the "for candidate staff" claim & edit section, and gated everything behind a showCandidateStaffAndChat prop so it only shows in the ballot container for now. API calls are stubbed out for later

Mobile/tablet: 
image

Desktop: 
image

Other uses of ItemActionBar are unaffected: 
image

To enable chat bubble + "For candidate staff" (candidates/politicians only), pass showCandidateStaffAndChat prop to anywhere `<ItemActionBar />` is used

Also added a border under ModalTemplateA's header that only appears on scroll

image

### DIFF
--- a/src/js/components/Ballot/BallotScrollingContainer.jsx
+++ b/src/js/components/Ballot/BallotScrollingContainer.jsx
@@ -93,10 +93,7 @@ class BallotScrollingContainer extends Component {
     const { oneCandidate } = this.props;
     const politicianWeVoteId = oneCandidate ? oneCandidate.politician_we_vote_id : null;
     const dataLayerObject = {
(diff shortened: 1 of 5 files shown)


DaleMcGrew:
Looks solid @ric0de! Thank you. My only comment is that it looks like you rolled back some multi-line spacing required by the new ESLint rules. For future pull requests, just double-check `npm run lint` doesn't produced spacing warnings for the files you are working on. 👍

PR #4713, opened 2026-02-28
Fix loading of external pages in Android Cordova

SailingSteve:
Plus lots of lint fixes related to camel case in Actions files and no-dupe-else-if in various place -- these were blocking the running of WebApp on my local.

A better fix for the camel case, would be to never pass in camel case, instead, covert snake case to camel case in the function, right before sending the variable to the server.

Lots of other lint errors have appeared, probably due to a recent lint version upgrade.   (To see them all:  npm run lint)

Fixes https://wevoteusa.atlassian.net/issues?filter=10029&selectedIssue=WV-2246

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -66,6 +66,7 @@
     "react/no-did-update-set-state": 0,
     "react/no-string-refs": 1,
     "react/no-unused-prop-types": 1,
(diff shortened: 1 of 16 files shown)


DaleMcGrew:
👍

PR #4719, opened 2026-03-04
Fix loading of external pages in Android Cordova (Part 2)

SailingSteve:
Found a way to load the Google Play store from Android, and implemented it.  So I reverted to the prior UI for the links.

Fixes https://wevoteusa.atlassian.net/issues?filter=10029&selectedIssue=WV-2246


DaleMcGrew:
👍

PR #4738, opened 2026-03-19
WV-2058: Fixed link non-flash issue and made click flash more prominent

Arjunsivakumar28:
- Made fix based on testing comment from @lavishaGit "Flashing cursor works well with click on element "Pro-choice " or "Pro-life" on main page  that open on same tab but i noticed under FAQ page when user click on "Let's get started!" link or "please donate monthly" link  the cursor flashing gets delayed"
- Increased flash size and ripples to make it more prominent
- Increased flash delay to 1 second, and added constant `flashDelayMs` to the page `flashCursorClickListener` (`src/js/common/utils/flashCursorClickListener.js`) which can be modified accordingly

### DIFF
--- a/src/css/main.css
+++ b/src/css/main.css
@@ -4847,28 +4847,96 @@ a {
   pointer-events: none;
   z-index: 2147483647;
 }
-
-/* ripple element from click */
+/* main ripple */
 .clickRipple {
   position: fixed;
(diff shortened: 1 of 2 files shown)


DaleMcGrew:
Thank you for this iteration @Arjunsivakumar28! 👍

PR #4749, opened 2026-03-28
Add ability to run WebApp in a local docker environment

josephevans:
This PR gives the ability for developers to run the WebApp locally in a docker container, to avoid installing all the dependencies globally on their local machine.

1. Create compose.yaml to define docker container config.
2. Add actual Dockerfile.dev for running in local development environment
3. Update new production Dockerfile to use node v25 (instead of v22)
4. Add DOCKER.md instructions for setting up local Docker-based dev environment
5. Add new dev target in package.json which sets host parameter to 0.0.0.0 (instead of localhost) since we will be running in a container.

### DIFF
--- a/compose.yaml
+++ b/compose.yaml
@@ -1,0 +1,17 @@
+name: wevote
+
+services:
+  webapp:
+    build:
+      dockerfile: docker/Dockerfile.dev
+    volumes:
+      - .:/app
+    ports:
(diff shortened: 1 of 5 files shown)


DaleMcGrew (approved):

DaleMcGrew:
👍

PR #4752, opened 2026-03-28
fix: use anonymous Docker volume for node_modules to prevent host conflicts

josephevans:
## Problem

When running the WebApp in Docker with the local directory bind-mounted (`- .:/app`), the host machine's `node_modules` directory (which may be empty, missing, or built for a different OS/architecture) gets mounted into the container, shadowing the `node_modules` that were installed during `RUN npm install` in the Dockerfile. This causes the container to fail to start or run with incorrect dependencies.

## Solution

1. **Add `.dockerignore`** — Excludes `node_modules` from the Docker build context so the host's `node_modules` are never copied into the image.

2. **Add anonymous volume for `/app/node_modules` in `compose.yaml`** — By declaring `- /app/node_modules` as a separate volume in the compose config, Docker preserves the container's own `node_modules` (installed at image build time) and prevents the bind-mounted host directory from overwriting it.

3. **Clean up `Dockerfile.dev`** — Removes the `ENV PATH` line that manually added `node_modules/.bin` to `PATH`, since this is no longer needed with the corrected volume setup.

## Changes

| File | Change |
|------|--------|
| `.dockerignore` | Added — excludes `node_modules` from build context |
| `compose.yaml` | Added anonymous volume mount for `/app/node_modules` |
| `docker/Dockerfile.dev` | Removed redundant `ENV PATH` line |

## Testing

After these changes, `docker compose up` should correctly use the container-installed `node_modules` regardless of whether a `node_modules` directory exists on the host.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,1 @@
+node_modules
--- a/compose.yaml
+++ b/compose.yaml
@@ -6,6 +6,10 @@ services:
       dockerfile: docker/Dockerfile.dev
     volumes:
       - .:/app
+      # Create an anonymous volume for node_modules inside the container
+      # This prevents the host's (potentially empty or incompatible) node_modules
+      # from shadowing the container's installed node_modules
+      - /app/node_modules
     ports:
       - "127.0.0.1:3000:3000" # this allows access to the WeVote WebApp at localhost:3000  
     networks:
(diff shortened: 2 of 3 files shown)


DaleMcGrew (approved):
👍

PR #4777, opened 2026-04-13
modify dockerfile.dev and compose.yaml

mjacquot1:
Epic: https://wevoteusa.atlassian.net/browse/WV-4285
Child: https://wevoteusa.atlassian.net/browse/WV-4286

- Run containers as a non-root user. This reduces the potential impact of any malicious scripts by limiting system-level access.
- Use a named volume for node_modules. This prevents leftover or “dangling” dependencies from persisting across builds and causing inconsistencies. (Addendum, anonymous volumes persist after docker compose down, and on relaunching the container a new volume is created. This leaves a dangling volume of ~750mb)
- Adopt multi-stage Docker builds. Dependencies will be installed during the image build process rather than at runtime, ensuring a more controlled and reproducible environment.

### DIFF
--- a/compose.yaml
+++ b/compose.yaml
@@ -6,10 +6,7 @@ services:
       dockerfile: docker/Dockerfile.dev
     volumes:
       - .:/app
-      # Create an anonymous volume for node_modules inside the container
-      # This prevents the host's (potentially empty or incompatible) node_modules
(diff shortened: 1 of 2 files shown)


josephevans (approved):
Thanks for the PR, looks good.

DaleMcGrew:
👍

PR #4787, opened 2026-04-20
[WV-4269] BallotScrollingContainer: Replace CSS :has() selectors with explicit $hasIssues prop

ric0de:
Prevent Support/Opinion columns from sliding left of the candidate photo when no topics exist by adding an 80px IssuesColumn spacer. 

Moved inline flex styles (ISSUES_WIDTH, OPINIONS_WIDTH) into styled components and added a no-op guard on the IssueStore listener to avoid unnecessary re-renders.

EXAMPLES:

image
image


DaleMcGrew:
Looks great @ric0de! 👍

PR #4789, opened 2026-04-21
'Make public' in ItemActionBar saves visibility directly instead of opening "edit opinion" modal

ric0de:
Wire candidates and measures to save SHOW_PUBLIC via voterPositionVisibilitySave, update snackbar copy to cover oppose/no votes.

After clicking "Make public":
image


DaleMcGrew:
👍

PR #4790, opened 2026-04-21
Hide Opinions header when candidate has no other content and no opinions

ric0de:
- CandidateOpinionsColumn hides OpinionsCountHeader when no issues/support/oppose column exists and opinionsCount is 0
- QOL change: VoterPositionEntryAndDisplay accepts noBottomMargin to suppress the 12px bottom margin on VoterPositionContainer to avoid awkward trailing spacing; passed from CandidateOpinionsColumn when there are no opinions below
- DRY: Hoist shared hasContentBeforeOpinions in BallotScrollingContainer to avoid re-evaluating the same expression twice per candidate

Examples:
image


DaleMcGrew:
👍

PR #4798, opened 2026-05-04
Fix mobile "Invite selected" kebab so its menu opens

ric0de:
## What this fixes

On mobile, the triple-dot button next to **Invite selected** near the bottom of the screen did nothing when clicked. It now opens the same options menu the desktop view already shows.

**How:** wires the overflow button into the existing menu state and anchors the dropdown above the fixed footer so the panel stays in view.

Mobile invite-selected kebab now opens menu

## Adjacent cleanup

Picked up while in the same files:

- Kebab hover contrast on selected vs unselected rows
- Dropdown repositioning when scrolling
- A couple of toast strings
- Hide / unhide wiring

Kebab hover contrast
Toast / wiring cleanup


DaleMcGrew:
👍

PR #4804, opened 2026-05-07
Reduce CLS on politician details page with endorsements skeleton and conditional min-height

Almee98:
### Summary
- Added `PoliticianDetailsEndorsementsSkeleton` component with two variants (`full` and `feedOnly`) that placeholder the "What people are saying" endorsements area while data loads, preventing content below (e.g. "Other candidates, same office") from shifting upward.
- Wrapped the real endorsements content in `EndorsementsContentWrapper` with a conditional `minHeight` that holds vertical space during loading and releases it once the position retrieval API has responded; whether the response contains endorsements or not.
- Added `positionListRetrieveCompleteByBallotItem` flag to `CandidateStore` so the store emits a change even when the positions API returns zero results. Previously, a zero-count response was silently discarded (`return state`), making it impossible for the page to distinguish "still loading" from "confirmed no endorsements."
- `PoliticianDetailsPage.onCandidateStoreChange` now checks this flag to set `positionRetrieveCompleted`, which controls whether the min-height wrapper stays active.

### How it works
1. While waiting for the politician ID, a `full` skeleton (title + composer + 5 card placeholders) reserves space.
2. Once the ID arrives, real title and composer render inside `EndorsementsContentWrapper` with `minHeight` active, and a `feedOnly` skeleton shows while positions hydrate.
3. When the positions API responds, `CandidateStore` sets the completion flag and emits; even for zero-count responses.
4. The page sees `positionRetrieveCompleted: true` and drops the `minHeight`. If endorsements exist, they fill the space naturally. If not, the area collapses to its real height with no wasted space.

## Files changed
- `src/js/common/pages/Politician/PoliticianDetailsEndorsementsSkeleton.jsx` : New skeleton component with `full` and `feedOnly` variants
- `src/js/common/pages/Politician/PoliticianDetailsPage.jsx` : Endorsements content wrapper with conditional min-height, position retrieval completion check
- `src/js/stores/CandidateStore.js` : `positionListRetrieveCompleteByBallotItem` dictionary, getter, and updated reducer to emit on zero-count responses

## Testing done
- Loaded a politician with many endorsements (Kamala Harris) : no CLS during load, endorsements render normally, no blank gap after loading
- Loaded a politician with no endorsements (Adrian Smith) : skeleton shows during load, blank gap disappears once API confirms empty
- Loaded a politician with few endorsements (Marianne Williamson) : no persistent blank gap below the endorsement cards
- Navigated between politicians : skeleton resets and shows for the new politician

### DIFF
--- a/src/js/common/pages/Politician/PoliticianDetailsEndorsementsSkeleton.jsx
+++ b/src/js/common/pages/Politician/PoliticianDetailsEndorsementsSkeleton.jsx
@@ -1,0 +1,179 @@
+import { Box, Skeleton } from '@mui/material';
+import PropTypes from 'prop-types';
+import React from 'react';
+import styled from 'styled-components';
+import DesignTokenColors from '../../components/Style/DesignTokenColors';
+import { CampaignSubSectionTitleWrapper, CommentsListWrapper } from '../../components/Style/CampaignDetailsStyles';
+import { renderLog } from '../../utils/logging';
+
+// Keep in sync with PoliticianDetailsPage PoliticianEndorsementsList startingNumberOfPositionsToDisplay
+const ENDORSEMENT_SKELETON_CARD_COUNT = 5;
+
+//  Reserve vertical space so the loaded endorsements area (title + composer + up to five endorsements)
+//  is less likely to be shorter than this placeholder, reducing CLS when real content mounts. 
+const EndorsementsSkeletonRoot = styled('div')(({ theme }) => ({
+  minHeight: '480px',
+  [theme.breakpoints.up('md')]: {
+    minHeight: '680px',
+  },
+}));
+
+const ComposerOuter = styled('div')`
+  align-items: flex-start;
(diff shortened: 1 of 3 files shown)


Almee98:
@DaleMcGrew, This ticket involved tracing data flow across several layers (page components, store reducers, retrieve controllers, and API responses) to understand why the endorsements area was causing layout shift and how to reliably detect when position retrieval had completed.

This is the first time I've worked through a change of this scope in this codebase, so while the solution works and I've tested it across several scenarios, I recognize there may be more idiomatic or efficient approaches that I'm not yet aware of. I'm very open to revisiting any part of this. I welcome any and all feedback on the implementation.

I really appreciate the opportunity to work on this issue. It was a great learning experience and helped me build a much deeper understanding of how the stores and data retrieval pipeline work together.

DaleMcGrew:
Great work @Almee98! 👍

PR #4818, opened 2026-05-19
[WV-2660] Election Finder: National filter fix, date-below-title layout, About office on candidates

ric0de:
- Fix /election-finder/na: match elections explicitly tagged state_code='NA' (or 'NA' in state_code_list) instead of treating empty state_code_list as national, which was sweeping in state-level elections that just have unpopulated lists.
- Add "National" option to the state dropdown (top of list, above alphabetical states) on Home and ForState.
- Stack title and date vertically: title on top, date below in "Mon D, YYYY" format. On Home, non-national rows prefix the election name with "<State> - <Election Name>"
- Add "About office" entry to the candidate row in both the desktop action bar and the kebab menu, gated behind nextReleaseFeaturesEnabled, stub onClick — mirroring the office row exactly.


DaleMcGrew:
👍

PR #4819, opened 2026-05-19
Clean up Tracking subtabs: shared styles file, alphabetize, fix lint warnings

ric0de:
- Move the shared Tracking styled-components into their own file (Style/SupporterTrackingStyles.jsx) so they don't pollute Style/ManageMyCandidates or SupporterTracking.

- Add a small TrackingHeaderActionContext file so the three subtabs can share the header-action slot without a circular import.

- Alphabetize the styled-components in all the Tracking files and the shared style files, just to keep things easy to find.

- Add PropTypes to ActionPill, the Menus exports, SupportersJoined/Invited/ToRemind, and CardOpinionRow. Also move getPendingActions/getActionsLabel out of SupportersJoined so the useMemo hook stops complaining about missing deps.

### DIFF
--- a/src/js/components/ManageMyCandidates/ActionPill.jsx
+++ b/src/js/components/ManageMyCandidates/ActionPill.jsx
@@ -1,36 +1,57 @@
 import React from 'react';
+import PropTypes from 'prop-types';
 import styled from 'styled-components';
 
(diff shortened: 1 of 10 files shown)


DaleMcGrew:
👍

PR #4821, opened 2026-05-20
[WV-4547] Election Finder UI iter #2

ric0de:
- **ElectionFinderHeader**: state label now renders after the breadcrumbs (was above the H1), so the layout reads Election Finder → breadcrumb → STATE → election title → date.
- **ElectionFinderForElection**:
  - 3-crumb breadcrumb: `← Election Finder Home / Upcoming|Past Elections - State (N) / Election Name - Date (officeCount)`.
  - Election title now shows office count, e.g. `State Primary (2)`.
  - Date rendered below the title via a new `ElectionDetailDate` styled component (matches state label color so the two frame the title in the same hue).
  - Empty-state copy split: `No results found.` is preserved for search-no-match; a new `Our team hasn't assembled the data for this election yet. We usually have election data 45 days before each election.` message appears when the election has zero offices and no active search.
- **ElectionFinderStateLabel**: slightly larger state label scoped to the Finder (extends shared `ElectionStateLabel`) so Ballot pages keep the original size.
- **DRY**: extracted `formatDateLong` into new `dateHelpers.js`; refactored `ElectionFinderHome` and `ElectionFinderForState` to import it (drops two duplicate copies).

---

**Depends on: (`[WV-2660]` Election Finder iter 1). I'll rebase this branch once that lands so the diff collapses to a single commit.

[WV-2660]: https://wevoteusa.atlassian.net/browse/WV-2660?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/js/pages/ElectionFinder/ElectionFinderForElection.jsx
+++ b/src/js/pages/ElectionFinder/ElectionFinderForElection.jsx
@@ -18,12 +18,13 @@ import SnackNotifier from '../../common/components/Widgets/SnackNotifier';
 import ElectionStore from '../../stores/ElectionStore';
 import CopyChip from './CopyChip';
 import copyAndToast from './copyAndToast';
+import formatDateLong from './dateHelpers';
 import ElectionFinderHeader from './ElectionFinderHeader';
 import RowKebabMenu from './RowKebabMenu';
 import {
   ActionDivider, DarkTooltip,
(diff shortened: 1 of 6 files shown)


DaleMcGrew:
👍

PR #4829, opened 2026-06-01
WV-487: Fixed optimistic count update when following candidate and user is not logged or has missing information

Arjunsivakumar28:
Added count update to submitActionClick function in HeartFavoriteToggleLive.jsx, where after the profile modal is completed, it updates the counts locally before calling backend update functions


DaleMcGrew:

PR #4830, opened 2026-06-03
WV-2472: Fix the More menu toggle and close the gap above the footer (mobile)

0x416c616e:
The More button didn't toggle as expected because the `ClickAwayListener` and the More button's `onChange` toggle were both triggering on the same click — the menu would close and immediately reopen, making it look like it wasn't closing at all.

I put the More button's `anchorRef` into `MoreMenuOverlay` and made the `ClickAwayListener` ignore clicks when the target is inside the anchor, so the button's `onChange` is the sole toggler.

Also closed the small CSS gap between the expanded More menu and the footer below it by aligning the menu's bottom edge to the footer height (and removing a leftover `translateY` nudge).


DaleMcGrew:
👍

PR #4843, opened 2026-06-23
Regression failure fix for Privacy page

charanya-QA:


charanya-QA:
Verified the changes in browserstack and here is the link for the execution https://automation.browserstack.com/projects/Default+Project/builds/Verifying+WV/14?tab=tests&testListView=spec

PR #4859, opened 2026-07-07
WV-2664: Persist heart like/dislike icon and keep support tally in sync

ric0de:
TL;DR: The heart (like/dislike) and the Choose/Oppose position are now independent soclicking one no longer secretly flips the other. Support counts stay accurate and update live, and a batch of loading-race glitches are gone: phantom +1 counts, a heart stuck in the wrong state, and a frozen "complete your profile" popup.

- OrganizationStore optimistically mirrors the server's first-action-wins coupling, and organizationsFollowedRetrieve now rebuilds the politician follow/dislike lists from server truth so stale optimistic entries get reconciled (not just appended).
- ItemActionBar optimistically bumps the campaign tally only on a genuine clean-slate first action, gated on the follow lists being hydrated to avoid over-counting an already-hearted voter.
- HeartFavoriteToggleLive syncs counts live, floors them at 0, no-ops redundant actions, and no longer leaves the complete-your-profile modal stuck open.
- CampaignSupportThermometer reads the voter's current stance from SupportStore; CampaignStore preserves an explicit 0 supporters/opposers count.

### DIFF
--- a/src/js/common/components/CampaignSupport/CampaignSupportThermometer.jsx
+++ b/src/js/common/components/CampaignSupport/CampaignSupportThermometer.jsx
@@ -9,7 +9,7 @@ import { renderLog } from '../../utils/logging';
 import DesignTokenColors from '../Style/DesignTokenColors';
 import numberWithCommas from '../../utils/numberWithCommas';
 import CampaignStore from '../../stores/CampaignStore';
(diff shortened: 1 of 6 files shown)


DaleMcGrew:
👍

PR #4860, opened 2026-07-10
Preliminary changes for Cordova Android

SailingSteve:
Fixed "Import contacts from Gmail" in WebApp, there was a data problem in one of my contacts.  Still broken for Android -- the data is ok, but Google is requiring an updated OAuth flow.


DaleMcGrew:
Thank you @SailingSteve! 👍

PR #4866, opened 2026-07-15
Disabled the "Sign in with Google" (and get contacts from Google/gmai…

SailingSteve:
…l) for Android.

Android needs a new "Sign in with Google" implementation due to new OAuth requirements at Google. Dale said to disable it for now for Android, pending a UI redesign.

The Friends page navigation has been broken and loopy for years with tons of bad copied code and a patched UI with confusingly labeled buttons. So this "disabling change" for Android hit in many places.  Some of those pages are very hard to navigate to.  So I patched it some more -- Friends needs a "from zero" UI redesign and a complete code rewrite.


DaleMcGrew:
👍

PR #4882, opened 2026-07-28
Add tests for footer social links

amitha-m-joe:
- All 9/9 passing on BrowserStack (Chrome)

Screenshot 2026-07-28 at 1 23 11 PM

Test Coverage: icon visibility, platform existence, href correctness (Facebook/Instagram/Twitter), target="_blank" behavior, new-tab-stays-open, hover cursor state, broken image check.

### DIFF
--- a/tests/browserstack_automation/page_objects/footerSocialLinks.browser.js
+++ b/tests/browserstack_automation/page_objects/footerSocialLinks.browser.js
@@ -1,0 +1,66 @@
+import { $, browser,driver } from '@wdio/globals';
+import PageBrowser from './page.browser';
+
+// Page object for the WeVote homepage footer social media links.
+// All element IDs confirmed from DOM inspection of quality.wevote.us.
+// The homepage uses weVote-prefixed IDs (e.g. weVoteFacebook),
+// which are different from the credits page footer (footerLinkFacebook).
+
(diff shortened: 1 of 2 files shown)


charanya-QA (commented):
@amitha-m-joe  : 1) Can you  Please remove the driver.pause(waitTime) since you are already using waitforDisplayed already
2) Also if you could add the repeated steps of all the test case before each?beforeEach(async () => {
  await FooterSocialLinksPage.load();
  await FooterSocialLinksPage.scrollToFooter();
  await FooterSocialLinksPage.facebookLink.waitForDisplayed({
    timeout: waitTime,
  });
});
then each test would have only assertions
3) TC-011 open a new tab but doesn't closes it. When we execute all the modules this would be a problem , Can you pls close that too?

rest all looks good ! Thanks for attaching the screenshot of browserstack.

amitha-m-joe:
Thanks! Made all three changes. Re-ran on BrowserStack, still 9/9 passing: [new build link: https://automation.browserstack.com/builds/e32rzugko3ihkpzqsft7qajjohz9ptu96vb04bzg]

charanya-QA (approved):
Thanks @amitha-m-joe  !

PR #4885, opened 2026-07-29
WV-4818 Make Cordova photo upload button area look like WebApp's

SailingSteve:
Photo upload had also been completely broken and now it works again.
Also some fixes for AddressSpanLimited when in full screen webapp


DaleMcGrew:
👍

PR #4889, opened 2026-08-04
WV-4849 Add isClaimedProfile to candidate analytics

lyndabanh:
* CandidateStore.getAnalyticsCandidateDetails() previously omitted
isClaimedProfile entirely, so GA4 events fired via the candidate
path (support/oppose/favorite) showed (not set) for this field.

* Resolves isClaimedProfile from the linked Politician record first,
since Politician.is_claimed_profile is the source of truth


DaleMcGrew:
Thank you @lyndabanh! 👍

